PR #6, opened 2018-11-14
AmberScript: Parse BUFFER commands.

### LINKED ISSUE #5: AmberScirpt: Parse BUFFER commands
Add parsing for the buffer commands in amber script.

dj2:
This CL adds parsing for the various buffer commands available in
AmberScript.
```
BUFFER vertex my_buf DATA_TYPE vec2<int32> SIZE 5 FILL 0
BUFFER index my_buf DATA_TYPE int32 SIZE 5 SERIES_FROM 2 INC_BY 1
BUFFER framebuffer my_frame DIMS 256 256
BUFFER storage my_store DATA_TYPE uint32 DATA
1 2 3 4 5
END
```
For each type, the internal buffer will be filled with the data before
it leaves the parser.

Fixes #5

### DIFF
--- a/docs/amber_script.md
+++ b/docs/amber_script.md
@@ -15,12 +15,12 @@ TODO(dneto): What characters are valid in a name?
 ### Number literals
 
 Literal numbers are normally presented in decimal form.  They are interpreted
-as integers or floating point depending on context: a command parameter is predefined
-as either integral or floating point, or the data type is user-specified (such
-as for buffer data).
+as integers or floating point depending on context: a command parameter is
+predefined as either integral or floating point, or the data type is
+user-specified (such as for buffer data).
 
-Hex values: Whenever an integer is expected, you may use a hexadecimal number, which
-is the characters `0x` followed by hexadecimal digits.
+Hex values: Whenever an integer is expected, you may use a hexadecimal number,
+which is the characters `0x` followed by hexadecimal digits.
 
 ### Shaders
 
@@ -65,7 +65,7 @@ END
  * float
  * double
  * vec[2,3,4]\<type>
- * mat[2,3,4]\<type>    -- useful?
+ * mat[2,3,4]x[2,3,4]\<type>    -- useful?
 
 TODO(dneto): Support half-precision floating point.
 
@@ -77,7 +77,6 @@ Sized arrays and structures are not currently representable.
  * vertex
  * index
  * sampled
- * storage
  * color
  * depth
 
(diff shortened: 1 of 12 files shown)


docs/amber_script.md
```diff
@@ -90,11 +89,13 @@ BUFFER <buffer_type> <name> DATA_TYPE <type> DATA
 <value>+
 END
 
-BUFFER <buffer_type> <name> DATA_TYPE <type> SIZE <size_in_bytes> <initializer>
+BUFFER <buffer_type> <name> DATA_TYPE <type> SIZE <size_in_items> <initializer>
```
jaebaek: What does `size_in_items` mean?

src/amberscript/buffer.h
```diff
@@ -0,0 +1,70 @@
...
+
+enum class BufferType : uint8_t {
+  kColor = 0,
+  kDepth,
+  kFramebuffer,
+  kIndex,
+  kSampled,
+  kStorage,
+  kUniform,
+  kVertex
```
jaebaek: I think we should remove `kFramebuffer` and `kSampled` and add `kStorageTexel`, `kUniformTexel`.
If I understand this correctly, it should include
- color attachment and depth/stencil attachment for framebuffer
- vertex, index
- [descriptors](https://www.khronos.org/registry/vulkan/specs/1.1/html/vkspec.html#descriptorsets-types)

jaebaek: And I want to doublecheck, do `kColor` and `kDepth` mean color attachment and depth/stencil attachment for framebuffer that we will probe?

docs/amber_script.md
dj2: It's the number of items in the buffer, as opposed to the size_in_bytes, so 5 int32's would have a size_in_items of 5 and a size_in_bytes of 20. I wanted to make it clear what number we expected here.

src/amberscript/buffer.h
dj2: Why would descriptors be here? They aren't a buffer are they? I'm trying to keep this agnostic from Vulkan, so want to avoid things like attachments.

kFramebuffer is how I know the buffer is a framebuffer and can be used in the subsequent FRAMEBUFFER calls inside a PIPELINE. Does a framebuffer need to be defined by multiple buffers in Vulkan?

jaebaek: Storage and SampledImage are descriptors.
Framebuffer can have color and depth/stencil buffers.

I will file a bug to discuss how we should change those types.

Issue #3, opened 2018-11-14
Setup build bots


dj2:
Setup bots to run automatically on each PR.

dj2:
Need Linux (gcc and clang) and Mac bots to start. Need a Windows bot for VS2017 once the Windows build is running.

dj2:
Linux and Mac bots are up and running unittests. Vulkan tests are not being run at the moment, need to see if we can get Vulkan on the bots. Windows bots will be added once #4 is complete.

dj2:
The invocation logs appear to be missing from the public build logs and need to be fixed.

dj2:
All of the build bots are up and running, the remaining failure is a Windows Release build issue.

PR #45, opened 2018-11-14
Add kokoro build files.

dj2:
This CL adds the build files for Linux {gcc|clang} and Mac clang builds
in both Debug and Release mode.

Issue #3


tools/check_code_format.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+# Copyright 2018 Google LLC
```
dj2: Left this one as Google copyright.

Issue #7, opened 2018-11-14
Update image layout of FrameBuffer properly


jaebaek:
Image layout of FrameBuffer should be color attachment optimal when clear or draw on it.
It should be transfer src optimal when copy it back to host visible buffer.

PR #46, opened 2018-11-14
Update image layout properly

jaebaek:
Image layout of FrameBuffer for color attachment should have its layout as
`VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL` when we clear
or draw something on it. The one of depth/stencil attachment should have
`VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL`.

When we probe it, its layout must be
`VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL`. Note that we do not
change its layout to `VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL`
explicitly because the render pass will do.

Issue #7

### DIFF
--- a/src/vulkan/image.cc
+++ b/src/vulkan/image.cc
@@ -187,6 +187,7 @@ void Image::ChangeLayout(VkCommandBuffer command,
                               VK_ACCESS_TRANSFER_WRITE_BIT;
       break;
     default:
+      barrier.srcAccessMask = 0;
       break;
   }
 
@@ -218,6 +219,7 @@ void Image::ChangeLayout(VkCommandBuffer command,
                               VK_ACCESS_TRANSFER_WRITE_BIT;
       break;
     default:
+      barrier.dstAccessMask = 0;
       break;
   }
 
--- a/src/vulkan/image.cc
+++ b/src/vulkan/image.cc
@@ -147,5 +147,82 @@ Result Image::CopyToHost(VkCommandBuffer command) {
   return {};
 }
 
+void Image::ChangeLayout(VkCommandBuffer command,
+                         VkImageLayout old_layout,
+                         VkImageLayout new_layout,
+                         VkPipelineStageFlags from,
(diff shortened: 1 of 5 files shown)


src/vulkan/image.cc
```diff
@@ -147,5 +147,82 @@ Result Image::CopyToHost(VkCommandBuffer command) {
...
+      break;
+    case VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL:
+      // An image becomes "transfer dst" only when we send a buffer data to
+      // it.
+      barrier.srcAccessMask = VK_ACCESS_SHADER_READ_BIT |
+                              VK_ACCESS_SHADER_WRITE_BIT |
+                              VK_ACCESS_TRANSFER_WRITE_BIT;
+      break;
+    default:
+      break;
```
dj2: Let's do barrier.srcAccessMask = 0; just to be clear what's happening.

src/vulkan/image.cc
```diff
@@ -147,5 +147,82 @@ Result Image::CopyToHost(VkCommandBuffer command) {
...
+    case VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL:
+      // An image becomes "transfer dst" only when we send a buffer data to
+      // it.
+      barrier.dstAccessMask = VK_ACCESS_SHADER_READ_BIT |
+                              VK_ACCESS_SHADER_WRITE_BIT |
+                              VK_ACCESS_TRANSFER_WRITE_BIT;
+      break;
+    default:
+      break;
```
dj2: Same here barrier.dstAccessMask = 0;

src/vulkan/image.cc
jaebaek: Done.

src/vulkan/image.cc
jaebaek: Done.

PR #50, opened 2018-11-14
Vulkan: add memory barrier (#9)

### LINKED ISSUE #9: Vulkan: Add memory barrier
Draw after sending vertex data to vertex buffer is not guaranteed currently, as well as probe after copying frame buffer to host accessible buffer.
We must add memory barrier for safety.

jaebaek:
Draw after sending vertex data is not guaranteed currently, as well
as probe after copying frame buffer to host accessible buffer.

This CL adds a memory barrier after copying to the device or to the host
to make sure that memory updates are complete when we access the
associated buffers.

Fixes #9


jaebaek:
Looks good to me

PR #51, opened 2018-11-14
Vulkan: allow probe without draw or clear

jaebaek:
See tests/cases/probe_without_clear_command.expect_fail.amber.
We must allow probe without clear or draw even though it may
result in test failures.

### DIFF
--- a/tests/cases/probe_without_clear_command.expect_fail.amber
+++ b/tests/cases/probe_without_clear_command.expect_fail.amber
@@ -1,0 +1,30 @@
+# Copyright 2018 The Amber Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+[vertex shader passthrough]
+
+[fragment shader]
+#version 430
+
+layout(location = 0) in vec4 color_in;
+layout(location = 0) out vec4 color_out;
+
+void
+main()
+{
+  color_out = color_in;
+}
+
+[test]
+relative probe rect rgba (0.0, 0.0, 1.0, 1.0) (0.1, 0.2, 0.3, 0.4)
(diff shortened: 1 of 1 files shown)


tests/cases/probe_without_clear_command.expect_fail.amber
```diff
@@ -0,0 +1,30 @@
...
+layout(location = 0) out vec4 color_out;
+
+void
+main()
+{
+  color_out = color_in;
+}
+
+[test]
+relative probe rect rgba (0.0, 0.0, 1.0, 1.0) (0.1, 0.2, 0.3, 0.4)
```
dj2: So, this is going to be potentially flaky won't it? We have no idea what's going to be in the buffer at that position, so we could possibly match this probe by accident?

Should we set the color_out to be vec4(.5, .5, .5, .5) so it's a known value?

tests/cases/probe_without_clear_command.expect_fail.amber
```diff
@@ -0,0 +1,30 @@
...
+[fragment shader]
+#version 430
+
+layout(location = 0) in vec4 color_in;
+layout(location = 0) out vec4 color_out;
+
+void
+main()
+{
+  color_out = color_in;
```
dj2: If this is a known value (like vec4(.5, .5, .5, .5) then it's possible for this test to pass yes? Can we create a positive test  for probe without draw/clear where the shader sets a colour?

tests/cases/probe_without_clear_command.expect_fail.amber
dneto0: Can we set epsilon to infinity (or close enough), and then probe a nonsensical value and it should always pass.

jaebaek: The issue here is that we should accept the probe without any operation.
I discuss it with @dj2 in offline and we should print an error message if it probes without any operations i.e., no draw, no clear because it is not meaningful and it would be probably a mistake from the script writer.

Let me just keep the assert I removed. This test should fail without the assert. I will put a comment for the failure message at the bottom of this script.

jaebaek:
PTAL

tests/cases/probe_without_clear_command.expect_fail.amber
```diff
@@ -27,8 +27,8 @@ main()
 }
 
 [test]
-clear color 0.1 0.2 0.3 0.4
-
-# No clear command before probe. Random pixels are expected.
-
 relative probe rect rgba (0.0, 0.0, 1.0, 1.0) (0.1, 0.2, 0.3, 0.4)
+
+# Expected result: assert failure with the following message
```
dj2: Can you file a bug, this might be an interesting thing to add to the test harness so we can verify that this was the output.

jaebaek: Opened a bug as #60.

dneto0:
Please rebase to avoid conflicts

dj2:
This had a full set of red bots and no approval, why was it merged?

jaebaek:
It is my fault .. sorry. I will quickly check what is the problem.

jaebaek:
It is weird .. in the commit history https://github.com/google/amber/commits/master, it is shown as green.

jaebaek:
Sure, I will do. Thanks!

PR #52, opened 2018-11-14
Vulkan: add draw triangles

jaebaek:
Add tests/cases/draw_triangle_list.amber to test drawing simple
triangles and update engine to conduct draw based on parameters of
draw array command.

### DIFF
--- a/src/vulkan/engine_vulkan.cc
+++ b/src/vulkan/engine_vulkan.cc
@@ -15,6 +15,7 @@
 #include "src/vulkan/engine_vulkan.h"
 
 #include <algorithm>
+#include <cassert>
 
 #include "src/make_unique.h"
 #include "src/vulkan/format_data.h"
@@ -44,7 +45,7 @@ VkShaderStageFlagBits ToVkShaderStage(ShaderType type) {
       return VK_SHADER_STAGE_COMPUTE_BIT;
   }
 
-  // Unreachable
+  assert(false && "Vulkan::Unknown shader stage");
   return VK_SHADER_STAGE_FRAGMENT_BIT;
 }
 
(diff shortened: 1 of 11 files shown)


tests/cases/draw_triangle_list.amber
```diff
@@ -0,0 +1,104 @@
...
+# limitations under the License.
+
+[vertex shader]
+#version 430
+
+layout(location = 0) in vec4 position;
+layout(location = 1) in vec4 vert_color;
+layout(location = 0) out vec4 frag_color;
+
+void
```
dj2: Let's just do `void main() {` to make this 2 lines shorter.

tests/cases/draw_triangle_list.amber
```diff
@@ -0,0 +1,104 @@
...
+  frag_color = vert_color;
+}
+
+[fragment shader]
+#version 430
+
+layout(location = 0) in vec4 frag_color;
+layout(location = 0) out vec4 final_color;
+
+void
```
dj2: ditto

dj2:
Is there an associated bug for this CL?

tests/cases/draw_triangle_list.amber
jaebaek: done

tests/cases/draw_triangle_list.amber
jaebaek: done

jaebaek:
PTAL

src/vulkan/engine_vulkan.cc
```diff
@@ -44,7 +45,7 @@ VkShaderStageFlagBits ToVkShaderStage(ShaderType type) {
       return VK_SHADER_STAGE_COMPUTE_BIT;
   }
 
-  // Unreachable
+  assert(false && "Vulkan::Unknown shader stage");
```
dneto0: asserts only fire on debug builds. Should we have a project-wide macro or function to express the unreachable condition even in release builds?
Question for @dj2 as well.

tests/cases/draw_triangle_list.amber
```diff
@@ -0,0 +1,100 @@
...
+#         R8   G8     R8  G8  B8
+           0 -128      0 255   0
+         127  127      0 255   0
+           0  127      0 255   0
+
+           0 -128      0 255   0
+         127  127      0 255   0
+         127 -128      0 255   0
+
+#          Blue for quater frame
```
dneto0: "quarter"

src/vulkan/engine_vulkan.cc
dj2: There is an open bug to remove the asserts and return Results. Just have to do the refactoring.

tests/cases/draw_triangle_list.amber
jaebaek: Done

PR #56, opened 2018-11-15
Missing kokoro scripts

dj2:
These got missed when adding the original scripts.

Issue #3.


kokoro/scripts/linux/build.sh
```diff
@@ -0,0 +1,67 @@
...
+
+mkdir build && cd $SRC/build
+
+# Invoke the build.
+BUILD_SHA=${KOKORO_GITHUB_COMMIT:-$KOKORO_GITHUB_PULL_REQUEST_COMMIT}
+echo $(date): Starting build...
+cmake -GNinja -DCMAKE_BUILD_TYPE=$BUILD_TYPE $CMAKE_C_CXX_COMPILER ..
+
+echo $(date): Build everything...
+ninja
```
jaebaek: Do we want to compile it with only a single core?
If not, we should add `-j4` here?

kokoro/scripts/linux/build.sh
```diff
@@ -0,0 +1,67 @@
...
+echo $(date): Build completed.
+
+echo $(date): Starting ctest...
+ctest -j4 --output-on-failure --timeout 300
+echo $(date): ctest completed.
+
+echo $(date): Starting integration tests..
+../../test/run_tests.py
```
jaebaek: Later, this must check failure message for negative tests.
But it is fine now.
Could you please just put a TODO comment here?

kokoro/scripts/macos/build.sh
```diff
@@ -0,0 +1,51 @@
...
+
+CMAKE_C_CXX_COMPILER="-DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++"
+
+# Invoke the build.
+BUILD_SHA=${KOKORO_GITHUB_COMMIT:-$KOKORO_GITHUB_PULL_REQUEST_COMMIT}
+echo $(date): Starting build...
+cmake -GNinja -DCMAKE_BUILD_TYPE=$BUILD_TYPE $CMAKE_C_CXX_COMPILER ..
+
+echo $(date): Build everything...
+ninja
```
jaebaek: ditto

kokoro/scripts/macos/build.sh
```diff
@@ -0,0 +1,51 @@
...
+echo $(date): Build completed.
+
+echo $(date): Starting ctest...
+ctest -j4 --output-on-failure --timeout 300
+echo $(date): ctest completed.
+
+echo $(date): Starting integration tests..
+../../test/run_tests.py
```
jaebaek: ditto

kokoro/scripts/linux/build.sh
```diff
@@ -0,0 +1,67 @@
...
+
+BUILD_ROOT=$PWD
+SRC=$PWD/github/amber
+CONFIG=$1
+COMPILER=$2
+
+BUILD_TYPE="Debug"
+
+CMAKE_C_CXX_COMPILER=""
+if [ $COMPILER = "clang" ]
```
jaebaek: Is the default one gcc?

kokoro/scripts/macos/build.sh
```diff
@@ -0,0 +1,51 @@
...
+unzip -q ninja-mac.zip
+chmod +x ninja
+export PATH="$PWD:$PATH"
+
+cd $SRC
+./tools/git-sync-deps
+
+mkdir build && cd $SRC/build
+
+CMAKE_C_CXX_COMPILER="-DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++"
```
jaebaek: Must we compile it with clang in MacOS always?

kokoro/scripts/linux/build.sh
dj2: Ninja will pick the -j value that best matches your host if none is provided.

kokoro/scripts/linux/build.sh
dj2: Not sure what you mean? If any test fails (so, a negative test passes or a positive test fails) the run_test will return a non-zero exit which should fail the script I believe.

kokoro/scripts/linux/build.sh
dj2: Seems to be the case, this is what spirv-tools does

kokoro/scripts/macos/build.sh
dj2: yes

kokoro/scripts/linux/build.sh
dneto0: That's what "set -e  # fail on error" is for at the top.

kokoro/scripts/macos/build.sh
dneto0: Same resolution as previous.

kokoro/scripts/macos/build.sh
dneto0: Same resolution as previous.

kokoro/scripts/linux/build.sh
jaebaek: Oh sorry. It already handles it.

jaebaek:
LGTM

PR #63, opened 2018-11-16
Vulkan: add SSBO

### LINKED ISSUE #11: Vulkan: Support SSBO

jaebaek:
Add SSBO and Vulkan objects to use descriptors.

Fixes #11


src/vulkan/descriptor.cc
```diff
@@ -53,18 +53,45 @@ Descriptor::Descriptor(DescriptorType type,
                        VkDevice device,
                        uint32_t desc_set,
                        uint32_t binding)
-    : device_(device),
-      descriptor_set_(desc_set),
+    : descriptor_set_(desc_set),
       binding_(binding),
-      type_(type) {}
+      type_(type),
+      device_(device) {}
```
dj2: Why did this change around?

src/vulkan/descriptor.cc
```diff
@@ -79,26 +106,40 @@ void Descriptor::UpdateDescriptorSet(VkDescriptorSet descriptor_set,
...
+    VkDescriptorSet descriptor_set,
+    VkDescriptorType descriptor_type,
+    const VkBufferView* texel_view) {
+  VkWriteDescriptorSet write = {};
+  write.sType = VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET;
+  write.dstSet = descriptor_set;
+  write.dstBinding = binding_;
+  write.dstArrayElement = 0;
+  write.descriptorCount = 1;
+  write.descriptorType = descriptor_type;
```
dj2: This setup can be moved to a common method so we don't duplicate it.

src/vulkan/storage_buffer_descriptor.cc
```diff
@@ -108,16 +108,15 @@ void StorageBufferDescriptor::SendDataToGPUIfNeeded(VkCommandBuffer command) {
...
   VkDescriptorBufferInfo buffer_info = {};
   buffer_info.buffer = buffer_->GetVkBuffer();
   buffer_info.offset = 0;
   buffer_info.range = VK_WHOLE_SIZE;
 
-  Descriptor::UpdateDescriptorSet(descriptor_set,
-                                  VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, nullptr,
-                                  &buffer_info, nullptr);
+  return Descriptor::UpdateDescriptorSetForBuffer(
+      descriptor_set, VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, &buffer_info);
```
dj2: Why not pass buffer_info by const-ref? We don't change it internally.

src/vulkan/descriptor.cc
jaebaek: Because I just moved `device_` to `private:`, the order of members became different.
Without this change, there is an initializing order error from `-Werror=reorder`.

src/vulkan/descriptor.cc
jaebaek: Done

src/vulkan/storage_buffer_descriptor.cc
jaebaek: Done.

src/vulkan/descriptor.cc
```diff
@@ -91,8 +99,9 @@ Result Descriptor::UpdateDescriptorSetForBuffer(
 Result Descriptor::UpdateDescriptorSetForImage(
     VkDescriptorSet descriptor_set,
     VkDescriptorType descriptor_type,
-    const VkDescriptorImageInfo* image_info) {
-  VkWriteDescriptorSet write = {};
+    const VkDescriptorImageInfo& image_info) {
+  VkWriteDescriptorSet write =
+      GetWriteDescriptorSet(descriptor_set, descriptor_type);
   write.sType = VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET;
   write.dstSet = descriptor_set;
```
dj2: This can be removed now, yea?

src/vulkan/descriptor.cc
```diff
@@ -119,8 +128,9 @@ Result Descriptor::UpdateDescriptorSetForImage(
 Result Descriptor::UpdateDescriptorSetForBufferView(
     VkDescriptorSet descriptor_set,
     VkDescriptorType descriptor_type,
-    const VkBufferView* texel_view) {
-  VkWriteDescriptorSet write = {};
+    const VkBufferView& texel_view) {
+  VkWriteDescriptorSet write =
+      GetWriteDescriptorSet(descriptor_set, descriptor_type);
   write.sType = VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET;
   write.dstSet = descriptor_set;
```
dj2: ditto

src/vulkan/descriptor.cc
jaebaek: Ah .. my mistake. Done.

src/vulkan/descriptor.cc
jaebaek: ditto

PR #65, opened 2018-11-18
amber tool: Add -e option to set graphics API engine

dneto0:


samples/amber.cc
```diff
@@ -41,6 +42,7 @@ const char kUsage[] = R"(Usage: amber [options] SCRIPT
   -i <filename>  -- Write rendering to <filename> as a PPM image.
   -b <filename>  -- Write contents of a UBO or SSBO to <filename>.
   -B <buffer>    -- Index of buffer to write. Defaults buffer 0.
+  -e <engine>    -- Specify graphics engine: vulkan, dawn. Default is vulkan.
```
dneto0: Not clear to me if -e is the best choice.   Other ideas: -g  for "graphics" or  -x for "execute"

PR #66, opened 2018-11-19
AmberScript updates to buffers

dj2:


docs/amber_script.md
```diff
@@ -156,6 +145,29 @@ is to run no optimization passes.
...
+  FRAMEBUFFER <width> <height>
+  FRAMEBUFFER <width> <height> <fb_format>
+```
+
+#### Buffer Types
+ * uniform
+ * storage
+ * sampled
+ * color
+ * depth
```
dj2: I guess this should match the various Vulkan types?

docs/amber_script.md
```diff
@@ -319,15 +317,15 @@ PIPELINE graphics kRedPipeline
...
+  FRAMEBUFFER 256 256
   ENTRY_POINT kFragmentShader red
 END  # pipeline
 
 PIPELINE graphics kGreenPipeline
   ATTACH kVertexShader
   ATTACH kFragmentShader
 
-  FRAMEBUFFER kFrameBuffer
+  FRAMEBUFFER 256 256
```
dj2: How do we encode the case where we want the framebuffer from the two pipelines to be the same, or to be different? My previous assumption was they'd have unique framebuffers, except in cases like this were I was passing the same FRAMEBUFFER into both of them.

docs/amber_script.md
```diff
@@ -405,6 +403,6 @@ CLEAR_COLOR kGraphicsPipeline 255 0 0 255
 CLEAR kGraphicsPipeline
 
 RUN kGraphicsPipeline \
-  DRAW_ARRAY kIndices IN kBuffer AS triangle_list \
+  DRAW_ARRAY kIndices IN kData AS triangle_list \
```
dj2: Currently RUN is where we use vertex and indices buffers, do they need to be attached into a PIPELINE with the VERTEX_BUFFER and INDEX_BUFFER calls we mentioned, or is just referencing them here enough?

If they're attached into a PIPELINE, does that mean that `RUN pipeline DRAW_ARRAY` will just use them implicitly? So this would just be:

```
 RUN kGraphicsPipeline DRAW_ARRAY AS triangle_list START_IDX 0 COUNT 24
```

docs/amber_script.md
```diff
@@ -156,6 +145,29 @@ is to run no optimization passes.
...
 
+##### Framebuffer Formats
+ * R32G32B32A32_UINT
+
+Bind a provided framebuffer.
+
+```
+  FRAMEBUFFER <width> <height>
+  FRAMEBUFFER <width> <height> <fb_format>
+```
```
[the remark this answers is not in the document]
dneto0: We have to decide if FRAMEBUFFER is a new resource allocation or is an attribute of an existing buffer saying "it's an image, and it will be used as a framebuffer".   
In Vulkan it's more the latter, where a VkFramebuffer wraps around an image view (which points to the underlying image storage), and also has a format attached to it.

So I would think of FRAMEBUFFER as a named type, like

     FRAMEBUFFER <decl-name> <width> <height>

e.g.
    FRAMEBUFFER fb 640 480

See suggested modification in examples on how to use this.

docs/amber_script.md
dneto0: Example: 

       BIND myim AS fb DESCRIPTOR_SET 1 BINDING 0

Then the underlying resource (myim) is a declared buffer (image) that already exists.  And you can act on it twice, and it's orthogonal.

docs/amber_script.md
dneto0: In Vulkan both index buffer and vertex buffers are bound in the command buffer before doing the draw call.  So it's ok to do it implicitly here.  The backend will translate this RUN call into the right bind-index-buffer and bind-vertex-buffer commands.    

I believe the vertex and index buffer bindings are totally different numbering spaces than descriptor set/binding pairs.  Beware.(?!)

docs/amber_script.md
dj2: Added the comment above.

docs/amber_script.md
dj2: ack

dj2:
PTAL.

I decided to run with `VERTEX_DATA` and `INDEX_DATA` to see what it looked it, and seems to make the `DRAW_ARRAY` call nicer. This then does raise the question, can you attach multiple vertex data buffers to a pipeline? If you want to have multiple data types in there you'd need to, or we need `struct` support for the data types.

Changed up the `FRAMEBUFFER` settings a bit to match what was discussed.

docs/amber_script.md
```diff
@@ -319,15 +338,15 @@ PIPELINE graphics kRedPipeline
 
   ATTACH kFragmentShader
 
-  FRAMEBUFFER kFrameBuffer
+  FRAMEBUFFER kFramebuffer 256 256
```
dneto0: Need DIMS keyword

docs/amber_script.md
```diff
@@ -156,6 +148,46 @@ is to run no optimization passes.
...
+ * depth
+
+TODO(dsinclair): Sync the BufferTypes with the list of Vulkan Descriptor types.
+
+
+Bindings can be written as:
+
+```
+  BIND BUFFER <buffer_name> AS <buffer_type> DESCRIPTOR_SET <id> \
+       BINDING <id> IDX <val>
```
dneto0: We could default IDX to 0

docs/amber_script.md
```diff
@@ -394,17 +408,23 @@ BUFFER vertex kData TYPE vec3<int32> DATA
...
 
+PIPELINE graphics kGraphicsPipeline
+  ATTACH kVertexShader
+  ATTACH kFragmentShader
+
+  VERTEX_DATA kData
+  INDEX_DATA kIndices
+END  # pipeline
+
 CLEAR_COLOR kGraphicsPipeline 255 0 0 255
```
dneto0: Now that FRAMEBUFFER is not declared in a pipeline, the CLEAR and CLEAR_COLOR have to name an buffer rather than a pipeline, I think.

docs/amber_script.md
```diff
@@ -156,6 +148,46 @@ is to run no optimization passes.
...
+TODO(dsinclair): Sync the BufferTypes with the list of Vulkan Descriptor types.
+
+
+Bindings can be written as:
+
+```
+  BIND BUFFER <buffer_name> AS <buffer_type> DESCRIPTOR_SET <id> \
+       BINDING <id> IDX <val>
+
+  BIND SAMPLER <sampler_name>
```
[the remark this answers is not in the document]
dj2: Done.

docs/amber_script.md
```diff
@@ -169,30 +201,19 @@ is to run no optimization passes.
...
-Descriptor sets can be bound as:
-
-```
-  DESCRIPTOR_SET <id> BINDING <id> IDX <val> TO <buffer_name>
-```
-
 ### Run a pipeline.
 
+When running a `DRAW_ARRAY` command you must attach the vertex data and index
+data to the `PIPELINE` with the `VERTEX_DATA` and `INDEX_DATA` commands.
```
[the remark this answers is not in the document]
dj2: Done. Also updated DRAW_ARRAY to provide a DRAW_ARRAY INDEXED option.

docs/amber_script.md
```diff
@@ -169,30 +201,19 @@ is to run no optimization passes.
...
 RUN <pipeline_name> <x> <y> <z>
 
 RUN <pipeline_name> \
   DRAW_RECT POS <x_in_pixels> <y_in_pixels> \
   SIZE <width_in_pixels> <height_in_pixels>
 
-RUN <pipeline_name> \
-  DRAW_ARRAY <indices_buffer> IN <data_buffer> \
-  AS <topology> START_IDX <value> COUNT <value>
+RUN <pipeline_name> DRAW_ARRAY AS <topology> START_IDX <value> COUNT <value>
```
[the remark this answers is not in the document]
dj2: Done, although count I tried to re-word slightly. Wanted to tie it back to the SIZE provided to the BUFFER that is being used.

docs/amber_script.md
dj2: Done.

docs/amber_script.md
dj2: Done.

docs/amber_script.md
dj2: Well, you're still clearing the drawing buffer for the pipeline aren't you as opposed to some other buffer?

docs/amber_script.md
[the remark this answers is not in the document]
dneto0: Yes. That's better.  I had a hard time wording it in the first place.

docs/amber_script.md
dneto0: Let's resolve Framebuffer semantics in a future CL.

PR #70, opened 2018-11-19
Move VkScript VertexData to a Buffer

dj2:
This CL uses the shared Buffer to hold the contents of the VertexData
block in VkScript instead of the ad-hoc Cell structure.


src/vkscript/parser.cc
```diff
@@ -372,14 +375,19 @@ Result Parser::ProcessVertexDataBlock(const std::string& data) {
...
       }
     }
-    node->AddRow(std::move(row));
   }
 
-  node->SetHeaders(std::move(headers));
+  auto node = MakeUnique<VertexDataNode>();
+  for (size_t i = 0; i < headers.size(); ++i) {
+    auto buffer = MakeUnique<Buffer>(BufferType::kVertex);
+    buffer->SetData(std::move(values[i]));
```
dneto0: The Buffer's DatumType  and Size attributes are not set.  Is that important?

src/vkscript/parser_test.cc
```diff
@@ -312,17 +312,19 @@ TEST_F(VkScriptParserTest, VertexDataHeaderFormatString) {
...
-  EXPECT_EQ(1U, headers[1].location);
+  auto& header_1 = data->GetHeader(1);
+  EXPECT_EQ(1U, header_1.location);
   EXPECT_EQ(FormatType::kA8B8G8R8_UNORM_PACK32,
-            headers[1].format->GetFormatType());
+            header_1.format->GetFormatType());
+  EXPECT_TRUE(data->GetSegment(1).empty());
 }
 
 TEST_F(VkScriptParserTest, VertexDataHeaderGlslString) {
```
dneto0: Not for this CL, but I'm confused by the  1/int/vec3 attribute header declaration. Is it sane to have a mismatch between the gl_type and the component of the glsl_type ?

dj2: I have no idea, this is directly from VkRunner, so I have to replicate what they do to maintain compatibility.

src/buffer.h
```diff
@@ -39,8 +39,12 @@ class Buffer {
   const DatumType& GetDatumType() const { return datum_type_; }
 
   void SetSize(size_t size) { size_ = size; }
-  size_t GetSize() const { return size_; }
-  size_t GetSizeInBytes() const { return size_ * datum_type_.SizeInBytes(); }
+  size_t GetSize() const {
+    if (size_ == 0 && !data_.empty())
```
dneto0: The test for data_.empty() is redundant. :-)

src/vkscript/parser.cc
dneto0: I think Buffer needs comments to describe its own internal data,on the data members and accessors.  The particularly odd one is the datum_type_ which should be described as only valid for an index buffer.

dj2: Added a couple comments. I added the invalid DatumType comment in the VertexData node class as that's where it's invalid. Buffer is also used in the Amber code where the DatumType is valid. It's just with VertexData that it goes wonky due to things like form D24_UINT_8_SINT or D32_SFLOAT_8_UINT not really being representable.....

src/buffer.h
dj2: Yea, removed.

dj2:
I'm going to do a followup to this cl where I create:

```
class Buffer {
  std::vector<Value>
}

class FormatBuffer : public Buffer {
  Format format;
}

class DataBuffer: public Buffer {
  DatumType datum_type;
}
```

Then, the AmberScript buffers will be DataBuffers, the Index buffer will be a DataBuffer and the VertexData buffer will be a FormatBuffer. This will remove the weird restriction that the DatumType is wrong for VertexBuffer.

dj2:
Actually, I'm going to close this and do it in other steps ....

PR #59, opened 2018-11-15
WIP: Windows build

dj2:
Setup the Windows build.

Issue #4

PR #71, opened 2018-11-19
Windows build support

dneto0:
Builds on #59 

And then also:
- force static linking of C Runtime into executables
- Use fopen_s to avoid security warning in MSVC

This version completes the build, then fails a unit test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ third_party/googletest
 third_party/shaderc
 third_party/spirv-tools
 third_party/spirv-headers
+.vs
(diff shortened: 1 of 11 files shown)


dneto0:
I think the Windows build support is ready now.

Getting failures on Linux and Windows, as described by the bots:
```
[ RUN      ] CommandParserTest.MaxDepthBounds
1093
../src/vkscript/command_parser_test.cc:1666: Failure
1094
Expected equality of these values:
1095
  3.f
1096
    Which is: 3
1097
  cp.PipelineDataForTesting()->GetMaxDepthBounds()
1098
    Which is: 3.4000001
1099
[  FAILED  ] CommandParserTest.MaxDepthBounds (0 ms)
```

I don't know where that's coming from.  I tried a GCC UBSAN debug build and it didn't catch the problem.  ???

dneto0:
Ok. The test had been accidentally broken by the first commit in this MR.  :-)

dj2:
Hah, nice find. And oops.

PR #74, opened 2018-11-20
Update Buffer size when data provided.

dj2:
This CL sets the size of the Buffer to correspond to the number of
elements in the provided data array. This size will override any
previously set size on the buffer.


dj2:
PTAL. I had to divide the data size by the row and column count to get the right number of "elements" (a vec2 is 1 element, but 2 data values)

PR #77, opened 2018-11-20
Add a fence_timeout option for VkScript.

### LINKED ISSUE #69: vkWaitForFences timeout for slow implementations
The `CommandBuffer::SubmitAndReset` method uses a hardcoded timeout of 100 ms for the `vkWaitForFences` call. I'm sure this works great for all real Vulkan implementations, but for poor old emulators like Talvos this causes problems. For the `draw_triangles` test case I can scrape through if I bump the timeout to 200 ms (or switch to a Release build), but I'm guessing it might start failing again for larger test cases (or slower CPUs).

Assuming that you don't want to raise this hardcoded limit to O(seconds) just to accommodate Talvos' inefficiencies, would you perhaps accept a PR to enable the limit to be dynamically increased via an environment variable or command-line flag?

P.S. This looks to be an incredibly useful project, thanks for making it available!

dj2:
This CL adds the `fence_timeout` option to VkScript to control the value
passed for fence timeouts. The value must be an integer and is
interpreted as number of milliseconds.

Fixes #69.

### DIFF
--- a/docs/vk_script.md
+++ b/docs/vk_script.md
@@ -42,6 +42,9 @@ The _framebuffer_ and _depthstencil_ commands allow setting the format for the
 given buffer. The valid values are listed below in the *Image Formats*
 section.
 
+The _fence\_timeout_ option allows setting an integer number of milliseconds
+for any fence timeouts.
+
 The last option is _extensions_. Any string which isn't a _feature_,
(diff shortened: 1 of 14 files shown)


dj2:
@jrprice this should do what you want I believe.

dj2:
(Note, the 2017 bot is new and not working right yet ....)

jrprice:
Works for me - thanks!

PR #78, opened 2018-11-20
Dawn: Instantiate a Metal device.  Support SetShader

dneto0:


src/dawn/CMakeLists.txt
```diff
@@ -16,9 +16,19 @@ set(DAWN_ENGINE_SOURCES
     engine_dawn.cc
 )
 
+set(EXTRA_LIBS "")
```
dj2: Should this be AMBER_EXTRA_LIBS so we don't conflict with any other cmakefiles?

src/dawn/device_metal.h
```diff
@@ -0,0 +1,33 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef SRC_DAWN_DEVICE_METAL_H_
+#define SRC_DAWN_DEVICE_METAL_H_
+
+#if AMBER_DAWN_METAL
+
+#include "dawn/dawncpp.h"
```
dj2: nit: sort, this should be after amber/result.h

src/dawn/device_metal.mm
```diff
@@ -0,0 +1,51 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "device_metal.h"
+
+#include <Metal/Metal.h>
+#include <iostream>
```
dj2: nit: blank line between c and c++ headers

src/dawn/device_metal.mm
```diff
@@ -0,0 +1,51 @@
...
+
+#include "amber/result.h"
+#include "dawn/dawncpp.h"
+#include "dawn_native/DawnNative.h"
+#include "dawn_native/MetalBackend.h"
+
+namespace {
```
dj2: Any reason this isn't inside the amber::dawn namespace?

src/dawn/device_metal.mm
```diff
@@ -0,0 +1,51 @@
...
+void PrintDeviceError(const char* message, dawn::CallbackUserdata) {
+  std::cout << "Dawn device error: " << message << std::endl;
+}
+}  // namespace
+
+namespace amber {
+namespace dawn {
+
+amber::Result CreateMetalDevice(::dawn::Device* device_ptr) {
```
dj2: I don't think amber:: is needed here, just Result should work.

src/dawn/device_metal.mm
```diff
@@ -0,0 +1,51 @@
...
+  if (!device_ptr)
+    return Result("::amber::dawn::CreateMetalDevice: invalid device parameter");
+  dawnDevice cDevice =
+      ::dawn_native::metal::CreateDevice(MTLCreateSystemDefaultDevice());
+  if (!cDevice)
```
dj2: nit: add {}'s when the body is multiple lines.

src/dawn/device_metal.mm
```diff
@@ -0,0 +1,51 @@
...
+}
+}  // namespace
+
+namespace amber {
+namespace dawn {
+
+amber::Result CreateMetalDevice(::dawn::Device* device_ptr) {
+  if (!device_ptr)
+    return Result("::amber::dawn::CreateMetalDevice: invalid device parameter");
+  dawnDevice cDevice =
```
dj2: nit: add a blank line above (following the {}-less if)

src/dawn/engine_dawn.cc
```diff
@@ -15,6 +15,7 @@
 #include "src/dawn/engine_dawn.h"
 
 #include "dawn/dawncpp.h"
+#include "device_metal.h"
```
dj2: nit: dawn/device_metal.h

src/dawn/engine_dawn.cc
```diff
@@ -27,10 +28,14 @@ Result EngineDawn::Initialize() {
   if (device_)
     return Result("Dawn:Initialize device_ already exists");
 
-  // The constructor already gives us a default device.  So there is nothing
-  // left to do.
+#if AMBER_DAWN_METAL
+  Result result = CreateMetalDevice(&device_);
+  if (device_)
```
dj2: You should just be able to return result without doing the check. That way the device creation error will propagate instead of getting lost.

(which means, this can just be `return CreateMetalDevice(&device_);`

src/dawn/engine_dawn.cc
```diff
@@ -27,10 +28,14 @@ Result EngineDawn::Initialize() {
...
+    return result;
+  return Result("Dawn::Initialize: Device is null even after creation");
+#endif
```
dj2: nit: #endif  // AMBER_DAWN_META

src/dawn/engine_dawn.cc
```diff
@@ -58,8 +63,21 @@ Result EngineDawn::AddRequirement(Feature, const Format*) {
...
-Result EngineDawn::SetShader(ShaderType, const std::vector<uint32_t>&) {
-  return Result("Dawn:SetShader not implemented");
+Result EngineDawn::SetShader(ShaderType type,
+                             const std::vector<uint32_t>& code) {
+  ::dawn::ShaderModuleDescriptor descriptor;
+  descriptor.nextInChain = nullptr;
+  descriptor.code = code.data();
+  descriptor.codeSize = uint32_t(code.size());
+  if (!device_)
+    return Result("Dawn::SetShader: device is not created");
```
dj2: nit: blank line (and after 76)

src/dawn/engine_dawn.h
```diff
@@ -58,6 +59,8 @@ class EngineDawn : public Engine {
 
  private:
   ::dawn::Device device_;
+  enum { kNumShaderTypes = unsigned(ShaderType::kTessellationEvaluation) + 1 };
+  ::dawn::ShaderModule shader_[kNumShaderTypes];
```
dj2: Is this just because it's faster then map<ShaderType, ::dawn::ShaderModule> ?

src/dawn/device_metal.mm
dneto0: It's not visible externally, so put it in anonymous namespace.  So it doesn't pollute some other independent definition someone might dream up some day.

src/dawn/engine_dawn.cc
dneto0: Right.  This is a vestige of my debugging.

src/dawn/engine_dawn.h
dneto0: It hurts my head less.

src/dawn/engine_dawn.cc
dneto0: It's src/dawn/device_metal.h from this directory

src/dawn/device_metal.mm
dj2: Sorry, I ment why isn't the anonymous namespace inside the amber::dawn namespace?

src/dawn/engine_dawn.h
dj2: My concern is that we've now coupled this to ShaderType never changing order and never getting a new type added after TessellationEvaluation.

src/dawn/device_metal.mm
dneto0: It doesn't matter what the enclosing namespace is. This material is only accessible within this translation unit.  I could have declared it static instead and it would be the same.

dneto0: Changed it to inside amber::dawn, but under protest. :-)

dj2: Thanks, I find it easier to read when you open a bunch of namespaces and then close them instead of opening and closing randomly.

PR #82, opened 2018-11-21
Fixup clear stencil value

### LINKED ISSUE #81: vkscript spec: stencil value is an integer in Vulkan and Dawn
The stencil value should be an unsigned integer. See https://www.khronos.org/registry/vulkan/specs/1.0-wsi_extensions/html/vkspec.html#VkClearDepthStencilValue 

Similarly, Dawn has stencil value as uint32_t:

            RenderPassDescriptorBuilder const& SetDepthStencilAttachmentClearValue(float clearDepth, uint32_t clearStencil) const;

docs/vk_script.md currently says stencil defaults to 0.0
I think it should default to 0 (an integer).

dj2:
The clear stencil value is a uint32_t and the clear depth value is a
float. This CL clarifies the expected values and fixes up the stencil
example to be the correct type.

Fixes #81.

### DIFF
--- a/docs/vk_script.md
+++ b/docs/vk_script.md
@@ -259,13 +259,13 @@ Sets the clear color. Defaults to (0, 0, 0, 0).
 ### Clear Depth
  * `clear depth _value_`
 
-Sets the depth clear value. Defaults to 1.0.
+Sets the depth clear value. The _value_ is a float and defaults to 1.0.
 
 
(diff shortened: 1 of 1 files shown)


dj2:
The parser already forces the value to be an integer: https://github.com/google/amber/blob/master/src/vkscript/command_parser.cc#L393 so just the documentation was wrong.

PR #84, opened 2018-11-21
More AmberScript docs

dj2:
Add BUFFER with a FORMAT. Shuffle some text around.


docs/amber_script.md
```diff
@@ -75,31 +80,33 @@ TODO(dneto): Support half-precision floating point.
...
 <value>+
 END
 
+// Defines a buffer which is filled with data as specified by the `initializer`.
 BUFFER <name> DATA_TYPE <type> SIZE <size_in_items> <initializer>
+
+// Creates a buffer which will store the given `FORMAT` of data. These
+// buffers are used as image and depth buffers in the `PIPELINE` commands.
+// The buffer will be sized based on the `RENDER_SIZE` of the `PIPELINE`.
+BUFFER <name> FORMAT <format_string>
```
dj2: Thoughts on using the <format_string> which is what vKScript takes, or changing DATA_TYPE to allow more complex things? Need to be able to represent things like D24_FLOAT_S8_UINT nicely, and R8G8B8A8_UINT_PACK32.

docs/amber_script.md
```diff
@@ -124,33 +130,30 @@ The following commands are all specified within the `PIPELINE` command. If you
 have multiple entry points for a given shader, you'd create multiple pipelines
 each with a different `ENTRY_POINT`.
 
-Bind the entry point to use for a given shader. The default entry point is main.
 
 ```
+  // Bind the entry point to use for a given shader. The default entry point is
+  // main.
```
dneto0: This doesn't quite work.  We can use the same shader module to hold different shader stages.
E.g. the vertex and fragment shaders could be in the same module.  Then this ENTRY_POINT command is ambiguous.

Better to modify the ATTACH command. (see below)

docs/amber_script.md
```diff
@@ -124,33 +130,30 @@ The following commands are all specified within the `PIPELINE` command. If you
...
 
-Shaders can be added into pipelines with the `ATTACH` call. Shaders may be
-attached to multiple pipelines at the same time.
-
 ```
-  ATTACH <name_of_vertex_shader>
-  ATTACH <name_of_fragment_shader>
+  // Shaders can be added into pipelines with the `ATTACH` call. Shaders may be
+  // attached to multiple pipelines at the same time.
+  ATTACH <name_of_shader>
```
dneto0: Suggest this:
```
# Assuming the module only has one entry point for that stage, then use that entry point.
# The <shader-type> must match the type used to create the module
ATTACH <shader-type> <name_of_shader_module> 
# Explicitly name the entry point 
ATTACH <shader-stage> <name_of_shader_module> ENTRY_POINT myEntry
```

docs/amber_script.md
```diff
@@ -218,26 +223,55 @@ value for `START_IDX` is 0. The default value for `COUNT` is the item count of
 vertex buffer minus the `START_IDX`.
 
 ```
+// Run the given |pipeline_name| which must be a `compute` pipeline. The
+// pipeline will be run with the given |x|, |y|, |z| values.
```
dneto0: VkScript isn't clear about what these are.  
We should match vkCmdDispatch, in which case these are the number of workgroups to run in the x,y, and z dimensions.

docs/amber_script.md
```diff
@@ -75,31 +80,33 @@ TODO(dneto): Support half-precision floating point.
...
-
-Fill the buffer with an increasing value from \<start> increasing by \<inc>.
-Floating point data uses floating point addition to generate increasting values.
-Likewise, integer data uses integer addition to generate increasing values.
 
-```
+// Fill the buffer with an increasing value from \<start> increasing by \<inc>.
+// Floating point data uses floating point addition to generate increasting
+// values. Likewise, integer data uses integer addition to generate increasing
+// values.
```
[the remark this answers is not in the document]
dj2: Right now, it doesn't because I wasn't sure how to make it make sense. SERIES_FROM 0 INC_BY 2 for a vec2, would you do <0, 2>, <4, 8> or <0, 0>, <2, 2>.

We do FILL but not SERIES_FROM until we can figure out semantics.

docs/amber_script.md
dj2: Fixed in the Shader CL. ENTRY_POINT is removed and added to ATTACH.

docs/amber_script.md
dj2: See shader Cl, we can move discussion there.

docs/amber_script.md
dj2: Done.

dneto0:
Whoops.  I merged the "multi" shader module case, causing this to have a conflict.

docs/amber_script.md
```diff
@@ -183,22 +183,24 @@ needed `DIMS`. If the `FORMAT` is missing it is defaulted to
 TODO(dsinclair): Sync the BufferTypes with the list of Vulkan Descriptor types.
 
 
-When adding a buffer binding, the `IDX` parameter is optional and will default
-to 0.
-
 ```
+  # Bind the buffer of the given |buffer_type| at the given descritor set
+  # and binding. The buffer will use a start index of 0.
```
dneto0: I was confused by the "start index" statement for a while.  But it's the starting array index when the binding is an arrayed resource.  See the dstArrayElement in https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VkWriteDescriptorSet

So no change, this works.

docs/amber_script.md
```diff
@@ -229,26 +231,56 @@ value for `START_IDX` is 0. The default value for `COUNT` is the item count of
 vertex buffer minus the `START_IDX`.
 
 ```
+# Run the given |pipeline_name| which must be a `compute` pipeline. The
+# pipeline will be run with the given workgroup |x|, |y|, |z| dimensions.
```
dneto0: Actually it's the number of workgroups in the x, y, and z dimensions.  See vkCmdDispatch.

The workgroup size is specified in the SPIR-V module by a LocalSize execution mode (or a variable that has specializable components.)

docs/amber_script.md
```diff
@@ -229,26 +231,56 @@ value for `START_IDX` is 0. The default value for `COUNT` is the item count of
...
+# data must be attached to the pipeline. A start index of 0 will be used
+# and a count of the number of elements in the vertex buffer.
 RUN <pipeline_name> DRAW_ARRAY AS <topology>
-RUN <pipeline_name> DRAW_ARRAY AS <topology> START_IDX <value>
-RUN <pipeline_name> DRAW_ARRAY AS <topology> START_IDX <value> COUNT <value>
-
+# Run the |pipeline_name| which must be a `graphics` pipeline. The vertex
+# data must be attached to the pipeline. A start index of |value| will be used
+# and a count of the number of elements in the vertex buffer.
+RUN <pipeline_name DRAW_ARRAY AS <topology> START_IDX <value>
```
dneto0: <pipeline_name> . (close the angle bracket)

docs/amber_script.md
```diff
@@ -229,26 +231,56 @@ value for `START_IDX` is 0. The default value for `COUNT` is the item count of
...
+# Run the |pipeline_name| which must be a `graphics` pipeline. The vertex
+# data must be attached to the pipeline. A start index of |value| will be used
+# and a count |count_value| will be used.
+RUN <pipeline> DRAW_ARRAY AS <topology> START_IDX <value> COUNT <count_value>
```
dneto0: <pipeline_name>

docs/amber_script.md
```diff
@@ -229,26 +231,56 @@ value for `START_IDX` is 0. The default value for `COUNT` is the item count of
...
+# Run the |pipeline_name| which must be a `graphics` pipeline. The vertex
+# data must be attached to the pipeline. A start index of 0 will be used
+# and a count of the number of elements in the vertex buffer.
 RUN <pipeline_name> DRAW_ARRAY AS <topology>
-RUN <pipeline_name> DRAW_ARRAY AS <topology> START_IDX <value>
-RUN <pipeline_name> DRAW_ARRAY AS <topology> START_IDX <value> COUNT <value>
-
+# Run the |pipeline_name| which must be a `graphics` pipeline. The vertex
+# data must be attached to the pipeline. A start index of |value| will be used
+# and a count of the number of elements in the vertex buffer.
```
dneto0: count of the number of _remaining_ elements of the vertex buffer.

or use the phrasing similar in to what you have in the indexed case.

docs/amber_script.md
```diff
@@ -229,26 +231,56 @@ value for `START_IDX` is 0. The default value for `COUNT` is the item count of
...
+# the count of |count_value| items will be processed.
 RUN <pipeline_name> DRAW_ARRAY INDEXED AS <topology> \
-  START_IDX <value> COUNT <value>
+  START_IDX <value> COUNT <count_value>
 ```
 
 ### Commands
 ```
+# Sets the clear colour to use for |pipeline| which must be a `graphics`
+# pipeline. The colours are integers from 0 - 255.
```
dneto0: what about alpha?

Also, the type of value is intimately tied to the format of the framebuffer.  Maybe add a TODO to revisit this?

docs/amber_script.md
```diff
@@ -265,9 +297,13 @@ CLEAR <pipeline>
...
 
 ```
+# Checks that |buffer_name| at |x|, |y| has the given |value|s when compared
+# with the given |comparator|.
 EXPECT <buffer_name> IDX <x> <y> <comparator> <value>+
 
-EXPECT <framebuffer_name> IDX <x_in_pixels> <y_in_pixels> \
+# Checks that |buffer_name| at |x|, |y| for |width|x|height| pixels has the
+# given |r|, |g|, |b| values. Each r, g, b value is an integer from 0-255.
+EXPECT <buffer_name> IDX <x_in_pixels> <y_in_pixels> \
```
dneto0: I guess we never probe alpha?  Add a TODO for that?

docs/amber_script.md
dneto0: Actually, the *Vulkan* call lets you update multiple descriptors in a row, and in that API call the index parameter is the start index.
But in AmberScript this call is binding a single entry.  So it's just an *index* instead of a start index.

PR #86, opened 2018-11-21
Allow for multi SPIR-V shaders

dj2:
This CL updates AmberScript to allow SPIR-V to provide multiple shaders
in a single SHADER block. This is done with a 'multi' shader type and an
extension to ATTACH to specify the shader TYPE.

### DIFF
--- a/docs/amber_script.md
+++ b/docs/amber_script.md
@@ -128,27 +128,25 @@ END
 
 ### Pipeline Content
 
-The following commands are all specified within the `PIPELINE` command. If you
-have multiple entry points for a given shader, you'd create multiple pipelines
-each with a different `ENTRY_POINT`.
-
-Bind the entry point to use for a given shader. The default entry point is main.
-
-```
-  ENTRY_POINT <shader_name> <entry_point_name>
-```
+The following commands are all specified within the `PIPELINE` command.
 
 Shaders can be added into pipelines with the `ATTACH` call. Shaders may be
 attached to multiple pipelines at the same time.
 
 ```
-  # The provided shader for ATTACH must _not_ be a 'multi' shader.
-  ATTACH <name_of_vertex_shader>
-  ATTACH <name_of_fragment_shader>
+  # Attach the shader provided by |name_of_shader| to the pipeline and set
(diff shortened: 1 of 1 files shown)


docs/amber_script.md
```diff
@@ -31,11 +31,19 @@ which is the characters `0x` followed by hexadecimal digits.
...
 
 The compute pipeline can only contain compute shaders. The graphics pipeline
 can not contain compute shaders, and must contain a vertex shader and a fragment
 shader.
 
+The provided `multi` shader can only be used with `SPIRV-ASM` and `SPIRV-HEX`
+and allows for providing multiple shaders in a single module (so the `vertex`
+and `fragment` shaders can be provided together.)
+
+Note, `SPIRV-ASM` and `SPIRV-HEX` can also be used with each of the other shader
```
dneto0: I don't like this rule.  We often have many compute shaders in the same module which are then distinguished by entry point name.

docs/amber_script.md
```diff
@@ -134,8 +142,13 @@ Shaders can be added into pipelines with the `ATTACH` call. Shaders may be
 attached to multiple pipelines at the same time.
 
 ```
+  # The provided shader for ATTACH must _not_ be a 'multi' shader.
```
dneto0: What happens with   compute shader from OpenCL-C?  Do we declare it as a 'compute' type but then we would also need to give the entry-point name.

docs/amber_script.md
dj2: This rule doesn't preclude that? You can have a multi shader with a compute, vertex and fragment. You can't can't have a vertex and compute if it isn't multi. But, you can have a compute SPIRV-ASM as well.

docs/amber_script.md
dj2: Ah, so you're saying we need to combine ENTRY_POINT and ATTACH, that makes sense.

docs/amber_script.md
dneto0: Ok.  This works.  It's different from my mental model but it resolves to correct and unambiguous behaviour.

PR #87, opened 2018-11-22
Add locations to the VERTEX_DATA commands.

dj2:
The vertex data needs to know which location to bind too. This CL adds
the IDX parameter to VERTEX_DATA and fixes up the vertex example.

### DIFF
--- a/docs/amber_script.md
+++ b/docs/amber_script.md
@@ -197,9 +197,10 @@ TODO(dsinclair): Sync the BufferTypes with the list of Vulkan Descriptor types.
 ```
 
 ```
-  # Attach a vertex data buffer to the pipeline
-  VERTEX_DATA <buffer_name>
-  # Attach an index data buffer to the pipeline
+  # Set |buffer_name| as the vertex data at location |val|.
+  VERTEX_DATA <buffer_name> IDX <val>
+
+  # Set |buffer_name| as the index data to use for `INDEXED` draw commands.
   INDEX_DATA <buffer_name>
 ```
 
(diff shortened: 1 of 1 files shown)


docs/amber_script.md
```diff
@@ -412,27 +415,53 @@ SHADER fragment kFragmentShader GLSL
...
+-1  1
+ 0  1
+# Bottom-right
+ 0  0
+ 1  0
+ 0  1
+ 1  1
+END
+
+BUFFER kColorData TYPE uint32 DATA
```
dj2: The split is kinda annoying, but not too bad. Otherwise we'd need to know the locations of the different parts of the buffer which would be wonky I think.

docs/amber_script.md
```diff
@@ -197,9 +197,10 @@ TODO(dsinclair): Sync the BufferTypes with the list of Vulkan Descriptor types.
 ```
 
 ```
-  # Attach a vertex data buffer to the pipeline
-  VERTEX_DATA <buffer_name>
-  # Attach an index data buffer to the pipeline
+  # Set |buffer_name| as the vertex data at location |val|.
```
dneto0: Is the IDX val the shader's input Location attribute for the vertex data.
(That is, this is for populating VkVertexInputAttributeDescription::location )

So I think it should use keyword LOCATION instead of IDX.

docs/amber_script.md
dneto0: Better to be explicit about the different locations.  So I agree with this change.

docs/amber_script.md
dj2: done

dj2: done

PR #88, opened 2018-11-22
Add copyright check

dj2:
This CL imports the copyright check from shaderc and adds it to the
check-format kokoro bot.

### DIFF
--- a/tools/copyright.py
+++ b/tools/copyright.py
@@ -1,0 +1,147 @@
+#!/usr/bin/env python
+# Copyright 2018 The Amber Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


tools/copyright.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env python
+# Copyright 2018 The Amber Authors.
```
dj2: Switched this from Shaderc to Amber, given we have the copyright on both I figured that was OK?

dneto0: Yes, all contributions to that script in the Shaderc repo were from Googlers, so Google owns the copyright to the original contributions.

tools/copyright.py
```diff
@@ -0,0 +1,147 @@
...
+                    break
+        if not has_copyright:
+            print(file, ' has no copyright message.')
+            printed_count += 1
+    return printed_count
+
+
+def main():
+    glob_comment_pairs = [('*.h', '//'), ('*.hpp', '//'), ('*.cc', '//'),
+                          ('*.py', '#'), ('*.cpp', '//')]
```
dneto0: I'd add .mm to this list.  :-)

dj2: done

dneto0:
The check-format failure is because there was no code to format.  Spurious failure.

PR #89, opened 2018-11-22
Framebuffer updates

dj2:
This Cl removes the FRAMEBUFFER commands from AmberScript and converts
them into more generic buffer commands.

### DIFF
--- a/docs/amber_script.md
+++ b/docs/amber_script.md
@@ -159,18 +159,13 @@ The following commands are all specified within the `PIPELINE` command.
   END
 ```
 
-#### Bindings
-
-Bind a provided framebuffer. The third example `FRAMEBUFFER <name>` requires
-that `<name>` was declared in a previous `PIPELINE` and provided with the
-needed `DIMS`. If the `FORMAT` is missing it is defaulted to
-`R32G32B32A32_UINT`.
-
 ```
-  FRAMEBUFFER <name> DIMS <width> <height>
-  FRAMEBUFFER <name> DIMS <width> <height> FORMAT <fb_format>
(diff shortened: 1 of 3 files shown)


docs/amber_script.md
```diff
@@ -181,10 +176,22 @@ needed `DIMS`. If the `FORMAT` is missing it is defaulted to
 
 TODO(dsinclair): Sync the BufferTypes with the list of Vulkan Descriptor types.
 
+A `pipeline` can have buffers bound. This includes buffers to contain image
+attachment content, depth/stencil content, uniform buffers, etc.
 
 ```
-  # Bind the buffer of the given |buffer_type| at the given descritor set
-  # and binding. The buffer will use a index of 0.
+  # Attach |buffer_name| as an image attachment at slot |idx|. The provided
```
dneto0: I think this is for a colour output attachment?  I'd be more explicit about it.  (In HLSL these are SV_Targetn where n is 0 through 7, and target it short for "render target". https://docs.microsoft.com/en-us/windows/desktop/direct3dhlsl/dx-graphics-hlsl-semantics )

And "slot" and "idx" are not quite right.

Recommend changing to:
> Attach |buffer_name| as an output colour attachment at location index |loc|. ...

> BIND BUFFER <buffer_name> AS target LOCATION <idx>

I'm not 100% sold on the name "target" here.  I think it's better than "image".  Alternates: color, render, render_target (too long!).

Vulkan also has subpass input, for tiled rendering.   I'd add a TODO to reserve 'input', as in:

>   BUFFER <buffer_name> AS input IDX <idx>

dj2: done

PR #90, opened 2018-11-22
Add cpplint

dj2:
This CL adds CPP lint and runs it during the check-format bot run.

### DIFF
--- a/src/tokenizer_test.cc
+++ b/src/tokenizer_test.cc
@@ -12,8 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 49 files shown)


dj2:
Testing to see if the build fails since I still have to fix the lint checks ....

src/tokenizer_test.cc
```diff
@@ -12,8 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "src/tokenizer.h"
```
dneto0: I think the old code does it in the way that the style guide requests: Test code for a header should have its include first.

PR #91, opened 2018-11-22
Skip clang-format-diff script

dj2:


dj2:
TBR=@dneto

Landing this to unblock other CLs. The check-format passes now.

PR #92, opened 2018-11-22
Move probe code to a common verifier class.

dj2:
This CL removes the Probe calls from the engines and adds a verifier
which does the comparisons.

### DIFF
--- a/src/engine.h
+++ b/src/engine.h
@@ -97,17 +97,14 @@ class Engine {
   virtual Result DoPatchParameterVertices(
       const PatchParameterVerticesCommand* cmd) = 0;
 
-  // Execute the probe command
-  virtual Result DoProbe(const ProbeCommand* cmd) = 0;
-
-  // Execute the probe ssbo command
-  virtual Result DoProbeSSBO(const ProbeSSBOCommand* cmd) = 0;
-
   // Execute the buffer command
   virtual Result DoBuffer(const BufferCommand* cmd) = 0;
 
-  // Execute the tolerance command
-  virtual Result DoTolerance(const ToleranceCommand* cmd) = 0;
+  // Run all queued commands and copy result data to the host.
+  virtual Result DoProcessCommands(uint32_t* stride,
+                                   uint32_t* width,
+                                   uint32_t* height,
+                                   const void** buf) = 0;
 
  protected:
   Engine();
--- a/src/verifier.cc
+++ b/src/verifier.cc
@@ -1,0 +1,124 @@
+// Copyright 2018 The Amber Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "src/verifier.h"
+
+#include <cmath>
+
+namespace amber {
+namespace {
+
+const float kEpsilon = 0.002f;
+
+bool IsFloatPixelEqualInt(float pixel, uint8_t expected) {
+  // TODO(jaebaek): Change kEpsilon to tolerance.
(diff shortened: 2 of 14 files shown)


dj2:
There are some tests which are DISABLED, I need to determine if they're still useful or should just be removed.

src/engine.h
```diff
@@ -97,17 +97,14 @@ class Engine {
...
-  // Execute the probe ssbo command
-  virtual Result DoProbeSSBO(const ProbeSSBOCommand* cmd) = 0;
-
   // Execute the buffer command
   virtual Result DoBuffer(const BufferCommand* cmd) = 0;
 
-  // Execute the tolerance command
-  virtual Result DoTolerance(const ToleranceCommand* cmd) = 0;
+  // Run all queued commands and copy result data to the host.
+  virtual Result DoProcessCommands(uint32_t* stride,
```
dj2: This is yucky but need to figure out a better way to do it. I need to get all the resources to handle the probe, which in this case is the size, stride and the bits to compare.

dj2: The solution to this maybe to have VkScript use a FormatBuffer we haven't created yet. That format buffer will then have the bits filled in by the engine. It knows it's stride and size so you should then have all the things.

For VkScript we just can only have one output buffer, with AmberScript the EXPECT command tells the buffer to expect against.

This would be a followup change to implement FormatBuffer and use.

src/verifier.cc
```diff
@@ -0,0 +1,124 @@
...
+#include <cmath>
+
+namespace amber {
+namespace {
+
+const float kEpsilon = 0.002f;
+
+bool IsFloatPixelEqualInt(float pixel, uint8_t expected) {
+  // TODO(jaebaek): Change kEpsilon to tolerance.
+  return std::fabs(pixel - static_cast<float>(expected) / 255.0f) < kEpsilon;
```
dneto0: The 255 here is very specific to the assumed framebuffer format.  

As a first step I'd make it a declared constant (thus giving it a name).   

This also assumes pixel values are 0-based.   
Again, this is a dependency on the assumed target format.
There are signed integer format, e.g. VK_FORMAT_R8G8B8A8_SINT.  So I'd add a TODO to generalze this.

For now, make a named constant for the 255.0

src/engine.h
dneto0: Right for VkScript's single-output model this works.

For future we'll need another model that takes into account lots of outputs, in an unstructured way.
What comes to mind:


- Modify DoBuffer to take the buffer name as an argument.
- DoProcessWork : runs queued work, and gets back data for all buffers we've declared
- After DoProcessWork we can call GetBufferData(const std::string& buffer_name, const void** buf); 
      //// and whatever else is needed to be able to scan it out.  But I think the executor has the format and stride info?
    - An empty string for the buffer name will refer to a buffer the engine created for the framebufer because there was no explicitly declared buffer.   (Actually, maybe the executor takes responsibility for creating the implied buffer for the framebuffer?)

src/verifier.cc
dneto0: This was probably there before, so I won't block this MR for this issue.

src/verifier.cc
```diff
@@ -0,0 +1,124 @@
...
+  } else {
+    x = static_cast<uint32_t>(command->GetX());
+    y = static_cast<uint32_t>(command->GetY());
+    width = static_cast<uint32_t>(command->GetWidth());
+    height = static_cast<uint32_t>(command->GetHeight());
+  }
+
+  if (x + width > frame_width || y + height > frame_height) {
+    return Result(
+        "Vulkan::Probe Position(" + std::to_string(x + width - 1) + ", " +
```
dneto0: Will need to fix names soon.  I'll likely do that in a followup.

src/vkscript/executor.cc
```diff
@@ -99,7 +99,22 @@ Result Executor::Execute(Engine* engine, const amber::Script* src_script) {
...
+        if (!r.IsSuccess())
+          return r;
+
+        r = verifier_.Probe(cmd->AsProbe(), stride, width, height, buf);
+      } else if (cmd->IsProbeSSBO()) {
+        r = verifier_.ProbeSSBO(cmd->AsProbeSSBO());
+      } else if (cmd->IsTolerance()) {
+        r = verifier_.Tolerance(cmd->AsTolerance());
+
+      } else if (cmd->IsClear()) {
```
dneto0: Cool. This will let me progress the Dawn engine with one more easy step. :-)

PR #93, opened 2018-11-22
Dawn: Support DoClear (by queueing the work)

dneto0:


src/dawn/engine_dawn_test.cc
```diff
@@ -0,0 +1,44 @@
...
+namespace dawn {
+
+namespace {
+
+using EngineDawnTest = testing::Test;
+
+TEST_F(EngineDawnTest, PendingWorkDefaultIsEmpty) {
+  EngineDawn e;
+  auto& pending = e.GetPendingWorkForTest();
+  EXPECT_EQ(0, pending.size());
```
dj2: EXPECT_TRUE(pending.empty());

src/dawn/engine_dawn_test.cc
```diff
@@ -0,0 +1,44 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "src/dawn/engine_dawn.h"
+
+#include "gmock/gmock.h"
```
dj2: gtest/gtest.h? Since you aren't mocking anything?

dneto0: Will fix.  I reflexively use gmock.h because I usually use EXPECT_THAT with the more powerful predicates.

dneto0:
I'm changing strategy:  Instead of queueing work, I'm going to execute them immediately (actually by adding to the command buffer)

Issue #32, opened 2018-11-14
Vulkan: Implement PROBE for SSBO and general images.


dj2:

jaebaek:
#123 handles SSBO. Let me close this and open a new issue for image resource.

jaebaek:
#142 is the issue for image resource

PR #101, opened 2018-11-27
Vulkan: handle probe ssbo command

jaebaek:
Related to #32

### DIFF
--- a/src/vulkan/buffer.cc
+++ b/src/vulkan/buffer.cc
@@ -36,6 +36,8 @@ Result Buffer::Initialize(const VkBufferUsageFlags usage) {
 
   if (CheckMemoryHostAccessible(allocate_result.memory_type_index)) {
     is_buffer_host_accessible_ = true;
+    is_buffer_host_coherent_ =
+        CheckMemoryHostCoherent(allocate_result.memory_type_index);
     return MapMemory(memory_);
   }
 
@@ -65,10 +67,25 @@ void Buffer::CopyToDevice(VkCommandBuffer command) {
   VkBufferCopy region = {};
   region.srcOffset = 0;
   region.dstOffset = 0;
-  region.size = GetSize();
+  region.size = GetSizeInBytes();
 
   vkCmdCopyBuffer(command, GetHostAccessibleBuffer(), buffer_, 1, &region);
   MemoryBarrier(command);
+}
+
+Result Buffer::CopyToHost(VkCommandBuffer command) {
+  if (is_buffer_host_accessible_) {
+    return FlushAndInvalidateMemoryIfNeeded();
+  }
+
+  VkBufferCopy region = {};
+  region.srcOffset = 0;
+  region.dstOffset = 0;
+  region.size = GetSizeInBytes();
+
+  vkCmdCopyBuffer(command, buffer_, GetHostAccessibleBuffer(), 1, &region);
+  MemoryBarrier(command);
+  return {};
 }
 
 void Buffer::Shutdown() {
@@ -88,6 +105,24 @@ void Buffer::Shutdown() {
     vkFreeMemory(GetDevice(), memory_, nullptr);
     memory_ = VK_NULL_HANDLE;
   }
+}
+
+Result Buffer::FlushAndInvalidateMemoryIfNeeded() {
+  if (is_buffer_host_coherent_)
+    return {};
+
+  VkMappedMemoryRange range = {};
+  range.sType = VK_STRUCTURE_TYPE_MAPPED_MEMORY_RANGE;
+  range.memory = GetHostAccessMemory();
+  range.offset = 0;
+  range.size = VK_WHOLE_SIZE;
+  if (vkFlushMappedMemoryRanges(GetDevice(), 1, &range) != VK_SUCCESS)
+    return Result("Vulkan: vkFlushMappedMemoryRanges fail");
+
+  if (vkInvalidateMappedMemoryRanges(GetDevice(), 1, &range) != VK_SUCCESS)
+    return Result("Vulkan: vkInvalidateMappedMemoryRanges fail");
+
+  return {};
 }
 
 }  // namespace vulkan
--- a/src/vulkan/buffer.cc
+++ b/src/vulkan/buffer.cc
@@ -34,10 +34,10 @@ Result Buffer::Initialize(const VkBufferUsageFlags usage) {
   if (!allocate_result.r.IsSuccess())
     return allocate_result.r;
 
-  if (CheckMemoryHostAccessible(allocate_result.memory_type_index)) {
+  if (IsMemoryHostAccessible(allocate_result.memory_type_index)) {
     is_buffer_host_accessible_ = true;
     is_buffer_host_coherent_ =
-        CheckMemoryHostCoherent(allocate_result.memory_type_index);
+        IsMemoryHostCoherent(allocate_result.memory_type_index);
     return MapMemory(memory_);
   }
 
@@ -60,9 +60,9 @@ Result Buffer::CreateVkBufferView(VkFormat format) {
   return {};
 }
 
-void Buffer::CopyToDevice(VkCommandBuffer command) {
+Result Buffer::CopyToDevice(VkCommandBuffer command) {
   if (is_buffer_host_accessible_)
-    return;
+    return FlushMemoryIfNeeded();
 
   VkBufferCopy region = {};
   region.srcOffset = 0;
@@ -71,12 +71,12 @@ void Buffer::CopyToDevice(VkCommandBuffer command) {
 
   vkCmdCopyBuffer(command, GetHostAccessibleBuffer(), buffer_, 1, &region);
   MemoryBarrier(command);
+  return {};
 }
 
 Result Buffer::CopyToHost(VkCommandBuffer command) {
-  if (is_buffer_host_accessible_) {
-    return FlushAndInvalidateMemoryIfNeeded();
-  }
+  if (is_buffer_host_accessible_)
+    return InvalidateMemoryIfNeeded();
 
   VkBufferCopy region = {};
   region.srcOffset = 0;
@@ -107,7 +107,22 @@ void Buffer::Shutdown() {
   }
 }
 
-Result Buffer::FlushAndInvalidateMemoryIfNeeded() {
+Result Buffer::InvalidateMemoryIfNeeded() {
+  if (is_buffer_host_coherent_)
+    return {};
+
+  VkMappedMemoryRange range = {};
+  range.sType = VK_STRUCTURE_TYPE_MAPPED_MEMORY_RANGE;
+  range.memory = GetHostAccessMemory();
+  range.offset = 0;
+  range.size = VK_WHOLE_SIZE;
+  if (vkInvalidateMappedMemoryRanges(GetDevice(), 1, &range) != VK_SUCCESS)
+    return Result("Vulkan: vkInvalidateMappedMemoryRanges fail");
+
+  return {};
+}
+
+Result Buffer::FlushMemoryIfNeeded() {
   if (is_buffer_host_coherent_)
     return {};
 
@@ -118,9 +133,6 @@ Result Buffer::FlushAndInvalidateMemoryIfNeeded() {
   range.size = VK_WHOLE_SIZE;
   if (vkFlushMappedMemoryRanges(GetDevice(), 1, &range) != VK_SUCCESS)
     return Result("Vulkan: vkFlushMappedMemoryRanges fail");
-
-  if (vkInvalidateMappedMemoryRanges(GetDevice(), 1, &range) != VK_SUCCESS)
-    return Result("Vulkan: vkInvalidateMappedMemoryRanges fail");
 
   return {};
 }
--- a/src/vulkan/buffer.h
+++ b/src/vulkan/buffer.h
@@ -34,7 +34,10 @@ class Buffer : public Resource {
   Result CreateVkBufferView(VkFormat format);
   VkBufferView GetVkBufferView() const { return view_; }
 
-  // TODO(jaebaek): Determine copy all or partial data
+  // If this buffer is host accessible, this method does nothing.
+  // Otherwise, this method only records the command for copying
+  // this buffer to the host accessible buffer. The actual
+  // submission of the command must be done later.
   void CopyToDevice(VkCommandBuffer command);
 
   // Resource
@@ -45,13 +48,24 @@ class Buffer : public Resource {
     return Resource::GetHostAccessMemory();
   }
 
+  // If this buffer is host accessible and has coherent memory, this
+  // method does nothing. If this buffer is host accessible but it does
+  // not have coherent memory, this method flush and invalidate it.
+  // Otherwise, this method only records the command for copying this
+  // buffer to its secondary host-accessible buffer. The actual
+  // submission of the command must be done later.
+  Result CopyToHost(VkCommandBuffer command) override;
+
   void Shutdown() override;
 
  private:
+  Result FlushAndInvalidateMemoryIfNeeded();
+
   VkBuffer buffer_ = VK_NULL_HANDLE;
   VkBufferView view_ = VK_NULL_HANDLE;
   VkDeviceMemory memory_ = VK_NULL_HANDLE;
   bool is_buffer_host_accessible_ = false;
+  bool is_buffer_host_coherent_ = false;
 };
 
 }  // namespace vulkan
--- a/src/vulkan/buffer.h
+++ b/src/vulkan/buffer.h
@@ -34,11 +34,14 @@ class Buffer : public Resource {
   Result CreateVkBufferView(VkFormat format);
   VkBufferView GetVkBufferView() const { return view_; }
 
-  // If this buffer is host accessible, this method does nothing.
-  // Otherwise, this method only records the command for copying
-  // this buffer to the host accessible buffer. The actual
-  // submission of the command must be done later.
-  void CopyToDevice(VkCommandBuffer command);
+  // If this buffer is host accessible and has coherent memory, this
+  // method does nothing. If this buffer is host accessible but it does
+  // not have coherent memory, this method flush the memory to make the
+  // writes from host effective to device. Otherwise, this method records
+  // the command for copying the secondary host-accessible buffer to
+  // this device local buffer. Note that it only records the command and
+  // the actual submission must be done later.
+  Result CopyToDevice(VkCommandBuffer command);
 
   // Resource
   VkDeviceMemory GetHostAccessMemory() const override {
@@ -50,16 +53,18 @@ class Buffer : public Resource {
 
   // If this buffer is host accessible and has coherent memory, this
   // method does nothing. If this buffer is host accessible but it does
-  // not have coherent memory, this method flush and invalidate it.
-  // Otherwise, this method only records the command for copying this
-  // buffer to its secondary host-accessible buffer. The actual
-  // submission of the command must be done later.
+  // not have coherent memory, this method invalidate the memory to make the
+  // writes from device visible to host. Otherwise, this method records
+  // the command for copying this buffer to its secondary host-accessible
+  // buffer. Note that it only records the command and the actual
+  // submission must be done later.
   Result CopyToHost(VkCommandBuffer command) override;
 
   void Shutdown() override;
 
  private:
-  Result FlushAndInvalidateMemoryIfNeeded();
+  Result InvalidateMemoryIfNeeded();
+  Result FlushMemoryIfNeeded();
 
   VkBuffer buffer_ = VK_NULL_HANDLE;
   VkBufferView view_ = VK_NULL_HANDLE;
--- a/src/vulkan/descriptor.h
+++ b/src/vulkan/descriptor.h
@@ -18,6 +18,7 @@
 #include <memory>
 
 #include "amber/result.h"
+#include "src/engine.h"
 #include "vulkan/vulkan.h"
 
 namespace amber {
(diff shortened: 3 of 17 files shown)


src/engine.h
```diff
@@ -33,6 +33,30 @@ enum class PipelineType : uint8_t {
...
 
+enum class ResourceInfoType : uint8_t {
+  kBuffer = 0,
+  kImage,
+};
+
+struct ResourceInfo {
+  ResourceInfoType type;
+
+  union {
```
dj2: Let's not use a union here.

src/verifier.cc
```diff
@@ -15,15 +15,143 @@
...
+}
+
+template <typename T>
+Result CheckIntValue(ProbeSSBOCommand::Comparator comp,
+                     const uint8_t* memory,
+                     const std::vector<Value>& values) {
+  const T* ptr = reinterpret_cast<const T*>(memory);
+  switch (comp) {
+    case ProbeSSBOCommand::Comparator::kEqual:
+      for (auto v : values) {
```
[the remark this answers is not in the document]
dj2: Move the for() outside the switch, copy the value into an const T val = static_cast<T>(v.AsUint64); should make this easier to read

src/verifier.cc
```diff
@@ -15,15 +15,143 @@
...
+        if (*ptr != static_cast<T>(v.AsUint64()))
+          return Result("Verifier::ProbeSSBO failed");
+        ++ptr;
+      }
+      break;
+    case ProbeSSBOCommand::Comparator::kNotEqual:
+      for (auto v : values) {
+        if (*ptr == static_cast<T>(v.AsUint64()))
+          return Result("Verifier::ProbeSSBO failed");
```
dj2: These results should have more context:

Result("Verifier failed: " + std::to_string(*ptr) + " != " + std::to_string(val));

src/verifier.cc
```diff
@@ -15,15 +15,143 @@
...
+template <typename T>
+Result CheckFloatValue(ProbeSSBOCommand::Comparator comp,
+                       const uint8_t* memory,
+                       const std::vector<Value>& values) {
+  const T* ptr = reinterpret_cast<const T*>(memory);
+  switch (comp) {
+    case ProbeSSBOCommand::Comparator::kEqual:
+      for (auto v : values) {
+        if (*ptr != static_cast<T>(v.AsDouble()))
+          return Result("Verifier::ProbeSSBO failed");
```
dj2: Same comments as above

src/verifier.cc
```diff
@@ -15,15 +15,143 @@
...
+      break;
+    default:
+      return Result("Verifier::ProbeSSBO unknown comparator");
+  }
+
+  return {};
+}
+
+template <typename T>
+Result CheckFloatValue(ProbeSSBOCommand::Comparator comp,
```
dj2: This is almost identical to the above they can probably be combined just have to change the
static_cast<T>(v.AsDouble) to something like static_cast<T>(v.IsInteger() ? v.AsUint64() : v.AsDouble()).

src/verifier.cc
```diff
@@ -73,11 +201,15 @@ Result Verifier::Probe(const ProbeCommand* command,
     const uint8_t* p = ptr + stride * frame_width * (j + y) + stride * x;
     for (uint32_t i = 0; i < width; ++i) {
       // TODO(jaebaek): Get actual pixel values based on frame buffer formats.
-      if (!IsFloatPixelEqualInt(command->GetR(), p[stride * i]) ||
-          !IsFloatPixelEqualInt(command->GetG(), p[stride * i + 1]) ||
-          !IsFloatPixelEqualInt(command->GetB(), p[stride * i + 2]) ||
+      if (!IsFuzzyEqual(static_cast<double>(command->GetR()),
+                        static_cast<double>(p[stride * i]) / 255.0) ||
```
dj2: Why move the / 255.0 out here instead of leaving it in the IsFuzzyEqual?

src/verifier.cc
```diff
@@ -113,8 +245,43 @@ Result Verifier::Probe(const ProbeCommand* command,
   return {};
 }
 
-Result Verifier::ProbeSSBO(const ProbeSSBOCommand*) {
-  return Result("Verifier::ProbeSSBO Not Implemented");
+Result Verifier::ProbeSSBO(const ProbeSSBOCommand* command,
+                           size_t size,
```
dj2: Instead of passing a size and a const void*, passing a std::vector<uint8_t>* That will give you both data and size and it will get auto-cleanedup

src/vulkan/compute_pipeline.h
```diff
@@ -37,6 +37,9 @@ class ComputePipeline : public Pipeline {
 
   Result Compute(uint32_t x, uint32_t y, uint32_t z);
 
+  // Pipeline
```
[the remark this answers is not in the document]
dj2: This is chromium style, put the name of the class you're overriding as a comment before the overrides.

src/vulkan/descriptor.h
```diff
@@ -87,6 +90,8 @@ class Descriptor {
 
   virtual Result UpdateDescriptorSet(VkDescriptorSet descriptor_set) = 0;
   virtual void SendDataToGPUIfNeeded(VkCommandBuffer command) = 0;
```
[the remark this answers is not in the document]
dj2: So, this was me, heh. I find Host/Device confusing so asked for CPU/GPU to make it clear where the memory was/is.

src/vulkan/image.cc
```diff
@@ -122,9 +122,9 @@ void Image::Shutdown() {
   memory_ = VK_NULL_HANDLE;
 }
 
-Result Image::CopyToHost(VkCommandBuffer command) {
+void Image::CopyToCPU(VkCommandBuffer command) {
   if (is_image_host_accessible_)
-    return {};
+    return;
```
dj2: Why isn't this an error if we can't access the image, something has gone wrong, no?

src/vulkan/graphics_pipeline.cc
```diff
@@ -514,9 +514,7 @@ Result GraphicsPipeline::ProcessCommands() {
   ActivateRenderPassIfNeeded();
   DeactivateRenderPassIfNeeded();
 
-  r = frame_->CopyColorImageToHost(command_->GetCommandBuffer());
-  if (!r.IsSuccess())
-    return r;
+  frame_->CopyColorImageToHost(command_->GetCommandBuffer());
```
dj2: What happens if the copy fails?

src/vulkan/graphics_pipeline.cc
```diff
@@ -528,10 +526,6 @@ Result GraphicsPipeline::ProcessCommands() {
 void GraphicsPipeline::Shutdown() {
   DeactivateRenderPassIfNeeded();
 
-  Result r = command_->End();
```
dj2: What happens to any commands we've Begin, but not End'd?

src/vulkan/image.cc
```diff
@@ -122,9 +122,9 @@ void Image::Shutdown() {
   memory_ = VK_NULL_HANDLE;
 }
 
-Result Image::CopyToHost(VkCommandBuffer command) {
+void Image::CopyToCPU(VkCommandBuffer command) {
```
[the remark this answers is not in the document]
dj2: I tend to think in terms of where is the memory accessible, We're making it available to the CPU

dneto0:
Bots are failing too.

src/vulkan/pipeline.cc
```diff
@@ -36,8 +37,8 @@ Pipeline::Pipeline(
     : device_(device),
       memory_properties_(properties),
       pipeline_type_(type),
-      fence_timeout_ms_(fence_timeout_ms),
-      shader_stage_info_(shader_stage_info) {}
+      shader_stage_info_(shader_stage_info),
+      fence_timeout_ms_(fence_timeout_ms) {}
```
dj2: Why flip the order?

src/vulkan/pipeline.cc
```diff
@@ -287,5 +288,40 @@ void Pipeline::BindVkPipeline() {
...
+
+  return Result("Vulkan::Pipeline descriptor with descriptor set: " +
+                std::to_string(descriptor_set) +
+                ", binding: " + std::to_string(binding) + " does not exist");
+}
+
+Result Pipeline::GetDescriptorInfo(const uint32_t descriptor_set,
+                                   const uint32_t binding,
+                                   ResourceInfo* info) {
+  for (size_t i = 0; i < descriptors_.size(); ++i) {
```
dj2: assert(info)

src/vulkan/descriptor.h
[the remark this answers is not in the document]
dj2: Chatted with David about this and I was wrong. These should be Host/Device because there are cases where Host/Device is CPU/GPU and cases where it's CPU/CPU. So Host/Device is the more unique and correct naming.

Sorry about that.

src/verifier.cc
jaebaek: Sorry for making you confused. `size` here means the size of the ssbo in bytes. I don't know the number of elements at this moment. `cpu_memory` is just a pointer to ssbo. In my personal opinion, we would better not make a vector for this.

src/verifier.cc
```diff
@@ -15,15 +15,73 @@
...
+        break;
+      case ProbeSSBOCommand::Comparator::kLessOrEqual:
+        if (*ptr > val)
+          return Result("Verifier failed: " + std::to_string(*ptr) + " > " +
+                        std::to_string(val));
+        break;
+      case ProbeSSBOCommand::Comparator::kGreater:
+        if (*ptr <= val)
+          return Result("Verifier failed: " + std::to_string(*ptr) +
+                        " <= " + std::to_string(val));
```
dj2: For these error messages, I think we should use the comparator not the comparison used. So, this should be std::to_string(*ptr) + " > " + std::to_string(val); because that's the check in the script which failed.

src/verifier.cc
```diff
@@ -15,15 +15,73 @@
...
+                                : static_cast<T>(v.AsDouble());
+
+    switch (comp) {
+      case ProbeSSBOCommand::Comparator::kEqual:
+        if (*ptr != val)
+          return Result("Verifier failed: " + std::to_string(*ptr) +
+                        " != " + std::to_string(val));
+        break;
+      case ProbeSSBOCommand::Comparator::kNotEqual:
+        if (*ptr == val)
```
dj2: This and the != case probably need to be a fuzzy equal

src/vulkan/image.cc
```diff
@@ -144,7 +140,7 @@ Result Image::CopyToHost(VkCommandBuffer command) {
                          GetHostAccessibleBuffer(), 1, &copy_region);
 
   MemoryBarrier(command);
-  return {};
+  return;
```
dj2: nit: return at end of method not needed.

dneto0:
Please rebase.

src/vulkan/image.cc
jaebaek: done

src/verifier.cc

src/dawn/pipeline_info.h
```diff
@@ -50,6 +50,7 @@ struct RenderPipelineInfo {
   // The number of data bytes in the framebuffer host-side buffer.
   uint32_t fb_size = 0;
   bool fb_is_mapped = false;
+  const void* fb_data = false;
```
dj2: This should be nullptr I think.

src/verifier.cc
```diff
@@ -15,15 +15,95 @@
...
 
-const float kEpsilon = 0.002f;
+const double kEpsilon = 0.000001;
+const double kTolerance = 0.002;
 
-bool IsFloatPixelEqualInt(float pixel, uint8_t expected) {
+bool IsFuzzyEqual(const double real,
+                  const double expected,
+                  const double tolerance = kTolerance) {
   // TODO(jaebaek): Change kEpsilon to tolerance.
```
dj2: This comment seems invalid now?

src/dawn/pipeline_info.h
jaebaek: done

src/verifier.cc

src/dawn/pipeline_info.h
```diff
@@ -50,6 +50,7 @@ struct RenderPipelineInfo {
   // The number of data bytes in the framebuffer host-side buffer.
   uint32_t fb_size = 0;
   bool fb_is_mapped = false;
+  const void* fb_data = nullptr;
```
dneto0: Every time fb_is_mapped is set, the fb_data member should be set.  For example where the buffer is unmapped.

src/dawn/engine_dawn.cc
```diff
@@ -443,5 +430,28 @@ Result EngineDawn::CreateFramebufferIfNeeded() {
...
+
+  if (!render_pipeline_info_.fb_is_mapped)
+    return Result("Dawn: FrameBuffer is not mapped");
+
+  // TODO(dneto): Need to pass back a row_stride.
+  info->image_info.texel_stride = render_pipeline_info_.fb_texel_stride;
+  info->image_info.width = kFramebufferWidth;
+  info->image_info.height = kFramebufferHeight;
+  info->image_info.depth = 1U;
+  info->image_info.size_in_bytes = render_pipeline_info_.fb_texel_stride *
```
dneto0: This member doesn't exist in ResourceInfo.  I think you didn't compile this code.  :-)

src/engine.h
```diff
@@ -104,11 +132,20 @@ class Engine {
...
-  virtual Result DoProcessCommands(uint32_t* stride,
-                                   uint32_t* width,
-                                   uint32_t* height,
-                                   const void** buf) = 0;
+  // Run all queued commands and copy frame buffer data to the host
+  // if graphics pipeline.
+  virtual Result DoProcessCommands() = 0;
+
+  // Get stride, width, height, and memory pointer of frame buffer.
+  virtual Result GetFrameBufferInfo(ResourceInfo* info) = 0;
```
dneto0: Should say this is only valid if the framebuffer buffer is mapped into the host address space. In particular, if we have run DoProcessCommands() and that since then no graphics pipeline drawing commands have occurred (such as DoClear, DoDrawArrays, DoDrawRect).

src/dawn/pipeline_info.h
dneto0: I had another thought.   I want fb_data to be nullptr if and only if the buffer is *unmapped*.   But that makes fb_is_mapped redundant because it encodes the same information.
So: remove the fb_is_mapped variable entirely, and only set and check fb_data.

src/verifier.cc
```diff
@@ -15,15 +15,80 @@
...
+#include "src/command.h"
 
 namespace amber {
 namespace {
 
-const float kEpsilon = 0.002f;
+const double kEpsilon = 0.000001;
+const double kTolerance = 0.002;
+
+bool IsFuzzyEqual(const double real,
```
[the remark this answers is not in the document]
dneto0: I filed https://github.com/google/amber/issues/111  to sort out floating point comparisons better.

src/vulkan/buffer.cc
```diff
@@ -71,6 +71,19 @@ void Buffer::CopyToDevice(VkCommandBuffer command) {
   MemoryBarrier(command);
 }
 
+void Buffer::CopyToHost(VkCommandBuffer command) {
+  if (is_buffer_host_accessible_)
```
dneto0: Even if the buffer is accessible to the host, you have to invalidate the mapped memory to ensure the host buffer gets the latest data.

https://www.khronos.org/registry/vulkan/specs/1.1/html/vkspec.html#vkInvalidateMappedMemoryRanges 

??

src/vulkan/buffer.h
```diff
@@ -34,7 +34,8 @@ class Buffer : public Resource {
   Result CreateVkBufferView(VkFormat format);
   VkBufferView GetVkBufferView() const { return view_; }
 
-  // TODO(jaebaek): Determine copy all or partial data
+  // Only record the copy command to the command buffer and
```
dneto0: Say what the method actually does.

src/vulkan/buffer.h
```diff
@@ -45,6 +46,10 @@ class Buffer : public Resource {
     return Resource::GetHostAccessMemory();
   }
 
+  // Only record the copy command to the command buffer and
```
dneto0: Say what the method actually does.

src/vulkan/compute_pipeline.h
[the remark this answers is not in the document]
dneto0: TIL.

src/vulkan/engine_vulkan.cc
```diff
@@ -272,28 +272,41 @@ Result EngineVulkan::DoPatchParameterVertices(
...
 
   auto graphics = pipeline_->AsGraphics();
-  Result r = graphics->ProcessCommands();
-
   auto frame = graphics->GetFrame();
-  *width = frame->GetWidth();
-  *height = frame->GetHeight();
-  *stride = VkFormatToByteSize(graphics->GetColorFormat());
-  *buf = frame->GetColorBufferPtr();
+  auto format_in_bytes = VkFormatToByteSize(graphics->GetColorFormat());
```
dneto0: bytes_per_texel

src/vulkan/image.h
```diff
@@ -49,12 +46,13 @@ class Image : public Resource {
 
   // Resource
   VkDeviceMemory GetHostAccessMemory() const override {
-    if (is_image_host_accessible_)
-      return memory_;
-
     return Resource::GetHostAccessMemory();
   }
 
+  // Only record the copy command to the command buffer and
```
dneto0: Say what the method actually does.

src/vulkan/resource.h
```diff
@@ -31,10 +31,14 @@ class Resource {
     return host_accessible_memory_;
   }
 
+  virtual void CopyToHost(VkCommandBuffer command) = 0;
+
   virtual void Shutdown();
 
   void* HostAccessibleMemoryPtr() const { return memory_ptr_; }
 
+  size_t GetSize() const { return size_; }
```
dneto0: What are the units?  bytes?
What are you measuring? the whole underlying storage ?

src/verifier.cc
```diff
@@ -64,6 +202,10 @@ Result Verifier::Probe(const ProbeCommand* command,
         std::to_string(frame_width) + "," + std::to_string(frame_height) + ")");
   }
 
+  double tolerance[4] = {};
+  bool is_tolerance_percent[4] = {};
+  SetupToleranceForTexels(command, tolerance, is_tolerance_percent);
```
dj2: Do we need this? It seems like we could do something simple in the for loop below where if the tolerances size > 1 we take tolerance[i % 4] instead of building this array

src/verifier.cc
```diff
@@ -15,15 +15,153 @@
...
+  if (is_tolerance_percent)
+    tolerance *= difference;
+  return difference < tolerance;
+}
 
-bool IsFloatPixelEqualInt(float pixel, uint8_t expected) {
-  // TODO(jaebaek): Change kEpsilon to tolerance.
-  return std::fabs(pixel - static_cast<float>(expected) / 255.0f) < kEpsilon;
+template <typename T>
+Result CheckValue(const ProbeSSBOCommand* command,
```
dj2: Can these verifier changes be pulled out to a separate Cl and committed with the tests?

src/vkscript/executor_test.cc
```diff
@@ -195,10 +195,11 @@ class EngineStub : public Engine {
     return {};
   }
 
-  Result DoProcessCommands(uint32_t*,
-                           uint32_t*,
-                           uint32_t*,
-                           const void**) override {
+  Result DoProcessCommands() override { return {}; }
```
dj2: The change of DoProcessCommand into 3 methods could also possibly be split out to another CL

src/vulkan/storage_buffer_descriptor.cc
```diff
@@ -119,6 +126,12 @@ Result StorageBufferDescriptor::UpdateDescriptorSet(
       descriptor_set, VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, buffer_info);
 }
 
+void StorageBufferDescriptor::GetResourceInfo(ResourceInfo* info) {
```
dj2: Can this return a ResourceInfo instead of taking a pointer?

jaebaek: done

src/vulkan/buffer.cc
```diff
@@ -65,12 +67,27 @@ void Buffer::CopyToDevice(VkCommandBuffer command) {
...
-  region.size = GetSize();
+  region.size = GetSizeInBytes();
 
   vkCmdCopyBuffer(command, GetHostAccessibleBuffer(), buffer_, 1, &region);
   MemoryBarrier(command);
 }
 
+Result Buffer::CopyToHost(VkCommandBuffer command) {
+  if (is_buffer_host_accessible_) {
+    return FlushAndInvalidateMemoryIfNeeded();
```
dneto0: It seems weird to put both flush and invalidate in the same call.
For copying to host you want to invalidate.
For copying to device from host, you want flush

src/vulkan/buffer.h
```diff
@@ -34,7 +34,10 @@ class Buffer : public Resource {
   Result CreateVkBufferView(VkFormat format);
   VkBufferView GetVkBufferView() const { return view_; }
 
-  // TODO(jaebaek): Determine copy all or partial data
+  // If this buffer is host accessible, this method does nothing.
+  // Otherwise, this method only records the command for copying
+  // this buffer to the host accessible buffer. The actual
```
dneto0: .. "this buffer to the host accessible buffer".  ??  Isn't that the wrong way around?

dneto0: Also, compared to CopyToHost, why is this method different regarding buffers marked as COHERENT?

src/vulkan/resource.h
```diff
@@ -68,6 +70,13 @@ class Resource {
            VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT;
   }
 
+  bool CheckMemoryHostCoherent(uint32_t memory_type_index) {
```
dneto0: "Check.." should be renamed to "Is...".

The "Check" is always ambiguous: i don't know what the true value means.  With "Is" it's always clear.

src/vulkan/buffer.cc
jaebaek: Oh sorry. I was confused.

Updated.

src/vulkan/buffer.h
jaebaek: Updated.

src/vulkan/resource.h
jaebaek: done

dj2:
Can you rebase on top of ToT? The android and ndk-builds should work now.

jaebaek:
Rebased

PR #102, opened 2018-11-27
Destructor on abstract class should be virtual

dneto0:
Otherwise you get a warning from Clang on macOS. (For good reason)


dj2:
Do you know why the bots passed? Shouldn't the mac bots fail on this?

PR #104, opened 2018-11-27
Fail on unrecognized command line option

dneto0:


dneto0:
Too often I was running
     
      amber -d dawn foo.amber

And not actually running the dawn backend...

PR #110, opened 2018-11-28
Delay VkScript Shader Compilation.

dj2:
This CL changes VkScript to delay shader compilation until execution,
instead of at parse time. This allows VkScript to share the Script
object with AmberScript.


dj2:
@alan-baker this moves the shader stuff into a class where you can add your data map if needed.

PR #123, opened 2018-11-29
Vulkan: handle probe ssbo command

jaebaek:
This commit is a part of PR #101

Related to #32


jaebaek:
I just reuse #101 and drop this

PR #130, opened 2018-12-03
Add tests tolerance and comparator

### LINKED ISSUE #40: Vulkan: Implement DoTolerance command
Implement the DoTolerance command in the Vulkan engine.

jaebaek:
Fixes #40

### DIFF
--- a/src/verifier.cc
+++ b/src/verifier.cc
@@ -28,18 +28,21 @@ const double kDefaultTexelTolerance = 0.002;
 
 // It returns true if the difference is within the given error.
 // If |is_tolerance_percent| is true, the actual tolerance will be
-// relative value i.e., |tolerance| * fabs(real - expected).
+// relative value i.e., |tolerance| / 100 * fabs(expected).
 // Otherwise, this method uses the absolute value i.e., |tolerance|.
 bool IsEqualWithTolerance(const double real,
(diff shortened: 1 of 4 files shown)


src/verifier.cc
```diff
@@ -28,18 +28,21 @@ const double kDefaultTexelTolerance = 0.002;
...
                           double tolerance,
                           const bool is_tolerance_percent = true) {
   double difference = std::fabs(real - expected);
-  if (difference == 0.0)
-    return true;
-  if (is_tolerance_percent)
-    tolerance *= difference;
-  return difference < tolerance;
+  if (is_tolerance_percent) {
+    if (difference > tolerance / 100.0 * std::fabs(expected))
```
dj2: Can you add some bracketing here? Makes it easier to see it's (tolerance / 100.0) * std::fabs(expected)

src/verifier.cc
```diff
@@ -28,18 +28,21 @@ const double kDefaultTexelTolerance = 0.002;
...
+      return false;
+  } else {
+    if (difference > tolerance)
```
dj2: make it } else if (diff ...) {

src/verifier.cc

src/verifier.cc

PR #132, opened 2018-12-03
Add row_stride to key image metrics in ResourceInfo

dneto0:
This is needed for Dawn on Metal, which enforces minimum a row pitch
of 256 bytes.

### DIFF
--- a/src/verifier.cc
+++ b/src/verifier.cc
@@ -250,10 +250,10 @@ Result Verifier::Probe(const ProbeCommand* command,
   }
 
   if (count_of_invalid_pixels) {
-    const uint8_t* p =
-        ptr + row_stride * (first_invalid_j + y) + texel_stride * x;
+    const uint8_t* p = ptr + row_stride * (first_invalid_j + y) +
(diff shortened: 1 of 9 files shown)


src/verifier.cc
```diff
@@ -254,12 +261,15 @@ Result Verifier::Probe(const ProbeCommand* command,
         (command->IsRGBA() ? ", " + std::to_string(command->GetA() * 255) +
                                  "\n  Actual RGBA: "
                            : "\n  Actual RGB: ") +
-        std::to_string(static_cast<int>(p[stride * first_invalid_i])) + ", " +
-        std::to_string(static_cast<int>(p[stride * first_invalid_i + 1])) +
+        std::to_string(static_cast<int>(p[texel_stride * first_invalid_i])) +
```
dj2: If we store texel_stride * first_invalid_i somewhere and reuse it in these messages, does this format a bit nicer?

dneto0: Yep.  Looks better when we fold more right into the p base pointer.

PR #134, opened 2018-12-03
Vulkan: handle DoDrawRect

### LINKED ISSUE #39: Vulkan: Implement the DoDrawRect command
Implement the DoDrawRect command in the vulkan engine.

jaebaek:
`draw rect` command creates new vertices for a rectangle every
time. Thus, DoDrawRect() resets vertex buffer and graphics pipeline
every time and creates new one accordingly.

Fixes #39

### DIFF
--- a/src/vulkan/engine_vulkan.cc
+++ b/src/vulkan/engine_vulkan.cc
@@ -227,8 +227,61 @@ Result EngineVulkan::DoClear(const ClearCommand*) {
   return pipeline_->AsGraphics()->Clear();
 }
 
-Result EngineVulkan::DoDrawRect(const DrawRectCommand*) {
-  return Result("Vulkan::DoDrawRect Not Implemented");
+Result EngineVulkan::DoDrawRect(const DrawRectCommand* command) {
+  if (!pipeline_->IsGraphics())
+    return Result("Vulkan::DrawArrays for Non-Graphics Pipeline");
+
+  auto* graphics = pipeline_->AsGraphics();
+
+  Result r = graphics->ResetPipelineAndVertexBuffer();
+  if (!r.IsSuccess())
+    return r;
+
+  Format format;
+  format.SetFormatType(FormatType::kR32G32_SFLOAT);
+  format.AddComponent(FormatComponentType::kR, FormatMode::kSFloat, 32);
+  format.AddComponent(FormatComponentType::kG, FormatMode::kSFloat, 32);
+
+  float x = command->GetX();
+  float y = command->GetY();
+  float width = command->GetWidth();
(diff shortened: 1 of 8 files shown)


src/vulkan/engine_vulkan.cc
```diff
@@ -244,8 +244,61 @@ Result EngineVulkan::DoClear(const ClearCommand*) {
...
+  r = graphics->SetBuffer(BufferType::kVertex, 0, format, values);
+  if (!r.IsSuccess())
+    return r;
+
+  PipelineData data;
+  DrawArraysCommand draw(data);
+  draw.SetTopology(command->IsPatch() ? Topology::kPatchList
+                                      : Topology::kTriangleStrip);
+  draw.SetFirstVertexIndex(0);
+  draw.SetVertexCount(6);
```
dneto0: There are only 4 vertices.

src/vulkan/engine_vulkan.cc
```diff
@@ -244,8 +244,61 @@ Result EngineVulkan::DoClear(const ClearCommand*) {
...
+    return Result("Vulkan::DrawArrays for Non-Graphics Pipeline");
+
+  auto* graphics = pipeline_->AsGraphics();
+
+  Result r = graphics->ResetPipelineAndVertexBuffer();
+  if (!r.IsSuccess())
+    return r;
+
+  Format format;
+  format.SetFormatType(FormatType::kR32G32_SFLOAT);
```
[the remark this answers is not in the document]
dj2: Can you add a comment here which explains that?

src/vulkan/engine_vulkan.cc
```diff
@@ -244,8 +244,61 @@ Result EngineVulkan::DoClear(const ClearCommand*) {
...
+    const float frame_height = static_cast<float>(graphics->GetHeight());
+    x = ((x / frame_width) * 2.0f) - 1.0f;
+    y = ((y / frame_height) * 2.0f) - 1.0f;
+    width = ((width / frame_width) * 2.0f) - 1.0f;
+    height = ((height / frame_height) * 2.0f) - 1.0f;
+  }
+
+  std::vector<Value> values(8);
+  // Bottom left
+  values[0].SetDoubleValue(static_cast<double>(x));
```
dneto0: The values are floats, and they should automatically promote to doubles. So I think you can (and should) skip the static casts.

dneto0: @dj2 says GCC's -Wdouble-promotion will issue a warning if the static casts are removed.  Ok then.  I find that sad.
An alternate is to make the x, y, width, height variables doubles.. But that's probably overkill.

PR #137, opened 2018-12-04
Extend engine configuration options.

dj2:
Currently an InitializeWithDevice method is provided to the engines
which is populated with the device given when calling Execute. This CL
extends that mechanism to have EngineConfig structs for each engine.
Those structs are given to Execute and passed to the engine. This will
allow extending the Vulkan code to accept PhysicalDevice and available
feature/extension information along with the device.


dj2:
Does this seem like it will handle all we need? (I can't test it as I don't have vulkan or dawn on this device .... and we have nothing which exercises that interface ....)

src/engine.h
```diff
@@ -77,8 +77,15 @@ class Engine {
                             const std::vector<std::string>& extensions) = 0;
 
   // Initialize the engine with the provided device. The device is _not_ owned
-  // by the engine and should not be destroyed.
-  virtual Result InitializeWithDevice(void* default_device) = 0;
+  // by the engine and should not be destroyed. The |features| and |extensions|
+  // are passed into InitializeWithConfig for validatino purposes only. If
```
jaebaek: validatino --> validation

dj2: Good catch, done.

PR #138, opened 2018-12-04
Allow host to provide a shader map.

dj2:
This CL extends the Amber interface to allow a Shader map to be
provided. This map allows the caller to pre-compile the shaders and
provide the name to data mapping. The ShaderCompiler will then pull from
the map before attempting to do a compile.

### DIFF
--- a/include/amber/amber.h
+++ b/include/amber/amber.h
@@ -57,7 +57,8 @@ class Amber {
   /// |shader_map| to lookup shader data before attempting to compile the
   /// shader if possible.
   amber::Result ExecuteWithShaderData(const std::string& data,
-      const Options& opts, const ShaderMap& shader_data);
+                                      const Options& opts,
+                                      const ShaderMap& shader_data);
 };
 
 }  // namespace amber
--- a/include/amber/amber.h
+++ b/include/amber/amber.h
@@ -23,6 +23,8 @@
 
 namespace amber {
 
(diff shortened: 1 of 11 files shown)


include/amber/amber.h
```diff
@@ -15,12 +15,16 @@
...
 
+#include <map>
 #include <string>
+#include <vector>
 
 #include "amber/result.h"
 
 namespace amber {
 
+using ShaderMap = std::map<std::string, std::vector<uint32_t>>;
```
jaebaek: `ShaderMap` is not intuitive to me. Could you please add a comment to explain meaning of its key and value?

dj2: done

jaebaek: nit: represntation --> representation

include/amber/amber.h
```diff
@@ -48,6 +54,13 @@ class Amber {
   /// Executes the given |data| script with the provided |opts|. Returns a
   /// |Result| which indicates if the execution succeded.
   amber::Result Execute(const std::string& data, const Options& opts);
+
+  /// Executes the given |data| script with the provided |opts. Will use
```
jaebaek: nit: |opts --> |opts|

include/amber/amber.h
dj2: done

include/amber/amber.h
dj2: done

PR #141, opened 2018-12-04
Vulkan: enable required features

jaebaek:
Device class of Vulkan engine chooses a physical device that supports
features given in `[require]` section and create a logical device
that enables those features at the initialization time.

Related to #97


src/vulkan/device.cc
```diff
@@ -20,8 +20,24 @@
...
 namespace vulkan {
+namespace {
+
+bool AreAllRequiredFeaturesSupported(
+    const VkPhysicalDeviceFeatures& features,
+    const VkPhysicalDeviceFeatures& required_features) {
+  const VkBool32* supported = reinterpret_cast<const VkBool32*>(&features);
+  const VkBool32* required =
+      reinterpret_cast<const VkBool32*>(&required_features);
+  for (size_t i = 0; i < (sizeof(VkPhysicalDeviceFeatures) / sizeof(VkBool32));
```
dj2: This is scary. Is this how vulkan normally checks features? Casting the VkPhysicalDeviceFeatures to a VkBool32 seems .... wrong?

Why not just pass the std::vector<Feature> into here, then loop over that vector, have a switch(feature) and check each feature? It would be a lot clearer what's going on.

The switch below basically moves up here?

dj2: Maybe not even a switch, simple if statements might be simplest:

if (feature == kFeatureFoo && features.feature_foo == Vk_FALSE)
  return false;
if (feature == kFeatureBar && features.feature_bar == VK_FALSE)
  return false;

src/vulkan/device.h
```diff
@@ -30,7 +30,7 @@ class Device {
   explicit Device(VkDevice device);
   ~Device();
 
-  Result Initialize();
+  Result Initialize(const VkPhysicalDeviceFeatures& features);
```
dj2: are these the required_features or the available features? (and below) can you rename to make clear?

src/vulkan/engine_vulkan.cc
```diff
@@ -93,7 +274,7 @@ Result EngineVulkan::InitializeWithConfig(EngineConfig* config,
     return Result("Vulkan::InitializeWithConfig device handle is null.");
 
   device_ = MakeUnique<Device>(vk_config->device);
-  return InitDeviceAndCreateCommand();
+  return InitDeviceAndCreateCommand({});
```
dj2: We should be passing the features in here, no? That way we can verify they exist on the provided device? If you need a PhysicalDevice, you can add that to the struct in amber_vulkan.h to be provided as well.

tests/cases/ssbo_with_graphics_pipeline.amber
```diff
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+[require]
+vertexPipelineStoresAndAtomics
```
dj2: Can we move this to a different test since this one has a different purpose?

dj2: Is it possible to add a negative test for this? Do we know if there is a feature that isn't supported anywhere?

src/vulkan/device.h
jaebaek: updated

src/vulkan/engine_vulkan.cc
jaebaek: Yes, we must pass features here. I only changed this line and let me handle `EngineVulkan::InitializeWithConfig()` in other commits.

tests/cases/ssbo_with_graphics_pipeline.amber
jaebaek: This test works without this requirement but just reports an validation error when enabling the validation layer.
Later, I will add some positive/negative tests that are suitable for checking required features.

src/vulkan/engine_vulkan.cc
```diff
@@ -72,17 +73,197 @@ Result EngineVulkan::InitDeviceAndCreateCommand() {
   return {};
 }
 
-Result EngineVulkan::Initialize(const std::vector<Feature>&,
+VkPhysicalDeviceFeatures EngineVulkan::RequiredFeatures(
```
dj2: I think this should go away and we should just pass the std::vector<Features> into AreAllRequiredFeaturesAvailable. There is no reason to do this conversion.

jaebaek: Updated.

jaebaek:
PTAL

PR #143, opened 2018-12-05
 Break apart Amber::Execute method.

dj2:
This CL splits Amber::Execute into a parse and execute step. The Parse
step accepts a Recipe to fill out, that Recipe is then passed into the
Execute step. The recipe, in the future, will allow retrieving
information about the given script.

### DIFF
--- a/include/amber/shader_info.h
+++ b/include/amber/shader_info.h
@@ -12,33 +12,37 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef SRC_SHADER_COMPILER_H_
-#define SRC_SHADER_COMPILER_H_
+#ifndef AMBER_SHADER_INFO_H_
+#define AMBER_SHADER_INFO_H_
 
 #include <string>
-#include <utility>
-#include <vector>
-
-#include "amber/amber.h"
-#include "amber/result.h"
-#include "src/shader.h"
 
 namespace amber {
 
-class ShaderCompiler {
- public:
-  ShaderCompiler();
-  ~ShaderCompiler();
+enum class ShaderFormat : uint8_t {
+  kDefault = 0,
+  kText,
+  kGlsl,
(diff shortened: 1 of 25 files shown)


include/amber/shader_info.h
```diff
@@ -0,0 +1,48 @@
...
+  kVertex,
+  kTessellationControl,
+  kTessellationEvaluation,
+};
+
+struct ShaderInfo {
+  ShaderFormat format;
+  ShaderType type;
+  std::string shader_name;
+  std::string shader_data;
```
dj2: This should probably have a std::vector<std::string> optimizer_passes array as well.

include/amber/shader_info.h
```diff
@@ -0,0 +1,50 @@
...
+  kFragment,
+  kVertex,
+  kTessellationControl,
+  kTessellationEvaluation,
+};
+
+struct ShaderInfo {
+  ShaderFormat format;
+  ShaderType type;
+  std::string shader_name;
```
dneto0: Needs a comment that this is the name as declared implicitly or explicitly by the script. It could be a compound name partly, such as `<pipeline-name>.<declared-shader-name>`

The point is that this is not the entry point name provided within the module.  E.g. it's not "main".

include/amber/shader_info.h
```diff
@@ -0,0 +1,50 @@
...
+  std::string shader_data;
```
dneto0: Is the member name as clear as it could be?  Could use a comment.

A shader is the code and variables for a particular entry point in a SPIR-V module.   This field is really the source (from the vkscript or amberscript) for the SPIR-V module that contains the shader or shaders.

Suggest comment:

// Shader module contents, in the format specified by |format|

include/amber/shader_info.h
dj2: done

PR #152, opened 2018-12-07
Fix integer overflow for fence_timeout

jrprice:
The fence_timeout (in milliseconds) is stored as a uint32_t and is
later converted to nanoseconds (as required by vkWaitForFences), which
requires more than 32 bits for any timeout larger than ~4.3 seconds.

### DIFF
--- a/src/vulkan/command.cc
+++ b/src/vulkan/command.cc
@@ -107,8 +107,9 @@ Result CommandBuffer::SubmitAndReset(uint32_t timeout_ms) {
   if (vkQueueSubmit(queue_, 1, &submit_info, fence_) != VK_SUCCESS)
     return Result("Vulkan::Calling vkQueueSubmit Fail");
 
-  VkResult r = vkWaitForFences(device_, 1, &fence_, VK_TRUE,
-                               timeout_ms * 1000 * 1000 /* nanosecond */);
+  VkResult r = vkWaitForFences(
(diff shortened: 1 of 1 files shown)


src/vulkan/command.cc
```diff
@@ -108,7 +108,7 @@ Result CommandBuffer::SubmitAndReset(uint32_t timeout_ms) {
     return Result("Vulkan::Calling vkQueueSubmit Fail");
 
   VkResult r = vkWaitForFences(device_, 1, &fence_, VK_TRUE,
-                               timeout_ms * 1000 * 1000 /* nanosecond */);
+                               timeout_ms * 1000ULL * 1000ULL /* nanosecond */);
```
jaebaek: It may be fine but explicit type conversion for `timeout_ms` would be nice.
Could you please add `static_cast<uint64_t>` for `timeout_ms`?

i.e.,

```
static_cast<uint64_t>(timeout_ms) * 1000ULL * 1000ULL /* nanosecond */);
```

jaebaek:
It's weird that the presubmit test did not run .. Let me check it next Monday and merge it.

dj2:
Kokoro only runs automatically for contributors, you have to add the kokoro:run label to kick the bots off otherwise.

jaebaek:
@dj2 Thank you for letting me know it!

PR #154, opened 2018-12-10
Vulkan: enable extension support

### LINKED ISSUE #97: Add Vulkan feature and extension setup

jaebaek:
This commit checks if extensions given in `[require]` section are
supported by physical device and enables them when creating the
logical device.

Fixes #97


dj2:
Looks like device.h needs a #include <string>

PR #157, opened 2018-12-10
Remove IndicesNode and use Buffer directly

dj2:
This CL removes the IndicesNode and adds the Index buffer directly to
the Script object. The Executor is updated to pass all the set buffers
to the engine.


src/vkscript/executor.cc
```diff
@@ -90,13 +90,8 @@ Result Executor::Execute(Engine* engine,
...
 
-  // Process Indices nodes
-  for (const auto& node : script->Nodes()) {
-    if (!node->IsIndices())
-      continue;
-
-    r = engine->SetBuffer(BufferType::kIndex, 0, Format(),
-                          node->AsIndices()->Indices());
+  for (const auto& buf : script->GetBuffers()) {
+    r = engine->SetBuffer(buf->GetBufferType(), 0, Format(), buf->GetData());
```
jaebaek: I am confused. Why do we handle only the vertex data separately?
If we remove `IndicesNode`, I think we can do the same thing for `VertexDataNode`?

dj2: That's the next CL which I'll put up for review when this lands.

dj2: Vertex data has a more complicated Format than the indices DatumType. That has to be handled before vertex data can be converted.

jaebaek: ok

PR #159, opened 2018-12-10
Vulkan: fix discontinuous descriptor set error

### LINKED ISSUE #158: Error from discontinuous descriptor sets
Amber supports descriptors with consecutive descriptor sets from `[0, N]`, but it does not support discontinuous descriptor sets e.g., `1, 3, 5`.

jaebaek:
Without this commit, Amber supports descriptors with consecutive
descriptor sets from `[0, N]`, but it does not support discontinuous
descriptor sets e.g., `1, 3, 5`. To support discontinuous
descriptor sets, this commit creates empty descriptor set layouts
and puts them on pipeline layout.

Fixes #158

### DIFF
--- a/src/vulkan/pipeline.cc
+++ b/src/vulkan/pipeline.cc
@@ -366,17 +366,18 @@ Result Pipeline::SendDescriptorDataToDeviceIfNeeded() {
 }
 
 void Pipeline::BindVkDescriptorSets() {
-  for (size_t desc = 0, set = 0; desc < descriptor_set_layout_types_.size() &&
-                                 set < descriptor_sets_.size();
+  for (size_t desc = 0, non_empty_desc = 0;
+       desc < descriptor_set_layout_types_.size() &&
+       non_empty_desc < descriptor_sets_.size();
        ++desc) {
     if (descriptor_set_layout_types_[desc] ==
         DescriptorSetLayoutType::kNonEmpty) {
       vkCmdBindDescriptorSets(command_->GetCommandBuffer(),
                               IsGraphics() ? VK_PIPELINE_BIND_POINT_GRAPHICS
                                            : VK_PIPELINE_BIND_POINT_COMPUTE,
-                              pipeline_layout_, desc, 1, &descriptor_sets_[set],
-                              0, nullptr);
-      ++set;
+                              pipeline_layout_, desc, 1,
+                              &descriptor_sets_[non_empty_desc], 0, nullptr);
+      ++non_empty_desc;
     }
   }
 }
--- a/src/vulkan/pipeline.cc
+++ b/src/vulkan/pipeline.cc
@@ -65,13 +65,12 @@ void Pipeline::Shutdown() {
   if (r.IsSuccess())
     command_->SubmitAndReset(fence_timeout_ms_);
 
-  for (size_t desc = 0; desc < descriptor_sets_.size(); ++desc) {
-    vkDestroyDescriptorSetLayout(device_, descriptor_sets_[desc].layout,
-                                 nullptr);
-    if (descriptor_sets_[desc].empty)
+  for (size_t i = 0; i < descriptor_sets_.size(); ++i) {
+    vkDestroyDescriptorSetLayout(device_, descriptor_sets_[i].layout, nullptr);
+    if (descriptor_sets_[i].empty)
       continue;
 
-    vkDestroyDescriptorPool(device_, descriptor_sets_[desc].pool, nullptr);
+    vkDestroyDescriptorPool(device_, descriptor_sets_[i].pool, nullptr);
   }
 
   command_->Shutdown();
@@ -182,14 +181,14 @@ Result Pipeline::CreateDescriptorPools() {
 
     if (static_cast<uint32_t>(descriptor_sets_.size()) <= current_desc) {
       return Result(
-          "Pipeline::CreateDescriptorPools more actually used descriptor sets "
-          "than descriptor set layouts");
+          "Pipeline::CreateDescriptorPools: More descriptor sets requested "
+          "than descriptor set layouts provided");
     }
 
     if (descriptor_sets_[current_desc].empty) {
       return Result(
-          "Pipeline::CreateDescriptorPools actually used descriptor set has "
-          "empty descriptor set layout");
+          "Pipeline::CreateDescriptorPools: Used a descriptor set with an "
+          "empty layout");
     }
 
     descriptor_sets_[current_desc].pool = desc_pool;
@@ -199,22 +198,22 @@ Result Pipeline::CreateDescriptorPools() {
 }
 
 Result Pipeline::CreateDescriptorSets() {
-  for (size_t desc = 0; desc < descriptor_sets_.size(); ++desc) {
-    if (descriptor_sets_[desc].empty)
+  for (size_t i = 0; i < descriptor_sets_.size(); ++i) {
+    if (descriptor_sets_[i].empty)
       continue;
 
     VkDescriptorSetAllocateInfo desc_set_info = {};
     desc_set_info.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_ALLOCATE_INFO;
-    desc_set_info.descriptorPool = descriptor_sets_[desc].pool;
+    desc_set_info.descriptorPool = descriptor_sets_[i].pool;
     desc_set_info.descriptorSetCount = 1;
-    desc_set_info.pSetLayouts = &descriptor_sets_[desc].layout;
+    desc_set_info.pSetLayouts = &descriptor_sets_[i].layout;
 
     VkDescriptorSet desc_set = VK_NULL_HANDLE;
     if (vkAllocateDescriptorSets(device_, &desc_set_info, &desc_set) !=
         VK_SUCCESS) {
       return Result("Vulkan::Calling vkAllocateDescriptorSets Fail");
     }
-    descriptor_sets_[desc].vk_desc_set = desc_set;
+    descriptor_sets_[i].vk_desc_set = desc_set;
   }
 
   return {};
@@ -270,14 +269,14 @@ Result Pipeline::UpdateDescriptorSetsIfNeeded() {
          ++i) {
       if (static_cast<uint32_t>(descriptor_sets_.size()) <= current_desc) {
         return Result(
-            "Pipeline::UpdateDescriptorSetsIfNeeded more actually used "
-            "descriptor sets than descriptor set layouts");
+            "Pipeline::UpdateDescriptorSetsIfNeeded: More descriptor sets "
+            "requested than descriptor set layouts provided");
       }
 
       if (descriptor_sets_[current_desc].empty) {
         return Result(
-            "Pipeline::UpdateDescriptorSetsIfNeeded actually used descriptor "
-            "set has empty descriptor set layout");
+            "Pipeline::UpdateDescriptorSetsIfNeeded: Used a descriptor set "
+            "with an empty layout");
       }
 
       Result r = descriptors_[i]->UpdateDescriptorSetIfNeeded(
@@ -387,15 +386,15 @@ Result Pipeline::SendDescriptorDataToDeviceIfNeeded() {
 }
 
 void Pipeline::BindVkDescriptorSets() {
-  for (size_t desc = 0; desc < descriptor_sets_.size(); ++desc) {
-    if (descriptor_sets_[desc].empty)
+  for (size_t i = 0; i < descriptor_sets_.size(); ++i) {
+    if (descriptor_sets_[i].empty)
       continue;
 
     vkCmdBindDescriptorSets(command_->GetCommandBuffer(),
                             IsGraphics() ? VK_PIPELINE_BIND_POINT_GRAPHICS
                                          : VK_PIPELINE_BIND_POINT_COMPUTE,
-                            pipeline_layout_, desc, 1,
-                            &descriptor_sets_[desc].vk_desc_set, 0, nullptr);
+                            pipeline_layout_, i, 1,
+                            &descriptor_sets_[i].vk_desc_set, 0, nullptr);
   }
 }
 
--- a/src/vulkan/pipeline.cc
+++ b/src/vulkan/pipeline.cc
@@ -86,21 +86,18 @@ Result Pipeline::CreateDescriptorSetLayouts() {
     VkDescriptorSetLayoutCreateInfo desc_info = {};
     desc_info.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_LAYOUT_CREATE_INFO;
 
-    // If two descriptor sets are discontinuous, we must put empty
-    // descriptor set layouts between them.
+    // If there are no descriptors for this descriptor set we only
+    // need to create its layout and there will be no bindings.
     std::vector<VkDescriptorSetLayoutBinding> bindings;
-    if (!info.empty) {
-      for (auto& desc : info.descriptors_) {
-        bindings.emplace_back();
-        bindings.back().binding = desc->GetBinding();
-        bindings.back().descriptorType = ToVkDescriptorType(desc->GetType());
-        bindings.back().descriptorCount = 1;
-        bindings.back().stageFlags = VK_SHADER_STAGE_ALL;
-      }
-
-      desc_info.bindingCount = bindings.size();
-      desc_info.pBindings = bindings.data();
+    for (auto& desc : info.descriptors_) {
+      bindings.emplace_back();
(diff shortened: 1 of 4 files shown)


src/vulkan/pipeline.cc
```diff
@@ -185,10 +221,35 @@ Result Pipeline::CreateDescriptorSets() {
 }
 
 Result Pipeline::CreatePipelineLayout() {
+  // Put all elements of |descriptor_set_layouts_| and
+  // |empty_descriptor_set_layouts_| together. Note that their
+  // order and the order of actually used descriptor sets must
+  // exactly match.
```
dj2: I don't understand what this means "their order and the order of actually used descriptor sets must exactly match" ?

src/vulkan/pipeline.cc
```diff
@@ -185,10 +221,35 @@ Result Pipeline::CreateDescriptorSets() {
...
+  std::vector<VkDescriptorSetLayout> layouts;
+  uint32_t desc = 0;
+  size_t empty_descriptor_set_layouts_index = 0;
+  for (size_t i = 0; i < descriptor_set_numbers_.size(); ++i) {
```
dj2: Instead of storing to lists, why not just store a vector that you allocate to the max descriptor set value, then fill in the missing ones with empty descriptor sets?

src/vulkan/pipeline.cc
```diff
@@ -113,12 +118,43 @@ Result Pipeline::CreateDescriptorSetLayouts() {
...
-void Pipeline::DestoryDescriptorPools() {
+void Pipeline::DestroyEmptyDescriptorSetLayouts() {
+  for (size_t i = 0; i < empty_descriptor_set_layouts_.size(); ++i) {
+    vkDestroyDescriptorSetLayout(device_, empty_descriptor_set_layouts_[i],
+                                 nullptr);
+  }
+  empty_descriptor_set_layouts_.clear();
+}
+
+Result Pipeline::CreateEmptyDescriptorSetLayouts() {
```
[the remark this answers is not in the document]
jaebaek: I removed this method and put both empty and non-empty descriptor set layouts onto `descriptor_set_layouts_`.

src/vulkan/pipeline.cc
```diff
@@ -113,12 +118,43 @@ Result Pipeline::CreateDescriptorSetLayouts() {
...
+  uint32_t desc = 0;
+  for (size_t i = 0; i < descriptor_set_numbers_.size(); ++i) {
```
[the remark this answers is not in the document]
jaebaek: I removed `descritpor_set_numbers_`.
I use `std::vector<DescriptorSetLayoutType> descriptor_set_layout_types_` which determines whether each element of `descriptor_set_layouts_` is empty or not.

src/vulkan/pipeline.cc
jaebaek: I removed this code to put both empty and non-empty ones together.

src/vulkan/pipeline.cc
jaebaek: Updated. I put all together into `descriptor_set_layouts_`.

src/vulkan/pipeline.cc
```diff
@@ -312,11 +346,20 @@ Result Pipeline::SendDescriptorDataToDeviceIfNeeded() {
...
-  vkCmdBindDescriptorSets(command_->GetCommandBuffer(),
-                          IsGraphics() ? VK_PIPELINE_BIND_POINT_GRAPHICS
-                                       : VK_PIPELINE_BIND_POINT_COMPUTE,
-                          pipeline_layout_, 0, descriptor_sets_.size(),
-                          descriptor_sets_.data(), 0, nullptr);
+  for (size_t desc = 0, non_empty_desc = 0;
+       desc < descriptor_set_layout_types_.size() &&
+       non_empty_desc < descriptor_sets_.size();
+       ++desc) {
+    if (descriptor_set_layout_types_[desc] ==
```
dj2: if (descriptor_set_layout_types_[desc] != DescriptorSetLayotuType::kNonEmpty)
  continue;

src/vulkan/pipeline.cc
```diff
@@ -166,19 +192,25 @@ Result Pipeline::CreateDescriptorPools() {
...
-    desc_set_info.pSetLayouts = &descriptor_set_layouts_[i];
-
-    VkDescriptorSet desc_set = VK_NULL_HANDLE;
-    if (vkAllocateDescriptorSets(device_, &desc_set_info, &desc_set) !=
-        VK_SUCCESS) {
-      return Result("Vulkan::Calling vkAllocateDescriptorSets Fail");
+  for (size_t desc = 0, pool = 0; desc < descriptor_set_layout_types_.size() &&
+                                  pool < descriptor_pools_.size();
+       ++desc) {
+    if (descriptor_set_layout_types_[desc] ==
```
dj2: if (descriptor_set_layotu_types_[desc] != DescriptorSetLayoutType::kNonEmpty)
  continue;

src/vulkan/pipeline.cc
jaebaek: done

src/vulkan/pipeline.cc
jaebaek: done

src/vulkan/pipeline.cc
```diff
@@ -65,87 +65,107 @@ void Pipeline::Shutdown() {
   if (r.IsSuccess())
     command_->SubmitAndReset(fence_timeout_ms_);
 
-  DestoryDescriptorPools();
-  DestoryDescriptorSetLayouts();
+  for (size_t desc = 0; desc < descriptor_sets_.size(); ++desc) {
```
dj2: Let's change desc to i. Naming it desc makes me think it's the descriptor, not the index.

src/vulkan/pipeline.cc
```diff
@@ -65,87 +65,107 @@ void Pipeline::Shutdown() {
...
   size_t i = 0;
+  uint32_t next_desc = 0;
   while (i < descriptors_.size()) {
     std::vector<VkDescriptorSetLayoutBinding> bindings;
-    for (const uint32_t current_desc = descriptors_[i]->GetDescriptorSet();
-         i < descriptors_.size() &&
-         current_desc == descriptors_[i]->GetDescriptorSet();
-         ++i) {
+    const uint32_t current_desc = descriptors_[i]->GetDescriptorSet();
+    while (i < descriptors_.size() &&
```
dj2: What is the relation between descriptors_ and descriptor_sets_? Is it 1:1?

src/vulkan/pipeline.cc
```diff
@@ -159,36 +179,57 @@ Result Pipeline::CreateDescriptorPools() {
         VK_SUCCESS) {
       return Result("Vulkan::Calling vkCreateDescriptorPool Fail");
     }
-    descriptor_pools_.push_back(desc_pool);
+
+    if (static_cast<uint32_t>(descriptor_sets_.size()) <= current_desc) {
+      return Result(
+          "Pipeline::CreateDescriptorPools more actually used descriptor sets "
+          "than descriptor set layouts");
```
dj2: "Pipeline::CreateDescriptorPools: More descriptor sets requested then descriptor set layouts provided"

src/vulkan/pipeline.cc
```diff
@@ -159,36 +179,57 @@ Result Pipeline::CreateDescriptorPools() {
...
+    }
+
+    if (descriptor_sets_[current_desc].empty) {
+      return Result(
+          "Pipeline::CreateDescriptorPools actually used descriptor set has "
+          "empty descriptor set layout");
```
dj2: "Pipeline::CreateDescriptorPools: Used a descriptor set with an empty layout"

src/vulkan/pipeline.cc
```diff
@@ -159,36 +179,57 @@ Result Pipeline::CreateDescriptorPools() {
...
+
+    descriptor_sets_[current_desc].pool = desc_pool;
   }
 
   return {};
 }
 
 Result Pipeline::CreateDescriptorSets() {
-  for (size_t i = 0; i < descriptor_set_layouts_.size(); ++i) {
+  for (size_t desc = 0; desc < descriptor_sets_.size(); ++desc) {
```
dj2: Let's use i instead of desc.

src/vulkan/pipeline.cc
```diff
@@ -220,14 +260,28 @@ Result Pipeline::CreateVkDescriptorRelatedObjectsIfNeeded() {
...
     for (const uint32_t current_desc = descriptors_[i]->GetDescriptorSet();
          i < descriptors_.size() &&
          current_desc == descriptors_[i]->GetDescriptorSet();
          ++i) {
-      Result r =
-          descriptors_[i]->UpdateDescriptorSetIfNeeded(descriptor_sets_[j]);
+      if (static_cast<uint32_t>(descriptor_sets_.size()) <= current_desc) {
+        return Result(
+            "Pipeline::UpdateDescriptorSetsIfNeeded more actually used "
+            "descriptor sets than descriptor set layouts");
```
dj2: same change as above

src/vulkan/pipeline.cc
```diff
@@ -220,14 +260,28 @@ Result Pipeline::CreateVkDescriptorRelatedObjectsIfNeeded() {
...
+      }
+
+      if (descriptor_sets_[current_desc].empty) {
+        return Result(
+            "Pipeline::UpdateDescriptorSetsIfNeeded actually used descriptor "
+            "set has empty descriptor set layout");
```
dj2: same change as above.

src/vulkan/pipeline.h
```diff
@@ -83,29 +83,34 @@ class Pipeline {
...
-  std::vector<VkDescriptorSetLayout> descriptor_set_layouts_;
-  std::vector<VkDescriptorPool> descriptor_pools_;
-  std::vector<VkDescriptorSet> descriptor_sets_;
-
   VkDevice device_ = VK_NULL_HANDLE;
   VkPhysicalDeviceMemoryProperties memory_properties_;
   std::unique_ptr<CommandBuffer> command_;
 
  private:
+  struct DescriptorSet {
```
dj2: Should this just be part of the Descriptor object? It seems like they're always used together?

src/vulkan/pipeline.cc
jaebaek: done

src/vulkan/pipeline.cc
jaebaek: N:1
Each element of `descriptor_sets_` can be matched to one or more elements of `descriptors_`

src/vulkan/pipeline.cc
jaebaek: done
I used "than" not "then"

src/vulkan/pipeline.cc
jaebaek: done

src/vulkan/pipeline.cc
jaebaek: done

src/vulkan/pipeline.cc
jaebaek: done

src/vulkan/pipeline.cc
jaebaek: done

src/vulkan/pipeline.h
jaebaek: Hmm .. I don't think so.

- `DescriptorSet` is not 1:1 matched with `Descriptor`. `DescriptorSet` has one or more `Descriptor`.
- `Descriptor` does not use `DescriptorSet` inside. It is only used by `Pipeline`.

dj2: So, how are they related? Is a DescriptorSet required to make a Descriptor? One DescriptorSet can make multiple Descriptors but the DescriptorSets have to be contiguous? Can we change the DescriptorSets struct to have a vector<Descriptor> in it which is all the descriptors for a given DescriptorSet?

src/vulkan/pipeline.cc
```diff
@@ -65,75 +65,60 @@ void Pipeline::Shutdown() {
...
-
 Result Pipeline::CreateDescriptorSetLayouts() {
-  std::sort(descriptors_.begin(), descriptors_.end());
+  for (auto& info : descriptor_set_info_) {
+    VkDescriptorSetLayoutCreateInfo desc_info = {};
+    desc_info.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_LAYOUT_CREATE_INFO;
 
-  size_t i = 0;
-  while (i < descriptors_.size()) {
+    // If two descriptor sets are discontinuous, we must put empty
```
dj2: I think this comment is out of date. Something like:

If there are no descriptors for this descriptor_set we only need to create the layout and
there will be no bindings.

src/vulkan/pipeline.cc
```diff
@@ -299,11 +307,16 @@ Result Pipeline::SendDescriptorDataToDeviceIfNeeded() {
   if (!r.IsSuccess())
     return r;
 
-  for (const auto& desc : descriptors_) {
-    r = desc->CreateOrResizeIfNeeded(command_->GetCommandBuffer(),
-                                     memory_properties_);
-    if (!r.IsSuccess())
-      return r;
+  for (auto& info : descriptor_set_info_) {
+    if (info.empty)
```
dj2: This check isn't needed because the for() below won't do anything if it's empty

src/vulkan/pipeline.cc
```diff
@@ -281,15 +284,20 @@ Result Pipeline::AddDescriptor(const BufferCommand* buffer_command) {
...
-  if (descriptors_.size() == 0)
-    return {};
-
   bool data_send_needed = false;
-  for (const auto& desc : descriptors_) {
-    if (desc->HasDataNotSent()) {
-      data_send_needed = true;
-      break;
+  for (auto& info : descriptor_set_info_) {
+    if (info.empty)
```
dj2: Check not needed.

src/vulkan/pipeline.cc
```diff
@@ -216,49 +206,62 @@ Result Pipeline::CreateVkDescriptorRelatedObjectsIfNeeded() {
...
-  for (size_t i = 0, j = 0;
-       i < descriptors_.size() && j < descriptor_sets_.size(); ++j) {
-    for (const uint32_t current_desc = descriptors_[i]->GetDescriptorSet();
-         i < descriptors_.size() &&
-         current_desc == descriptors_[i]->GetDescriptorSet();
-         ++i) {
-      Result r =
-          descriptors_[i]->UpdateDescriptorSetIfNeeded(descriptor_sets_[j]);
+  for (auto& info : descriptor_set_info_) {
+    if (info.empty)
```
dj2: Check not needed

src/vulkan/pipeline.cc
```diff
@@ -324,19 +337,28 @@ Result Pipeline::SendDescriptorDataToDeviceIfNeeded() {
   if (!r.IsSuccess())
     return r;
 
-  for (const auto& desc : descriptors_) {
-    desc->UpdateResourceIfNeeded(command_->GetCommandBuffer());
+  for (auto& info : descriptor_set_info_) {
+    if (info.empty)
```
dj2: Check not needed

src/vulkan/pipeline.cc
jaebaek: done

src/vulkan/pipeline.cc
jaebaek: removed

src/vulkan/pipeline.cc
jaebaek: removed

src/vulkan/pipeline.cc
jaebaek: removed

src/vulkan/pipeline.cc
jaebaek: removed

src/vulkan/pipeline.cc
```diff
@@ -65,75 +65,60 @@ void Pipeline::Shutdown() {
...
-      bindings.back().binding = descriptors_[i]->GetBinding();
-      bindings.back().descriptorType =
-          ToVkDescriptorType(descriptors_[i]->GetType());
-      bindings.back().descriptorCount = 1;
-      // TODO(jaebaek): Amber script must contain which stages use each
-      //                descriptor. Set .stageFlags attribute based on it.
-      bindings.back().stageFlags = VK_SHADER_STAGE_ALL;
-      // TODO(jaebaek): Properly set .pImmutableSamplers attribute.
-    }
+    if (!info.empty) {
```
jaebaek: I also removed check here.

PR #162, opened 2018-12-10
Vulkan: separate descriptor set update from use

### LINKED ISSUE #161: Vulkan backend submits commands for updating descriptor sets and using them at the same time
If it does, it is not guaranteed that descriptors have correct values when they are used.

jaebaek:
If commands for updating a descriptor set and using the descriptor
set are submitted at the same time, it is not guaranteed that the
descriptor has correct values when using it. This CL separates a
command submission for updating a descriptor set from the one for
using it.

Fixes #161

### DIFF
--- a/src/vulkan/compute_pipeline.cc
+++ b/src/vulkan/compute_pipeline.cc
@@ -68,11 +68,30 @@ Result ComputePipeline::Compute(uint32_t x, uint32_t y, uint32_t z) {
   if (!r.IsSuccess())
     return r;
 
+  r = command_->End();
+  if (!r.IsSuccess())
+    return r;
+
+  r = command_->SubmitAndReset(GetFenceTimeout());
+  if (!r.IsSuccess())
+    return r;
+
   if (pipeline_ == VK_NULL_HANDLE) {
     Result r = CreateVkComputePipeline();
     if (!r.IsSuccess())
       return r;
   }
+
+  // Note that a command updating a descriptor set and a command using
+  // it must be submitted separately, because using a descriptor set
+  // while updating it is not safe.
+  r = UpdateDescriptorSetsIfNeeded();
+  if (!r.IsSuccess())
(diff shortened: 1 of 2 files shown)


src/vulkan/compute_pipeline.cc
```diff
@@ -68,12 +68,31 @@ Result ComputePipeline::Compute(uint32_t x, uint32_t y, uint32_t z) {
...
+  // while updating it is not safe.
+  r = UpdateDescriptorSetsIfNeeded();
+  if (!r.IsSuccess())
+    return r;
+
+  r = command_->BeginIfNotInRecording();
+  if (!r.IsSuccess())
+    return r;
+
   r = UpdateDescriptorSetsIfNeeded();
```
dj2: Is this still needed? It's weird we call UpdateDescriptorSetsIfNeeded before and after the BeginfNotInRecording.

src/vulkan/graphics_pipeline.cc
```diff
@@ -516,12 +516,31 @@ Result GraphicsPipeline::Draw(const DrawArraysCommand* command) {
...
+  // while updating it is not safe.
+  r = UpdateDescriptorSetsIfNeeded();
+  if (!r.IsSuccess())
+    return r;
+
+  r = command_->BeginIfNotInRecording();
+  if (!r.IsSuccess())
+    return r;
+
   r = UpdateDescriptorSetsIfNeeded();
```
dj2: ditto

src/vulkan/compute_pipeline.cc
jaebaek: Nice catch. This is my mistake. Removed

src/vulkan/graphics_pipeline.cc
jaebaek: Removed.

PR #166, opened 2018-12-11
Fix run-lint on mac

dj2:
This cl changes the find command to be portable to OSX.


dj2:
Check format passed on all bots, so this is safe to merge before others finish.

PR #167, opened 2018-12-11
Output any errors produced during test runs.

### LINKED ISSUE #151: Validation layer error is not shown/checked when running tests/run_tests.py
We must check error messages from Vulkan validation layer for tests and report them to users.

dj2:
This CL captures the stderr of the amber process and fails the tests if
there were error messages reported. This should allow catching vulkan
validation errors.

Fixes #151

### DIFF
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -53,7 +53,11 @@ class TestRunner:
     cmd += [tc.GetInputPath()]
 
     try:
-      subprocess.check_output(cmd)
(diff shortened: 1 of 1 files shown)


dj2:
@jaebaek can you try this on your machine and make sure that a validation failure will now fail the test properly? (Is it possible to add a test which fails validation but would pass otherwise?)

jaebaek:
It works but it does not print the error message. Could you please check it?

dj2:
Ok, try now.

PR #172, opened 2018-12-11
[ndk-build] Only build a single platform.

dj2:
This CL changes the NDK build to only build 'armeabi-v7a' instead of
building all ABIs. We use this as a check that the Android.mk files are
correct, which we can do with a single build, and it should be
substantially faster.


dj2:
This is my attempt to speed up the ndk-build bot (which takes ~45 minutes at the moment.) Good idea? Bad idea?

dj2:
(This took NDK to about 7 minutes, more inline with all the other bots)

jaebaek:
It looks good to me. Let's wait for reviews from @dneto0

PR #177, opened 2018-12-12
Vulkan: fix draw array after draw rect bug

### LINKED ISSUE #169: Vulkan engine: draw array after draw rect doesn't work
`draw rect` command generates a new vertex buffer for the rectangle to given it.
When we implemented this, `GraphicsPipeline` just destroys the existing vertex buffer and creates new one.
As a result, `draw array` after `draw rect` looses its vertex buffer and has errors.

jaebaek:
`draw rect` command generates a new vertex buffer for its target
rectangle. Without this commit, GraphicsPipeline just destroys the
existing vertex buffer and creates new one. As a result, draw array
after draw rect looses its vertex buffer and has errors.

Fixes #169

### DIFF
--- a/src/vulkan/engine_vulkan.cc
+++ b/src/vulkan/engine_vulkan.cc
@@ -103,6 +103,10 @@ Result EngineVulkan::Shutdown() {
     vkDestroyShaderModule(device_->GetDevice(), it->second, nullptr);
 
   pipeline_->Shutdown();
+
+  if (vertex_buffer_)
+    vertex_buffer_->Shutdown();
+
   pool_->Shutdown();
   device_->Shutdown();
   return {};
@@ -185,12 +189,14 @@ Result EngineVulkan::SetBuffer(BufferType type,
   if (!pipeline_)
     return Result("Vulkan::SetBuffer no Pipeline exists");
 
-  // TODO(jaebaek): Doublecheck those buffers are only for the graphics
-  //                pipeline.
   if (!pipeline_->IsGraphics())
     return Result("Vulkan::SetBuffer for Non-Graphics Pipeline");
 
-  pipeline_->AsGraphics()->SetBuffer(type, location, format, values);
+  if (!vertex_buffer_)
+    vertex_buffer_ = MakeUnique<VertexBuffer>(device_->GetDevice());
+
+  pipeline_->AsGraphics()->SetVertexBuffer(type, location, format, values,
+                                           vertex_buffer_.get());
   return {};
 }
 
@@ -229,7 +235,7 @@ Result EngineVulkan::DoDrawRect(const DrawRectCommand* command) {
 
   auto* graphics = pipeline_->AsGraphics();
 
-  Result r = graphics->ResetPipelineAndVertexBuffer();
+  Result r = graphics->ResetPipeline();
   if (!r.IsSuccess())
     return r;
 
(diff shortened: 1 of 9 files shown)


src/vulkan/engine_vulkan.cc
```diff
@@ -185,12 +189,14 @@ Result EngineVulkan::SetBuffer(BufferType type,
   if (!pipeline_)
     return Result("Vulkan::SetBuffer no Pipeline exists");
 
-  // TODO(jaebaek): Doublecheck those buffers are only for the graphics
-  //                pipeline.
   if (!pipeline_->IsGraphics())
     return Result("Vulkan::SetBuffer for Non-Graphics Pipeline");
 
-  pipeline_->AsGraphics()->SetBuffer(type, location, format, values);
+  if (!vertex_buffer_)
```
dj2: What happens if it's a kIndex buffer? Why do we create and set a vertex buffer in that case?

src/vulkan/graphics_pipeline.cc
```diff
@@ -327,32 +328,38 @@ Result GraphicsPipeline::Initialize(uint32_t width,
...
-    return {};
+  if (type != BufferType::kVertex) {
+    return Result(
+        "GraphicsPipeline::SetVertexBuffer: non-vertex buffer type not "
+        "implemented");
+  }
 
-  if (!vertex_buffer_)
-    vertex_buffer_ = MakeUnique<VertexBuffer>(device_);
+  if (vertex_buffer == nullptr) {
```
dj2: if (!vertex_buffer)

src/vulkan/graphics_pipeline.cc
```diff
@@ -327,32 +328,38 @@ Result GraphicsPipeline::Initialize(uint32_t width,
...
-Result GraphicsPipeline::SetBuffer(BufferType type,
-                                   uint8_t location,
-                                   const Format& format,
-                                   const std::vector<Value>& values) {
+Result GraphicsPipeline::SetVertexBuffer(BufferType type,
+                                         uint8_t location,
+                                         const Format& format,
+                                         const std::vector<Value>& values,
+                                         VertexBuffer* vertex_buffer) {
   // TODO(jaebaek): Handle indices data.
```
dj2: Why does SetVertexBuffer handle an Index buffer?

src/vulkan/graphics_pipeline.cc
```diff
@@ -327,32 +328,38 @@ Result GraphicsPipeline::Initialize(uint32_t width,
...
     return r;
 
-  vertex_buffer_->SetData(location, format, values);
+  vertex_buffer->SetData(location, format, values);
   return {};
 }
 
-Result GraphicsPipeline::SendBufferDataIfNeeded() {
-  if (!vertex_buffer_)
+Result GraphicsPipeline::SendBufferDataIfNeeded(VertexBuffer* vertex_buffer) {
```
dj2: Why does this take a VertexBuffer? Can't we send other types of buffer data as well?

src/vulkan/engine_vulkan.cc
jaebaek: This is a mistake. Updated.

src/vulkan/graphics_pipeline.cc
jaebaek: removed

src/vulkan/graphics_pipeline.cc
jaebaek: done

src/vulkan/graphics_pipeline.cc
jaebaek: Changed it as `SendVertexBufferDataIfNeeded`

PR #184, opened 2018-12-12
Vulkan: enable validation layer

### LINKED ISSUE #153: Instance layer and extension
#97 is related to features and extensions of VkDevice which is from VkRunner's design.
VkRunner does not support instance layer and extension.
"features and extensions" in VkRunner mean "features and extensions of VkDevice".

We must support instance layer and extension e.g., enable validation layer on Android.
@dj2 do you think Amber script must provide commands to set up instance layer and extension?

jaebaek:
Fixes #153


src/vulkan/device.cc
```diff
@@ -24,6 +25,23 @@ namespace amber {
...
+    // Note that the order of enabled layers is important. It is
+    // based on Android NDK Vulkan document. Don't change it.
+    "VK_LAYER_GOOGLE_threading",      "VK_LAYER_LUNARG_parameter_validation",
+    "VK_LAYER_LUNARG_object_tracker", "VK_LAYER_LUNARG_core_validation",
+    "VK_LAYER_LUNARG_device_limits",  "VK_LAYER_LUNARG_image",
+    "VK_LAYER_LUNARG_swapchain",      "VK_LAYER_GOOGLE_unique_objects",
+#else   // __ANDROID__
+    // We assume that it runs with LunarG Vulkan SDK if it is not
+    // Android.
+    "VK_LAYER_LUNARG_standard_validation",
```
dj2: I don't think this is true if you run with MoltenVK.

src/vulkan/device.cc
```diff
@@ -24,6 +25,23 @@ namespace amber {
 namespace vulkan {
 namespace {
 
+const std::vector<const char*> kRequiredValidationLayers = {
+#ifdef __ANDROID__
+    // Note that the order of enabled layers is important. It is
+    // based on Android NDK Vulkan document. Don't change it.
```
dj2: I'd remove the 'Don't change it.'

src/vulkan/device.cc
```diff
@@ -24,6 +25,23 @@ namespace amber {
 namespace vulkan {
 namespace {
 
+const std::vector<const char*> kRequiredValidationLayers = {
```
dj2: This will case issues with exit destructors. Change it to a

const char* kRequiredValidationLayers[] = { ... }

src/vulkan/device.cc
```diff
@@ -14,6 +14,7 @@
 
 #include "src/vulkan/device.h"
 
+#include <cstring>
```
dj2: What requires this?

src/vulkan/device.cc
jaebaek: done

src/vulkan/device.cc
jaebaek: done

src/vulkan/device.cc
jaebaek: Added a TODO

jaebaek:
@s-perron Could you please review this PR when you have some spare time?

src/vulkan/device.cc
dj2: I don't think you want this as a TODO as you're going to break MoltenVK by adding this. You at least should make the else conditional.

src/vulkan/device.cc
```diff
@@ -14,16 +14,62 @@
...
+      flag_message = "Error";
+      break;
+    case VK_DEBUG_REPORT_WARNING_BIT_EXT:
+      flag_message = "Warning";
+      break;
+    default:
+      flag_message = "No flag";
+      break;
+  }
+  LOGE("[%s] validation layer (%s): %s", flag_message, layerPrefix, msg);
```
dj2: Let's not use a macro here. Make a method, pass in the string error message and call __android_log_write instead.

src/vulkan/device.cc
```diff
@@ -14,16 +14,62 @@
...
+  const char* flag_message = nullptr;
+  switch (flag) {
+    case VK_DEBUG_REPORT_ERROR_BIT_EXT:
+      flag_message = "Error";
+      break;
+    case VK_DEBUG_REPORT_WARNING_BIT_EXT:
+      flag_message = "Warning";
+      break;
+    default:
+      flag_message = "No flag";
```
dj2: What does no flag mean?

src/vulkan/device.cc
```diff
@@ -551,21 +664,50 @@ bool Device::ChooseQueueFamilyIndex(const VkPhysicalDevice& physical_device) {
...
 }
 
+Result Device::CreateDebugReportCallback() {
+  VkDebugReportCallbackCreateInfoEXT info = {};
+  info.sType = VK_STRUCTURE_TYPE_DEBUG_REPORT_CALLBACK_CREATE_INFO_EXT;
+  info.flags = VK_DEBUG_REPORT_ERROR_BIT_EXT | VK_DEBUG_REPORT_WARNING_BIT_EXT;
+  info.pfnCallback = debugCallback;
+
+  auto vkCreateDebugReportCallbackEXT =
+      reinterpret_cast<PFN_vkCreateDebugReportCallbackEXT>(
```
dj2: This should probably be generalized somewhere as it's done a few times.

src/vulkan/device.cc
jaebaek: Based on [doc of MoltenVk](https://github.com/KhronosGroup/MoltenVK/blob/master/Docs/MoltenVK_Runtime_UserGuide.md#known-moltenvk-limitations), we must use validation layers of LunarG.

Let me just update the comment.

dj2: Have you tested it on Molten? Last time I tried Molten the driver did not report any layers as being available which means we'll always fail.

PR #187, opened 2018-12-14
Vulkan: fix validation errors

### LINKED ISSUE #185: Vulkan: Validation errors after enabling LunarG and Android validation layers

jaebaek:
Fixes #185

### DIFF
--- a/src/vulkan/buffer_descriptor.cc
+++ b/src/vulkan/buffer_descriptor.cc
@@ -122,6 +122,8 @@ Result BufferDescriptor::CreateOrResizeIfNeeded(
 
     new_buffer->CopyFromBuffer(command, *buffer_);
 
+    not_destroyed_buffers_.push_back(std::move(buffer_));
+
     buffer_ = std::move(new_buffer);
 
     SetUpdateDescriptorSetNeeded();
@@ -172,6 +174,8 @@ ResourceInfo BufferDescriptor::GetResourceInfo() {
 
 void BufferDescriptor::Shutdown() {
   buffer_->Shutdown();
+  for (auto& buffer : not_destroyed_buffers_)
+    buffer->Shutdown();
 }
 
 }  // namespace vulkan
--- a/src/vulkan/buffer_descriptor.h
+++ b/src/vulkan/buffer_descriptor.h
@@ -67,6 +67,7 @@ class BufferDescriptor : public Descriptor {
   }
 
   std::unique_ptr<Buffer> buffer_;
+  std::vector<std::unique_ptr<Buffer>> not_destroyed_buffers_;
 };
 
 }  // namespace vulkan
--- a/src/vulkan/buffer.cc
+++ b/src/vulkan/buffer.cc
@@ -106,15 +106,13 @@ void Buffer::Shutdown() {
     view_ = VK_NULL_HANDLE;
   }
 
-  if (buffer_ != VK_NULL_HANDLE) {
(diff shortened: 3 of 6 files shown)


src/vulkan/buffer_descriptor.h
```diff
@@ -67,6 +67,7 @@ class BufferDescriptor : public Descriptor {
   }
 
   std::unique_ptr<Buffer> buffer_;
+  std::vector<std::unique_ptr<Buffer>> not_destroyed_buffers_;
```
dneto0: This seems excessive.
Isn't it better to destroy the old buffer when you are resizing it?  That way you don't accumulate more GPU resources over time.

src/vulkan/buffer_descriptor.cc
```diff
@@ -121,6 +121,8 @@ Result BufferDescriptor::CreateOrResizeIfNeeded(
 
     new_buffer->CopyFromBuffer(command, *buffer_);
 
+    not_destroyed_buffers_.push_back(std::move(buffer_));
```
dneto0: Isn't it better to do buffer_->Shutdown() here instead?

jaebaek: The problem is that we must copy the content of `buffer_` to `new_buffer`.
The above line `new_buffer->CopyFromBuffer(command, *buffer_);` record a command for the copy but it did not submit the command yet.
If we destroy `buffer_` here, we lost the information when it actually submit the copy command.

An alternative is plumbing vulkan command buffer here and call submission and reset.

Do u think the alternative is better than keeping those buffers in `not_destroyed_buffers_`?

src/vulkan/buffer_descriptor.h
jaebaek: See the comment below

src/vulkan/buffer_descriptor.cc
dneto0: Ah, right you have to keep the data.  :-)

I still think it's better to destroy the buffer as soon as you can. So I recommend:

   resize is {
     create a new buffer with the bigger size
     copy the data from the old buffer to the new buffer
     destroy the old buffer
     set buffer_ to the newly allocated buffer.
  }

jaebaek:
Blocked on the [discussion](https://github.com/Igalia/vkrunner/issues/46) of the expected behavior when there are multiple updates for a SSBO in VkRunner spec.

src/vulkan/buffer_descriptor.cc
jaebaek: As open a bug #200, I will create a PR to replace it with allocation of maximum VkBuffer before starting the execution instead of resizing it.

jaebaek:
I opened two issues related to the discussion in PR #197 #200 
Let me create new PRs for those issues and now this PR is fine?

dj2:
Is this PR still valid or can it be closed now?

src/vulkan/buffer_descriptor.cc
jaebaek: We changed the design for this issue. We dropped #200 and will go for #207.

jaebaek:
This PR is still valid except the resizing issue. Why don't we just accept this PR and create a new PR for #207 ? In the new PR, I will remove `BufferDescriptor::not_destroyed_buffers_` because Amber will never recreate buffer for BufferDescriptor anymore.

Issue #197, opened 2018-12-20
Mismatch of multiple updates for a SSBO between VkRunner script and Amber script


jaebaek:
### Multiple updates for a SSBO in VkRunner

As debated [here](https://github.com/Igalia/vkrunner/issues/46), VkRunner applies multiple updates for a SSBO and uses the final one for all computes or draws.

For example,
```
ssbo 0 subdata float 0 1.0
compute 1 1 1
ssbo 0 subdata float 0 2.0
compute 1 1 1
ssbo 0 subdata float 36 1.0 2.0 3.0 ... ( more than 4000 numbers) ... 4096.0
compute 1 1 1
probe ssbo float 0 ~= 2.0
ssbo 0 subdata float 4000 1.0 2.0 3.0 ... ( more than 4000 numbers) ... 4096.0
compute 1 1 1
probe ssbo float 4000 ~= 1.0
```

This script has the same behavior as the following script does

```
ssbo 0 subdata float 0 1.0
ssbo 0 subdata float 0 2.0
ssbo 0 subdata float 36 1.0 2.0 3.0 ... ( more than 4000 numbers) ... 4096.0
ssbo 0 subdata float 4000 1.0 2.0 3.0 ... ( more than 4000 numbers) ... 4096.0

compute 1 1 1
compute 1 1 1
compute 1 1 1
probe ssbo float 0 ~= 2.0
compute 1 1 1
probe ssbo float 4000 ~= 1.0
```

### Multiple updates for a SSBO in Amber script

Amber script will have different semantics for those multiple updates for a SSBO.
It will apply updates immediately for each compute or draw.

For example,
```
ssbo 0 subdata float 0 1.0
compute 1 1 1
ssbo 0 subdata float 0 2.0
compute 1 1 1
ssbo 0 subdata float 36 1.0 2.0 3.0 ... ( more than 4000 numbers) ... 4096.0
compute 1 1 1
```

In this script,
* the first `compute` sees SSBO at `0:0` as `1.0`
* the second one sees `2.0` if the first compute does not change the SSBO
* the third one sees `2.0 0 0 ... 0 1.0 2.0 3.0 ... ( more than 4000 numbers) ... 4096.0` if the first and second computes does not change the SSBO

If each `compute` writes something to SSBO, it also remains in the SSBO.

### How to solve mismatch between those two scripts

It must be handled by executor layer and we will use the same Vulkan engine for both scripts.
If it is VkRunner script, the executor must call all `DoBuffer` first before doing anything.
[...]

PR #198, opened 2018-12-20
Documentation for multiple updates for a buffer

jaebaek:
Related to #197

### DIFF
--- a/docs/multiple_updates_for_same_buffer.md
+++ b/docs/multiple_updates_for_same_buffer.md
@@ -1,0 +1,76 @@
+This is copied from #197
+
+### Multiple updates for a SSBO in VkRunner
+
+As debated [here](https://github.com/Igalia/vkrunner/issues/46),
+VkRunner applies multiple updates for a SSBO and uses the final
+one for all computes or draws.
+
+For example,
+```
+ssbo 0 subdata float 0 1.0
+compute 1 1 1
+ssbo 0 subdata float 0 2.0
+compute 1 1 1
+ssbo 0 subdata float 36 1.0 2.0 3.0 \
+                        ... ( more than 4000 numbers) ... 4096.0
+compute 1 1 1
+probe ssbo float 0 ~= 2.0
+ssbo 0 subdata float 4000 1.0 2.0 3.0 \
+                          ... ( more than 4000 numbers) ... 4096.0
+compute 1 1 1
+probe ssbo float 4000 ~= 1.0
+```
+
+This script has the same behavior as the following script does
+
+```
+ssbo 0 subdata float 0 1.0
+ssbo 0 subdata float 0 2.0
+ssbo 0 subdata float 36 1.0 2.0 3.0 \
+                        ... ( more than 4000 numbers) ... 4096.0
+ssbo 0 subdata float 4000 1.0 2.0 3.0 \
+                          ... ( more than 4000 numbers) ... 4096.0
+
+compute 1 1 1
+compute 1 1 1
+compute 1 1 1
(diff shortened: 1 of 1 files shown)


docs/multiple_updates_for_same_buffer.md
```diff
@@ -0,0 +1,76 @@
...
+                        ... ( more than 4000 numbers) ... 4096.0
+compute 1 1 1
+probe ssbo float 0 ~= 2.0
+ssbo 0 subdata float 4000 1.0 2.0 3.0 \
+                          ... ( more than 4000 numbers) ... 4096.0
+compute 1 1 1
+probe ssbo float 4000 ~= 1.0
+```
+
+This script has the same behavior as the following script does
```
dj2: That's not quite correct. It uses all of them up to the probe. The probe commands submit the command buffer which collapses all the ssbo buffers. So, the first 3 ssbo commands will be used with the first 3 compute calls. The 4th ssbo command will be used with the 4th compute command.

docs/multiple_updates_for_same_buffer.md
```diff
@@ -0,0 +1,76 @@
...
+
+```
+ssbo 0 subdata float 0 1.0
+ssbo 0 subdata float 0 2.0
+ssbo 0 subdata float 36 1.0 2.0 3.0 \
+                        ... ( more than 4000 numbers) ... 4096.0
+ssbo 0 subdata float 4000 1.0 2.0 3.0 \
```
dj2: This should be after the first probe

docs/multiple_updates_for_same_buffer.md
```diff
@@ -0,0 +1,76 @@
...
+* the second one sees `2.0` if the first compute does not change the
+  SSBO
+* the third one sees
+  `2.0 0 0 ... 0 1.0 2.0 3.0 ... ( more than 4000 numbers) ... 4096.0`
+  if the first and second computes does not change the SSBO
+
+If each `compute` writes something to SSBO, it also remains in the SSBO.
+
+
+### How to solve mismatch between those two scripts
```
dj2: I don't think we should do anything. Just run it, with a note in the README that the behaviour is different. I don't think we want to be adjusting the script on the fly.

docs/multiple_updates_for_same_buffer.md
```diff
@@ -0,0 +1,76 @@
...
+compute 1 1 1
+compute 1 1 1
+probe ssbo float 0 ~= 2.0
+compute 1 1 1
+probe ssbo float 4000 ~= 1.0
+```
+
+### Multiple updates for a SSBO in Amber script
+
+Amber script will have different semantics for those multiple
```
dneto0: If Amber is running a vkscript then it should behave the same as VkRunner.

docs/multiple_updates_for_same_buffer.md
jaebaek: As @dneto0 mentioned, Amber must behave the same as VkRunner does if it is VkRunner script. However, Amber script will have different semantics to make users more clear. To overcome this mismatch, I recommend that we have slight different executors for Amber script and VkRunner script while we do not change Vulkan engine.

docs/multiple_updates_for_same_buffer.md
dj2: I don't think we want follow what VkRunner does here as the script is confusing. I think we should deviate from VkRunner in this case. We just have to document that we have deviated.

dj2:
I'm going to close this for now. Please re-open if it's still relevant.

PR #202, opened 2018-12-21
Add test cases for compute shaders compiled by clspv

jaebaek:


dj2:
In what way? Why not just add the test cases and name them appropriately. It doesn't matter where they came from or what's in them.

dj2:
Do these test cases test something in amber, or are they for testing other things? I think amber should only hold test cases related to testing amber. For test cases for other things we should create a new repo.

jaebaek:
Ok. This is not for testing amber itself. I will close this PR.

PR #203, opened 2018-12-21
Vulkan: support push constant

### LINKED ISSUE #149: Add push constant to handle uniform command of VkRunner

jaebaek:
When executor calls `DoBuffer()` of Vulkan engine for push constant,
Vulkan engine keeps the information of what values and at which
offset it must put constants in |push_constant_data_| of Pipeline
class. Vulkan engine conducts push constant i.e., vkCmdPushConstants
when executing `draw` or `compute` based on information in
|push_constant_data_|.

Fixes #149

### DIFF
--- a/src/vulkan/buffer_descriptor.cc
+++ b/src/vulkan/buffer_descriptor.cc
@@ -24,19 +24,6 @@
 
 namespace amber {
 namespace vulkan {
-namespace {
-
-// TODO(jaebaek): Make this as a protected method of Descriptor.
-template <typename T>
-void SetValueForBuffer(void* memory, const std::vector<Value>& values) {
-  T* ptr = static_cast<T*>(memory);
-  for (const auto& v : values) {
-    *ptr = static_cast<T>(v.IsInteger() ? v.AsUint64() : v.AsDouble());
-    ++ptr;
-  }
-}
-
-}  // namespace
 
 BufferDescriptor::BufferDescriptor(DescriptorType type,
                                    VkDevice device,
@@ -48,36 +35,6 @@ BufferDescriptor::BufferDescriptor(DescriptorType type,
 }
 
 BufferDescriptor::~BufferDescriptor() = default;
-
-// TODO(jaebaek): Add unittests for this method.
-void BufferDescriptor::FillBufferWithData(const BufferData& data) {
-  uint8_t* ptr =
-      static_cast<uint8_t*>(buffer_->HostAccessibleMemoryPtr()) + data.offset;
-  switch (data.type) {
-    case DataType::kInt8:
-    case DataType::kUint8:
-      SetValueForBuffer<uint8_t>(ptr, data.values);
-      break;
-    case DataType::kInt16:
-    case DataType::kUint16:
-      SetValueForBuffer<uint16_t>(ptr, data.values);
-      break;
-    case DataType::kInt32:
-    case DataType::kUint32:
-      SetValueForBuffer<uint32_t>(ptr, data.values);
-      break;
-    case DataType::kInt64:
-    case DataType::kUint64:
-      SetValueForBuffer<uint64_t>(ptr, data.values);
-      break;
-    case DataType::kFloat:
-      SetValueForBuffer<float>(ptr, data.values);
-      break;
-    case DataType::kDouble:
-      SetValueForBuffer<double>(ptr, data.values);
-      break;
-  }
-}
 
 Result BufferDescriptor::CreateOrResizeIfNeeded(
     VkCommandBuffer command,
@@ -135,8 +92,10 @@ void BufferDescriptor::UpdateResourceIfNeeded(VkCommandBuffer command) {
   if (buffer_data_queue.empty())
     return;
 
+  HandleValueWithMemory memory_updater;
   for (const auto& data : buffer_data_queue) {
-    FillBufferWithData(data);
+    memory_updater.UpdateMemoryWithData(buffer_->HostAccessibleMemoryPtr(),
+                                        data);
   }
   ClearBufferDataQueue();
 
--- a/src/vulkan/handle_value_with_memory.cc
+++ b/src/vulkan/handle_value_with_memory.cc
@@ -1,0 +1,66 @@
+// Copyright 2018 The Amber Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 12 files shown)


src/vulkan/buffer_descriptor.cc
```diff
@@ -49,36 +36,6 @@ BufferDescriptor::BufferDescriptor(DescriptorType type,
 
 BufferDescriptor::~BufferDescriptor() = default;
```
dj2: Do push constants not map to the buffer concept? Can we not just subclass buffer for this?

src/vulkan/handle_value_with_memory.cc
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+}  // namespace
+
+HandleValueWithMemory::HandleValueWithMemory() = default;
+
+HandleValueWithMemory::~HandleValueWithMemory() = default;
+
+void HandleValueWithMemory::UpdateMemoryWithData(void* memory,
```
dneto0: It seems somewhat unsafe to have an interface that disables the possibility of checking that you did not overflow the memory buffer.

src/vulkan/pipeline.h
```diff
@@ -105,6 +111,7 @@ class Pipeline {
     std::vector<std::unique_ptr<Descriptor>> descriptors_;
   };
 
+  VkPushConstantRange GetPushConstantRange();
```
dneto0: **METHOD COMMENTS PLEASE**

dneto0: Such as, are any stage bits set?
What's the offset?

src/vulkan/pipeline.cc
```diff
@@ -172,6 +172,45 @@ Result Pipeline::CreateDescriptorSets() {
...
+  if (push_constant_data_.empty())
+    return {};
+
+  auto it =
+      std::min_element(push_constant_data_.begin(), push_constant_data_.end(),
+                       [](const BufferData& a, const BufferData& b) {
+                         return a.offset < b.offset;
+                       });
+  if (it == push_constant_data_.end())
+    return {};
```
dneto0: If I understand std::min_element correctly, this case should never happen.   It seems harmless to keep, except it's slightly confusing.  Maybe I would add an assert here that it isn't the end iterator.

src/vulkan/pipeline.cc
```diff
@@ -172,6 +172,45 @@ Result Pipeline::CreateDescriptorSets() {
...
+  uint32_t first_offset = it->offset;
+
+  it = std::max_element(
+      push_constant_data_.begin(), push_constant_data_.end(),
+      [](const BufferData& a, const BufferData& b) {
+        return a.offset + static_cast<uint32_t>(a.size_in_bytes) <
+               b.offset + static_cast<uint32_t>(b.size_in_bytes);
+      });
+  if (it == push_constant_data_.end())
+    return {};
```
dneto0: Same here. Assert the iterator is not the end iterator.

src/vulkan/pipeline.cc
```diff
@@ -228,6 +273,58 @@ Result Pipeline::UpdateDescriptorSetsIfNeeded() {
...
+Result Pipeline::AddPushConstant(const BufferCommand* command) {
+  if (!command->IsPushConstant())
+    return Result(
+        "Pipeline::AddPushConstant BufferCommand type is not push constant");
+
+  // TODO(jaebaek): push constant -> compute -> push constant -> compute
+  //                The second compute maybe have different push constant
+  //                ranges. Doublecheck if we update |pipeline_layout_| properly
+  //                in that case.
+  if (descriptor_related_objects_already_created_) {
```
dneto0: when you resolve the TODO, I suggest making this block of code a single call to destroy the needed objects.

src/vulkan/pipeline.cc
```diff
@@ -228,6 +273,58 @@ Result Pipeline::UpdateDescriptorSetsIfNeeded() {
...
+      vkDestroyDescriptorPool(device_, info.pool, nullptr);
+    }
+
+    vkDestroyPipelineLayout(device_, pipeline_layout_, nullptr);
+    vkDestroyPipeline(device_, pipeline_, nullptr);
+
+    descriptor_related_objects_already_created_ = false;
+  }
+
+  push_constant_data_.push_back({command->GetDatumType().GetType(),
```
dneto0: Suggest emplace_back here since you're doing the initialization and a copy, and emplace_back would save you the copy.

src/vulkan/pipeline.h
```diff
@@ -48,6 +49,10 @@ class Pipeline {
 
   Result AddDescriptor(const BufferCommand*);
 
+  // Add information of how and what to do with push constant to
```
dneto0: // Add a new set of values in an offset range to the push constants to be used on the next pipeline execution.

src/vulkan/pipeline.h
```diff
@@ -114,6 +121,10 @@ class Pipeline {
   PipelineType pipeline_type_;
   std::vector<DescriptorSetInfo> descriptor_set_info_;
   std::vector<VkPipelineShaderStageCreateInfo> shader_stage_info_;
+
+  // Keep the information of what and how to conduct push constant.
```
dneto0: Please add: 

// These are applied from lowest index to highest index, so that if address ranges overlap, then the later values take effect.

src/vulkan/handle_value_with_memory.cc
jaebaek: updated.

src/vulkan/handle_value_with_memory.cc
```diff
@@ -0,0 +1,66 @@
...
+                                                 const BufferData& data) {
+  uint8_t* ptr = static_cast<uint8_t*>(memory) + data.offset;
+  switch (data.type) {
+    case DataType::kInt8:
+    case DataType::kUint8:
+      SetValueForBuffer<uint8_t>(ptr, data.values);
```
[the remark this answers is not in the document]
jaebaek: updated

src/vulkan/pipeline.cc
jaebaek: done

src/vulkan/pipeline.cc
jaebaek: done

src/vulkan/pipeline.cc
jaebaek: done

src/vulkan/pipeline.cc
jaebaek: done

src/vulkan/pipeline.h
jaebaek: done

src/vulkan/pipeline.h
jaebaek: done

src/vulkan/pipeline.h
jaebaek: done

PR #205, opened 2018-12-28
Vulkan: reuse device using config

### LINKED ISSUE #124: Initialize queue, queue family index, physical memory properties when we reuse VkDevice
Currently, we do not initialize queue, queue family index, physical memory properties when we reuse VkDevice. It is a bug.

`EngineVulkan::InitializeWithDevice()` must have `VkPhysicalDevice` as an additional argument.

jaebaek:
Fixes #124


jaebaek:
@dj2 @dneto0 as a proof of concept implementation, in `samples/config_helper.cc`, I just reuse `Device` class in `src/vulkan/device.cc`. This makes the layering weird but simple way to present how to use config for reusing `VkDevice`. Do you think I should not use `src/vulkan/device.cc` in `samples/config_helper.cc` and implement creation of `VkDevice` under `samples/` ?

jaebaek:
In addition, it assume that it reuses `VkDevice` only when multiple tests are given.

dj2:
Please don't re-use the src/vulkan/device from samples. The samples should contain all the code they need.

jaebaek:
Updated. PTAL

samples/amber.cc
```diff
@@ -45,6 +49,10 @@ const char kUsage[] = R"(Usage: amber [options] SCRIPT
   -e <engine>    -- Specify graphics engine: vulkan, dawn. Default is vulkan.
   -V, --version  -- Output version information for Amber and libraries.
   -h             -- This help text.
+
+ Note:
```
dj2: I'd just remove this note

samples/amber.cc
```diff
@@ -35,7 +39,7 @@ struct Options {
   amber::EngineType engine = amber::EngineType::kVulkan;
 };
 
-const char kUsage[] = R"(Usage: amber [options] SCRIPT
+const char kUsage[] = R"(Usage: amber [options] SCRIPT [more SCRIPTs]
```
dj2: amber [options] SCRIPT [SCRIPTS...]

samples/amber.cc
```diff
@@ -108,7 +116,7 @@ bool ParseArgs(const std::vector<std::string>& args, Options* opts) {
       std::cerr << "Unrecognized option " << arg << std::endl;
       return false;
     } else {
-      opts->input_filename = args[i];
+      opts->input_filenames.push_back(args[i]);
```
dj2: This should verify the filename is not empty otherwise I could do `amber "" "" ""`

samples/config_helper.cc
```diff
@@ -0,0 +1,487 @@
+// Copyright 2018 The Amber Authors.
```
dj2: 2019

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
   amber_options.engine = options.engine;
-  result = am.Execute(&recipe, amber_options);
-  if (!result.IsSuccess()) {
-    std::cerr << result.Error() << std::endl;
-    return 1;
+  if (options.input_filenames.size() > 1UL && !options.parse_only) {
+    amber_options.config = amber::MakeUnique<amber::VulkanEngineConfig>();
+
+    amber::VulkanEngineConfig* config =
+        static_cast<amber::VulkanEngineConfig*>(amber_options.config.get());
```
dj2: This cast shouldn't be needed as you made the specific VulkanEngineConfig type

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
+  amber::Result result;
+  sample::ConfigHelper config_helper;
   amber::Options amber_options;
   amber_options.engine = options.engine;
-  result = am.Execute(&recipe, amber_options);
-  if (!result.IsSuccess()) {
-    std::cerr << result.Error() << std::endl;
-    return 1;
+  if (options.input_filenames.size() > 1UL && !options.parse_only) {
+    amber_options.config = amber::MakeUnique<amber::VulkanEngineConfig>();
```
dj2: Why make this a unique pointer instead of just copying the struct?

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
+    if (!result.IsSuccess()) {
+      std::cerr << result.Error() << std::endl;
+      return 1;
+    }
   }
 
-  if (!options.buffer_filename.empty()) {
-    // TODO(dsinclair): Write buffer file
-    assert(false);
+  std::vector<amber::Amber> ambers(options.input_filenames.size());
```
dj2: This isn't needed, you don't need to store the Amber object, just create a new one later

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
+  std::vector<amber::Recipe> recipes(options.input_filenames.size());
+  std::vector<std::string> failures;
+  for (size_t i = 0; i < options.input_filenames.size(); ++i) {
+    if (options.input_filenames.size() > 1UL)
+      std::cout << "case " << options.input_filenames[i] << ": run... ";
```
dj2: Remove this if() we aren't running anything yet, just parsing

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
+    }
   }
 
-  if (!options.buffer_filename.empty()) {
-    // TODO(dsinclair): Write buffer file
-    assert(false);
+  std::vector<amber::Amber> ambers(options.input_filenames.size());
+  std::vector<amber::Recipe> recipes(options.input_filenames.size());
+  std::vector<std::string> failures;
+  for (size_t i = 0; i < options.input_filenames.size(); ++i) {
```
dj2: for (auto& file : options.input_filenames)

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
+    if (options.input_filenames.size() > 1UL)
+      std::cout << "case " << options.input_filenames[i] << ": run... ";
+
+    auto data = ReadFile(options.input_filenames[i]);
+    if (data.empty())
+      return 1;
```
dj2: This should report an error

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
+        std::cerr << result.Error() << std::endl;
+        return 1;
+      }
+
+      std::cout << "fail" << std::endl;
+      failures.push_back(options.input_filenames[i]);
+      std::cerr << result.Error() << std::endl << std::endl;
+      continue;
+    }
+
```
dj2: The inital loop should end here so we do all the parsing in one pass The if (!success) above can just always output the filename and error. The Recipe should be appended to the recipes list.

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
+    if (options.parse_only)
+      continue;
+
```
dj2: for (auto& recipe : recipes)

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
+    result = am.Execute(&recipe, amber_options);
+    if (!result.IsSuccess()) {
+      if (options.input_filenames.size() == 1UL) {
+        std::cerr << result.Error() << std::endl;
+        return 1;
+      }
+
+      std::cout << "fail" << std::endl;
```
dj2: Remove this as it doesn't add anything

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
+    }
+
+    if (options.parse_only)
+      continue;
+
+    result = am.Execute(&recipe, amber_options);
+    if (!result.IsSuccess()) {
+      if (options.input_filenames.size() == 1UL) {
+        std::cerr << result.Error() << std::endl;
+        return 1;
```
dj2: This and the below can just output the error to cerr and continue.

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
+      std::cout << "fail" << std::endl;
+      failures.push_back(options.input_filenames[i]);
+      std::cerr << result.Error() << std::endl << std::endl;
+      continue;
+    }
+
+    if (!options.buffer_filename.empty()) {
+      // TODO(dsinclair): Write buffer file
+      if (options.input_filenames.size() > 1UL)
+        std::cout << "fail" << std::endl;
```
dj2: Remove

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
+      assert(false);
+    }
+
+    if (!options.image_filename.empty()) {
+      // TODO(dsinclair): Write image file
+      if (options.input_filenames.size() > 1UL)
+        std::cout << "fail" << std::endl;
```
dj2: Remove

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
+      assert(false);
+    }
+
+    if (options.input_filenames.size() > 1UL)
+      std::cout << "pass" << std::endl;
```
dj2: Remove

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
   }
 
-  if (!options.image_filename.empty()) {
-    // TODO(dsinclair): Write image file
-    assert(false);
+  if (options.input_filenames.size() > 1UL) {
```
dj2: Just do this always

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
+
+    for (const auto& failure : failures)
+      std::cout << "  " << failure << std::endl;
+
+    std::cout << "\nSummary: "
+              << (options.input_filenames.size() - failures.size()) << " pass, "
+              << failures.size() << " fail" << std::endl;
+
+    if (!options.parse_only)
+      config_helper.Shutdown();
```
dj2: We should be conditionally creating  as wel

src/vulkan/device.cc
```diff
@@ -487,19 +487,29 @@ bool AreAllExtensionsSupported(
 }  // namespace
 
 Device::Device() = default;
-Device::Device(VkDevice device) : device_(device), destroy_device_(false) {}
+Device::Device(VkPhysicalDevice physical_device, VkDevice device)
+    : physical_device_(physical_device),
+      device_(device),
+      has_device_ownership_(false) {}
```
dj2: Let's leave this as destroy_device_

src/vulkan/device.cc
```diff
@@ -487,19 +487,29 @@ bool AreAllExtensionsSupported(
...
   }
 }
 
 Result Device::Initialize(const std::vector<Feature>& required_features,
                           const std::vector<std::string>& required_extensions) {
-  if (device_ == VK_NULL_HANDLE) {
+  if (has_device_ownership_ &&
+      (physical_device_ != VK_NULL_HANDLE || device_ != VK_NULL_HANDLE)) {
+    return Result(
+        "Vulkan: Device::Initialize Device class has VkDevice ownership but "
```
dj2: I don't understand why this happens here, we don't pass physical_device_ or device_ into this method?

include/amber/amber_vulkan.h
```diff
@@ -15,14 +15,21 @@
...
 
+#include "amber/amber.h"
 #include "vulkan/vulkan.h"
 
 namespace amber {
 
 /// Configuration for the Vulkan Engine.
 struct VulkanEngineConfig : public EngineConfig {
+  /// The VkInstance to use for the tests.
+  VkInstance instance = VK_NULL_HANDLE;
```
dj2: What is the instance for?

samples/config_helper.cc
```diff
@@ -0,0 +1,487 @@
...
+  info.ppEnabledExtensionNames = kRequiredExtensions;
+
+  if (vkCreateDevice(physical_device_, &info, nullptr, &device_) != VK_SUCCESS)
+    return amber::Result("Vulkan::Calling vkCreateDevice Fail");
+
+  return {};
+}
+
+std::pair<amber::Result, amber::VulkanEngineConfig>
+ConfigHelper::CreateVulkanConfig() {
```
dj2: Why not amber::Result COnfigHelper::CreateVulkanConfig(amber::VulkanEngineConfig* config) then we just fill in the config?

src/vulkan/device.cc
```diff
@@ -511,13 +521,33 @@ Result Device::Initialize(const std::vector<Feature>& required_features,
     r = CreateDevice(required_features, required_extensions);
     if (!r.IsSuccess())
       return r;
-  }
+  } else {
+    if (!AreAllRequiredFeaturesSupported(physical_device_, required_features)) {
```
dj2: The CTS can give us a bunch of this information, see https://github.com/KhronosGroup/VK-GL-CTS/blob/master/external/vulkancts/modules/vulkan/vktTestCase.hpp. Would it be easier to just pass in the features, extensions and queue instead of re-querying? In that case, do we need the physical device?

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
-  result = am.Execute(&recipe, amber_options);
-  if (!result.IsSuccess()) {
-    std::cerr << result.Error() << std::endl;
-    return 1;
+  if (options.input_filenames.size() > 1UL && !options.parse_only) {
+    amber_options.config = amber::MakeUnique<amber::VulkanEngineConfig>();
+
+    amber::VulkanEngineConfig* config =
+        static_cast<amber::VulkanEngineConfig*>(amber_options.config.get());
+    std::tie(result, *config) = config_helper.CreateVulkanConfig();
```
dj2: result = config_helper.InitializeVulkanConfig(amber_options.config.get());

samples/amber.cc
```diff
@@ -178,42 +186,101 @@ int main(int argc, const char** argv) {
...
-  if (!result.IsSuccess()) {
-    std::cerr << result.Error() << std::endl;
-    return 1;
-  }
-
-  if (options.parse_only)
-    return 0;
-
+  amber::Result result;
+  sample::ConfigHelper config_helper;
```
dj2: auto* config = config_helper.CreateConfig(engine);

CreateConfig -> unique_ptr<AmberConfig>

include/amber/amber_vulkan.h
```diff
@@ -15,16 +15,43 @@
...
+
+  /// The VkQueue to be used for the tests.
+  VkQueue queue;
+
+  /// The VkDevice to be used for the tests.
+  VkDevice device = VK_NULL_HANDLE;
 };
 
+/// Returns required physical device features from |recipes|.
+VkPhysicalDeviceFeatures GetRequiredVulkanFeatures(
```
dj2: These should be methods on Recipe themselves.

std::vector<std::string> Recipe::GetRequiredFeatures()
std::vector<std::string> Recipe::GetRequiredExtensions()

include/amber/amber_vulkan.h
```diff
@@ -15,16 +15,43 @@
...
+  /// The VkPhysicalDevice to be used for the tests.
+  VkPhysicalDevice physical_device = VK_NULL_HANDLE;
+
+  /// Physical device features available for |physical_device|.
+  VkPhysicalDeviceFeatures available_features = {};
+
+  /// Physical device extensions available for |physical_device|.
+  std::vector<std::string> available_extensions;
+
+  /// The given queue family index to be used for the tests.
```
dj2: I'd remove the 'for the tests' part of each of these.

src/vulkan/device.cc
```diff
@@ -450,35 +448,44 @@ bool AreAllRequiredFeaturesSupported(
...
-  if (required_extensions.empty())
-    return true;
-
+std::vector<std::string> GetAvailableExtensions(
+    const VkPhysicalDevice& physical_device) {
+  std::vector<std::string> available_extensions;
   uint32_t available_extension_count = 0;
   std::vector<VkExtensionProperties> available_extension_properties;
 
   if (vkEnumerateDeviceExtensionProperties(physical_device, nullptr,
```
dj2: Why do we need to do this if the extensions are passed in now?

include/amber/amber_vulkan.h
```diff
@@ -15,16 +15,43 @@
...
 #include "vulkan/vulkan.h"
 
 namespace amber {
 
 /// Configuration for the Vulkan Engine.
 struct VulkanEngineConfig : public EngineConfig {
-  /// The VkDevice to use for the tests.
-  VkDevice device;
+  /// The VkPhysicalDevice to be used for the tests.
+  VkPhysicalDevice physical_device = VK_NULL_HANDLE;
```
dj2: I thought if we passed in the extensions, queue and features we wouldn't need physical_device anymore?

samples/amber.cc
```diff
@@ -108,7 +111,8 @@ bool ParseArgs(const std::vector<std::string>& args, Options* opts) {
       std::cerr << "Unrecognized option " << arg << std::endl;
       return false;
     } else {
```
dj2: else if (!arg.empty())

samples/amber.cc
```diff
@@ -178,43 +182,88 @@ int main(int argc, const char** argv) {
...
     std::cerr << "Input file must be provided." << std::endl;
     return 2;
   }
 
-  auto data = ReadFile(options.input_filename);
-  if (data.empty())
-    return 1;
+  amber::Result result;
+  std::vector<std::string> failures;
+  std::vector<std::pair<std::unique_ptr<amber::Recipe>, std::string>>
```
dj2: struct  Data {
  std::string file
  unique_ptr<Recipe> recipe
}
vector<Data> recipes;

samples/amber.cc
```diff
@@ -178,43 +182,88 @@ int main(int argc, const char** argv) {
...
 
+  sample::ConfigHelper config_helper;
   amber::Options amber_options;
   amber_options.engine = options.engine;
-  result = am.Execute(&recipe, amber_options);
-  if (!result.IsSuccess()) {
-    std::cerr << result.Error() << std::endl;
-    return 1;
-  }
+  if (options.input_filenames.size() > 1UL) {
```
dj2: This check isn't needed, as you may have less things in recipes then in input_files depending on if the parse failed.

samples/amber.cc
```diff
@@ -178,43 +182,88 @@ int main(int argc, const char** argv) {
...
+    std::vector<const amber::Recipe*> recipes;
+    for (const auto& recipe_and_file : recipe_and_files)
+      recipes.push_back(recipe_and_file.first.get());
 
-  if (!options.buffer_filename.empty()) {
-    // TODO(dsinclair): Write buffer file
-    assert(false);
+    amber_options.config =
+        config_helper.CreateConfig(amber_options.engine, recipes);
```
dj2: I'd rather this passed in a vector<string> extensions and vector<string> features then a vector of recipes.

samples/amber.cc
```diff
@@ -178,43 +182,88 @@ int main(int argc, const char** argv) {
...
-    assert(false);
+  for (const auto& recipe_and_file : recipe_and_files) {
+    const auto* recipe = recipe_and_file.first.get();
+    const auto& file = recipe_and_file.second;
+
+    amber::Amber am;
+    result = am.Execute(recipe, amber_options);
+    if (!result.IsSuccess()) {
+      std::cerr << "case " << file << ": run fail" << std::endl;
+      std::cerr << result.Error() << std::endl << std::endl;
```
dj2: These two lines can just be:

std::cerr <<  file << ": " << result.Error() << std::endl;

samples/amber.cc
```diff
@@ -178,43 +182,88 @@ int main(int argc, const char** argv) {
...
+      // TODO(dsinclair): Write image file
+      assert(false);
+    }
   }
 
-  return 0;
+  if (!failures.empty())
+    std::cout << "\nSummary of Failures:" << std::endl;
+
+  for (const auto& failure : failures)
```
dj2: may as well put this in the if () above

samples/amber.cc
```diff
@@ -178,43 +182,88 @@ int main(int argc, const char** argv) {
...
+    std::cout << "  " << failure << std::endl;
+
+  std::cout << "\nSummary: "
+            << (options.input_filenames.size() - failures.size()) << " pass, "
+            << failures.size() << " fail" << std::endl;
+
+  config_helper.Shutdown();
+
+  return failures.empty() ? 0 : 1;
```
dj2: return !failures.empty();

samples/config_helper.cc
```diff
@@ -0,0 +1,460 @@
...
+#include <cassert>
+#include <set>
+#include <string>
+#include <vector>
+
+#include "src/make_unique.h"
+
+namespace sample {
+
+#if AMBER_ENGINE_VULKAN
```
dj2: The code in these blocks should be in a config_helper_vulkan.h and .cc file which are pulled in conditionally in the cmake files instead of preprocssored out.

samples/config_helper.h
```diff
@@ -0,0 +1,87 @@
+// Copyright 2018 The Amber Authors.
```
dj2: 2019

samples/config_helper.h
```diff
@@ -0,0 +1,87 @@
...
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "amber/amber.h"
+
+#if AMBER_ENGINE_VULKAN
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wzero-as-null-pointer-constant"
```
dj2: These programs should be in amber_vulkan around the vulkan/vulkan.h include

samples/config_helper.h
```diff
@@ -0,0 +1,87 @@
...
+
+  // Create instance and device and return them as amber::EngineConfig.
+  // |recipes| contains the information of required features and
+  // extensions.
+  std::unique_ptr<amber::EngineConfig> CreateConfig(
+      amber::EngineType engine,
+      const std::vector<const amber::Recipe*>& recipes);
+
+  // Destroy Vulkan instance and device.
+  void Shutdown();
```
dj2: This should probably work along the lines of ConfigHelper has a unique_ptr<ConfileHelperImpl>. we then have a ConfigHelperVulkan : public ConfigHelperImpl which has the vulkan stuff. The ConfigHelper shutdown will just proxy to the impl. The create will use impl to create what's needed

src/vulkan/amber_vulkan.cc
```diff
@@ -0,0 +1,236 @@
+// Copyright 2018 The Amber Authors.
```
dj2: 2019

tests/run_tests.py
```diff
@@ -53,13 +53,8 @@ def RunTest(self, tc):
     cmd += [tc.GetInputPath()]
 
     try:
-      err = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
-      if err != "" and not tc.IsExpectedFail():
-        sys.stdout.write(err)
-        return False
-
+      subprocess.check_output(cmd)
```
dj2: I thought we decided to leave this as is and make the summary print a command line option?

include/amber/amber_vulkan.h
jaebaek: We need `VkPhysicalDeviceFeatures` which is not `std::vector<std::string>`. It is Vulkan specific object.
In addition, for WebGPU, we might need other method.

We must add `#if AMBER_ENGINE_VULKAN` when we use `VkPhysicalDeviceFeatures`.
I intended to put all vulkan related APIs in include/amber/amber_vulkan.h.

An alternative is to define a new structure `EngineFeature` and use `struct VulkanEngineFeature : public EngineFeature { VkPhysicalDeviceFeatures features; }` and add `std::unique<EngineFeature> Recipe::GetRequiredFeatures()` to `Recipe`.

dj2: We can convert to a VkPhysicalDevice inside the amber code, same as is done currently. The Recipe should return strings I think.

In amber_vulkan we will need to pass VkPhysicalDeviceFeatures if that's what CTS provides, but that's already contained in vulkan specific code and doesn't need an AMBER_ENGINE_VULKAN ifdef.

dj2: (The reason for returning strings instead of Vk items is that we'd have to implement the recipe code twice to do a Dawn equivalent. Returning strings will let us re-use the same logic for multiple backends)

samples/amber.cc
dj2: Always create the device externally, we don't need to special case if there is a single filename.

samples/amber.cc
```diff
@@ -15,30 +15,36 @@
...
   amber::EngineType engine = amber::EngineType::kVulkan;
 };
 
-const char kUsage[] = R"(Usage: amber [options] SCRIPT
+const char kUsage[] = R"(Usage: amber [options] SCRIPT [SCRIPTS...]
 
  options:
-  -p             -- Parse input files only; Don't execute
+  -p             -- Parse input files only; Don't execute.
+  -s             -- Print summary of pass/failure if any failures exist.
```
dj2: This should always print summary even if there were no failures.

samples/amber.cc
```diff
@@ -178,43 +186,100 @@ int main(int argc, const char** argv) {
...
+    if (!options.image_filename.empty()) {
+      // TODO(dsinclair): Write image file
+      assert(false);
+    }
   }
 
-  if (!options.image_filename.empty()) {
-    // TODO(dsinclair): Write image file
-    assert(false);
+  if (options.show_summary && !failures.empty()) {
```
dj2: if (optins.show_summary) {
  if (!failures.empty())
    std::cout ...

...
}

samples/config_helper_vulkan.h
```diff
@@ -0,0 +1,72 @@
...
+#include "amber/amber_vulkan.h"
+#include "samples/config_helper.h"
+
+namespace sample {
+
+// Child class of ConfigHelperImpl for Vulkan.
+class ConfigHelperVulkan : public ConfigHelperImpl {
+ public:
+  ConfigHelperVulkan();
+  ~ConfigHelperVulkan();
```
dj2: ~ConfigHelperVulkan() override

samples/config_helper.h
```diff
@@ -0,0 +1,66 @@
...
+
+  // Destroy instance and device.
+  virtual void Shutdown() = 0;
+};
+
+// Wrapper of ConfigHelperImpl.
+class ConfigHelper {
+ public:
+  ConfigHelper();
+  ~ConfigHelper();
```
dj2: virtual ~ConfigHelper

src/recipe.cc
```diff
@@ -29,4 +29,16 @@ std::vector<ShaderInfo> Recipe::GetShaderInfo() const {
   return impl_->GetShaderInfo();
 }
 
+std::vector<std::string> Recipe::GetRequiredFeatures() const {
+  if (!impl_)
+    return std::vector<std::string>();
+  return impl_->GetRequiredFeatures();
```
dj2: return impl_ ? impl_->GetRequiredFeatures() : {};

src/recipe.cc
```diff
@@ -29,4 +29,16 @@ std::vector<ShaderInfo> Recipe::GetShaderInfo() const {
...
+}
+
+std::vector<std::string> Recipe::GetRequiredExtensions() const {
+  if (!impl_)
+    return std::vector<std::string>();
+  return impl_->GetRequiredExtensions();
```
dj2: ditto

samples/config_helper_vulkan.h
jaebaek: done

src/recipe.cc
jaebaek: `{}` has compiler errors and done with `std::vector<std::string>()`.

src/recipe.cc
jaebaek: done

PR #217, opened 2019-01-11
Vulkan: readback SSBO/framebuffer right after draw or compute

### LINKED ISSUE #207: Copy back SSBO / framebuffer content right after compute / draw
This issue is related to #200 but we decided to drop the idea described in #200.

In terms of updating buffer including resizing, we will regulate the following rules:
- Put information for buffer into `buffer_data_queue_` of `Descriptor` class.
- Create the actual buffer whenever it is actually required i.e., compute or draw command.
- Copy back SSBO content to `buffer_data_queue_` of `Descriptor` class right after compute / draw and destroy the buffer i.e., host holds the data now.

jaebaek:
Mainly calling `FrameBuffer::CopyColorImageToHost()` after draw
conducts the immediate readback of framebuffer. For the immediate
readback of SSBO, we also read back other descriptors based on
buffer e.g., UBO just for simplicity of implementation. Calling
`Pipeline::ReadbackDescriptorsToHostDataQueue()` after draw and
compute conducts the immediate readback of SSBO.

Fixes #207


src/vulkan/buffer_descriptor.cc
```diff
@@ -80,76 +103,89 @@ void BufferDescriptor::FillBufferWithData(const BufferData& data) {
...
+        "has nullptr host accessible memory");
+  }
+
+  auto size_in_bytes = buffer_->GetSizeInBytes();
+  auto& buffer_data_queue = GetBufferDataQueue();
+  buffer_data_queue.emplace_back();
+  buffer_data_queue.back().offset = 0;
+  buffer_data_queue.back().size_in_bytes = size_in_bytes;
+  buffer_data_queue.back().raw_data =
+      std::unique_ptr<uint8_t>(new uint8_t[size_in_bytes]);
```
dj2: If we control raw_data, why not a vector that we resize?

src/vulkan/buffer_descriptor.cc
```diff
@@ -167,8 +203,35 @@ Result BufferDescriptor::UpdateDescriptorSetIfNeeded(
...
+  auto& buffer_data_queue = GetBufferDataQueue();
+  if (buffer_data_queue.empty()) {
+    info.size_in_bytes = 0;
+    info.cpu_memory = nullptr;
+    return info;
+  }
+
+  size_t size_in_bytes = GetBufferSizeInBytesForQueue(buffer_data_queue);
+
+  auto raw_data = std::unique_ptr<uint8_t>(new uint8_t[size_in_bytes]);
```
dj2: std::vector<uint8_t> raw_data;
raw_data.resize(size_in_bytes);

src/vulkan/buffer_descriptor.cc
```diff
@@ -167,8 +203,35 @@ Result BufferDescriptor::UpdateDescriptorSetIfNeeded(
...
+
+  for (const auto& data : buffer_data_queue) {
+    FillBufferWithData(raw_data.get(), data);
+  }
+  buffer_data_queue.clear();
+
+  buffer_data_queue.emplace_back();
+  buffer_data_queue.back().offset = 0;
+  buffer_data_queue.back().size_in_bytes = size_in_bytes;
+  buffer_data_queue.back().raw_data = std::move(raw_data);
```
dj2: buffer_data_queue.emplace_back({0, size_in_bytes, std::move(raw_data)}; ?

src/vulkan/descriptor.h
```diff
@@ -43,10 +43,15 @@ enum class DescriptorType : uint8_t {
 VkDescriptorType ToVkDescriptorType(DescriptorType type);
 
 struct BufferData {
-  DataType type;
   uint32_t offset;
   size_t size_in_bytes;
-  std::vector<Value> values;
+  DataType type;                      // Primitive type of |values|.
+  std::vector<Value> values;          // Data that has primitive
```
dj2: What is a 'primitive type' ?Why don't we just have Value objects everywhere?

src/vulkan/descriptor.h
```diff
@@ -43,10 +43,15 @@ enum class DescriptorType : uint8_t {
...
+                                      // type. If |raw_data| is not
+                                      // nullptr, it must be empty.
+  std::unique_ptr<uint8_t> raw_data;  // Data without primitive
```
dj2: std::vector<uint8_t>

tests/cases/probe_no_compute_with_multiple_ssbo_commands.amber
```diff
@@ -0,0 +1,37 @@
+# Copyright 2018 The Amber Authors.
```
dj2: 2019

tests/cases/probe_no_compute_with_ssbo.amber
```diff
@@ -0,0 +1,35 @@
+# Copyright 2018 The Amber Authors.
```
dj2: 2019

src/vulkan/descriptor.h
jaebaek: Changed 'primitive type' to type.
`values` will be used for data given by script while `raw_data` will be used for read back from Vulkan buffer.
Since `Value` class is specialized for a value with type, it is not suitable to keep the raw data from Vulkan buffer.

src/vulkan/descriptor.h
jaebaek: done

src/vulkan/descriptor.h
```diff
@@ -100,25 +109,42 @@ class Descriptor {
...
-  // Create new vulkan resource if needed i.e., if it was not created
-  // yet or if we need bigger one. If we recreated it for bigger size,
-  // it records the command for copying the old one's data to the new
-  // one. Note that it only records the command and the actual
+  // Create vulkan resource e.g., buffer or image used for this
+  // descriptor if |buffer_data_queue_| is not empty. This method
+  // assumes that the resource is empty when it is called that means
+  // the resource must be created only when it is actually needed
+  // i.e., compute or draw command and destroyed right after those
+  // commands. Note that it only records the command and the actual
```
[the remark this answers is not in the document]
jaebaek: done

src/vulkan/descriptor.h
```diff
@@ -100,25 +109,42 @@ class Descriptor {
...
   // Record a command for copying data in |buffer_data_queue_| to the
   // resource in device. Note that it only records the command and the
   // actual submission must be done later.
-  virtual void UpdateResourceIfNeeded(VkCommandBuffer command) = 0;
-
-  // Only record the copy command for sending the bound resource
-  // data to the host accessible memory. The actual submission of
-  // the command must be done later.
-  virtual Result SendDataToHostIfNeeded(VkCommandBuffer command) = 0;
+  virtual void CopyDataToResourceIfNeeded(VkCommandBuffer command) = 0;
```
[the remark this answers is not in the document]
jaebaek: done

src/vulkan/descriptor.h
```diff
@@ -100,25 +109,42 @@ class Descriptor {
...
+
+  // Only record the copy command for copying the resource data to
+  // the host accessible memory. The actual submission of the command
+  // must be done later.
```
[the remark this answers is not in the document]
jaebaek: Memory barrier to solve sync bug will be added by [other PR](https://github.com/jaebaek/amber/tree/sync).
I will send the PR right after testing it with Intel GPU.

src/vulkan/descriptor.h
```diff
@@ -100,25 +109,42 @@ class Descriptor {
...
+  virtual Result CopyDataToHost(VkCommandBuffer command) = 0;
```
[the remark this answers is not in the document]
jaebaek: done

tests/cases/probe_no_compute_with_multiple_ssbo_commands.amber
jaebaek: done

tests/cases/probe_no_compute_with_ssbo.amber
jaebaek: done

src/vulkan/descriptor.h
dj2: But, we created the vulkan buffer, so we should already know the type of the buffer we're reading from yea? So, we can create value objects of the correct type based off the buffer?

jaebaek: What if user overwrite a single ssbo with multiple types of values?
```
ssbo 0:0 subdata float 0 1 2 3
ssbo 0:0 subdata int   4 4 5 6
```
In the above example, how do we determine the type of ssbo at offset 4? `int` or `float`?

I doubt this approach because Vulkan buffers e.g., ssbo and ubo do not have types and they are just memory chunk. I think the meaning of value in those buffers must be interpreted by users when probing them. In Vulkan engine layer, I think we should keep it as raw data.

jaebaek: If you think we should not put `raw_data` into `struct BufferData`, we can separate them and keep `raw_data` in `Descriptor` class.

I also prefer this now. Does it sound good to you?

dj2: In that case, you can just read all of the numbers out as a float into the Value object and cast at the caller side if you want something different.

dj2: The main benefit I see of using Values is you only need to store one thing, instead of it being values, or raw if values is empty which I can see causing issues down the line. Just having a single storage place makes it easier to understand what's happening.

jaebaek: Since we don't know what is the meaning of GPU memory when we read it back to CPU memory, we would better keep it as a memory chunk without type. Let me split `struct BufferData` into `struct BufferInput` and `struct BufferOutput`. `struct BufferInput` must be used for data given by user and `struct BufferOutput` must be used for data from GPU memory.

src/vulkan/resource.h
```diff
@@ -29,11 +29,16 @@ namespace vulkan {
...
+struct BufferInput {
+  void UpdateBufferWithValues(void* buffer) const;
+
   uint32_t offset;
   size_t size_in_bytes;
-  std::vector<Value> values;
+  DataType type;              // Type of |values|.
+  std::vector<Value> values;  // Data whose type is |type|. It is
+                              // used to save data given by script.
+                              // If |raw_data| is not empty, it
```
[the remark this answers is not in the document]
jaebaek: done

src/vulkan/resource.cc
```diff
@@ -57,66 +58,75 @@ void SetValuesForBuffer(void* buffer, const std::vector<Value>& values) {
...
-  if (data.size_in_bytes >
-      (size_in_bytes_ - static_cast<size_t>(data.offset))) {
-    return Result(
-        "Vulkan: Resource::UpdateMemoryWithData BufferData offset + size in "
-        "bytes exceeds memory size");
-  }
-
-  uint8_t* ptr = static_cast<uint8_t*>(memory_ptr_) + data.offset;
-  switch (data.type) {
+void BufferInput::UpdateBufferWithValues(void* buffer) const {
```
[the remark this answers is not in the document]
jaebaek: This is also used by `ResourceInfo BufferDescriptor::GetResourceInfo()`. See [this code](https://github.com/google/amber/pull/217/files/a50bc98605eb9a946986be1f313338598687a3d1#diff-193c363dd1e0a72a59ded4594cf70d6cR197).

If we put it in Resource class, we cannot use it in `ResourceInfo BufferDescriptor::GetResourceInfo()`.

PR #218, opened 2019-01-14
Convert API to C++03 from C++11 for CTS compatibility

dj2:
The CTS requires code to not use c++11. This CL converts the various
c++11-ism's in the Amber API to C++03.


dj2:
Anything in `include` must be c++03.

PR #223, opened 2019-01-14
Print line number of failing probe command.

### LINKED ISSUE #219: Improve error report
Currently, we do not provide users with the line number of failure.

For example, if `(A)` has a problem in the following code,
```
probe ssbo int 0  0 == 0
probe ssbo int 0  4 == 2
probe ssbo int 0 16 == 4    # (A)
```
Amber says `...: Verifier failed: 4.000000 == 0.000000, at index 0`.
This does not let users know which line has problem.
In addition, printing `offset` from the buffer is more clear than `index`.

dj2:
This CL changes the verifier to output the line number of a failing
probe command. Each command now stores the line number from the
original script.

Fixes #219

### DIFF
--- a/src/command.h
+++ b/src/command.h
@@ -96,7 +96,7 @@ class Command {
   explicit Command(Type type);
 
   Type command_type_;
-  size_t line_;
+  size_t line_ = 1;
 };
 
 class DrawRectCommand : public Command {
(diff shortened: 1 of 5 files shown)


dj2:
@alan-baker could you try patching this in and seeing if it would have produced the correct line for your script?

PR #224, opened 2019-01-14
Vulkan: support all VkFormat for vertex and frame

### LINKED ISSUE #31: Vulkan: Support read/write of all image formats.

jaebaek:
Fixes #31


jaebaek:
Split`Verifier::IsTexelEqualToExpected()` into tree parts:
* `std::vector<Value> GetActualValuesFromTexel()`
* `std::vector<Value> ScaleTexelValuesIfNeeded()`
* `Verifier::IsTexelEqualToExpected()`

PTAL

src/bit_copy.h
```diff
@@ -45,15 +46,6 @@ class BitCopy {
...
-  // at |buffer| + |length_bytes|, this method ignores it. Note that
-  // this overflow is fine for BitCopy because we have only interest
-  // in data [0, length_bytes) bytes of |buffer|.
-  static void LeftShiftBufferBits(uint8_t* buffer,
-                                  uint8_t length_bytes,
-                                  uint8_t shift_bits);
-
   // Right-shift |length_bytes| bytes of buffer by |shift_bits| bits
   // based on assumption that |shift_bits| < 8. If there is underflow,
   // this method ignores it. Note that this underflow is fine for
```
dj2: RightShiftBufferBits is also removed, yea?

src/CMakeLists.txt
```diff
@@ -91,7 +92,7 @@ if (${AMBER_ENABLE_TESTS})
   )
 
   if (${Vulkan_FOUND})
```
dj2: This shouldn't be guarded by Vulkan anymore.

src/bit_copy.cc
```diff
@@ -0,0 +1,230 @@
+// Copyright 2018 The Amber Authors.
```
dj2: 2019

src/bit_copy.cc
```diff
@@ -0,0 +1,230 @@
...
+  // and if |bits| is 64, |dst_bit_offset| must be 0. No component
+  // has |bits| bigger than 64.
+  data <<= dst_bit_offset;
+
+  CopyBits(dst, data, dst_bit_offset, bits);
+  return {};
+}
+
+// static
+void BitCopy::CopyMemoryToBuffer(uint8_t* dst,
```
dj2: This is only used by verifyer, why not put it in the anonymous namespace of verifier.cc?

src/bit_copy.cc
```diff
@@ -0,0 +1,230 @@
...
+    ptr[i] = src[i];
+  }
+
+  data >>= src_bit_offset;
+
+  CopyBits(dst, data, 0, bits);
+}
+
+// static
+float BitCopy::HexFloatToFloat(const uint8_t* value, uint8_t bits) {
```
dj2: ditto

src/bit_copy.cc
```diff
@@ -0,0 +1,230 @@
...
+
+    bit_offset -= bit_offset;
+    bits -= target_bits;
+    ++dst;
+    ++ptr;
+  }
+}
+
+// static
+uint16_t BitCopy::FloatExponent(const uint32_t hex_float) {
```
dj2: This is only used internally, move to anonymous namespace (same with other float methods below)

src/bit_copy.h
```diff
@@ -12,33 +12,54 @@
...
 class BitCopy {
  public:
-  // Copy [0, bits) bits of src to [bit_offset, bit_offset + bits) of dst.
-  static void CopyValueToBuffer(uint8_t* dst,
-                                const Value& src,
-                                uint8_t bit_offset,
-                                uint8_t bits);
+  // Copy [0, bits) bits of |src| to
+  // [dst_bit_offset, dst_bit_offset + bits) of |dst|.
+  static Result CopyValueToBuffer(uint8_t* dst,
```
dj2: This is only used by vulkan_buffer, move to anoymous namespace

src/verifier.cc
```diff
@@ -177,18 +188,204 @@ void SetupToleranceForTexels(const ProbeCommand* command,
...
+std::vector<Value> GetActualValuesFromTexel(const uint8_t* texel,
+                                            const Format* texel_format) {
+  assert(texel_format && !texel_format->GetComponents().empty());
+
+  std::vector<Value> actual_values(texel_format->GetComponents().size());
+  uint8_t bit_offset = 0;
+
+  for (size_t i = 0; i < texel_format->GetComponents().size(); ++i) {
+    const auto& component = texel_format->GetComponents()[i];
+    uint8_t actual[8] = {};
```
dj2: This needs to be {0, 0, 0, 0, 0, 0, 0, 0} to make the cts happy

src/verifier.cc
```diff
@@ -177,18 +188,204 @@ void SetupToleranceForTexels(const ProbeCommand* command,
...
+        float* ptr = reinterpret_cast<float*>(actual);
+        actual_values[i].SetDoubleValue(static_cast<double>(*ptr));
+      } else if (component.num_bits == 64) {
+        double* ptr = reinterpret_cast<double*>(actual);
+        actual_values[i].SetDoubleValue(*ptr);
+      } else {
+        assert(false && "Bits of component is not for double nor float type");
+      }
+    } else {
+      uint16_t* ptr16 = nullptr;
```
dj2: Move the declaration down to the usages.

src/verifier.cc
```diff
@@ -177,18 +188,204 @@ void SetupToleranceForTexels(const ProbeCommand* command,
...
+      uint32_t* ptr32 = nullptr;
+      uint64_t* ptr64 = nullptr;
+      switch (component.num_bits) {
+        case 8:
+          actual_values[i].SetDoubleValue(static_cast<double>(*actual));
```
dj2: Why are these all set as doubles and not integers?

src/verifier.cc
```diff
@@ -177,18 +188,204 @@ void SetupToleranceForTexels(const ProbeCommand* command,
...
+  return actual_values;
+}
+
+// If component mode of |texel_format| is FormatMode::kUNorm or
+// ::kSNorm or ::kSRGB, scale the corresponding value in |texel|.
+// Note that we do not scale values with FormatMode::kUInt, ::kSInt,
+// ::kUFloat, ::kSFloat.
+std::vector<Value> ScaleTexelValuesIfNeeded(const std::vector<Value>& texel,
+                                            const Format* texel_format) {
+  std::vector<Value> scaled_texel(texel.size());
```
dj2: Why not update in place, why create a new vector?

src/verifier.cc
```diff
@@ -177,18 +188,204 @@ void SetupToleranceForTexels(const ProbeCommand* command,
...
 Verifier::Verifier() = default;
 
 Verifier::~Verifier() = default;
 
+bool Verifier::IsTexelEqualToExpected(const std::vector<Value>& texel,
+                                      const Format* texel_format,
+                                      const ProbeCommand* command,
+                                      const double* tolerance,
+                                      const bool* is_tolerance_percent) {
+  TexelErrorInfo texel_error_info = {};
```
dj2: TexelErrorInfo texel_error_info = TexelErrorInfo();  to make CTS happy

src/verifier.cc
```diff
@@ -268,22 +453,30 @@ Result Verifier::Probe(const ProbeCommand* command,
...
+        (command->IsRGBA() ? ", " +
+                                 std::to_string(command->GetA() *
+                                                scale_factor_for_error_report) +
                                  "\n  Actual RGBA: "
                            : "\n  Actual RGB: ") +
-        std::to_string(static_cast<int>(p[0])) + ", " +
-        std::to_string(static_cast<int>(p[1])) + ", " +
-        std::to_string(static_cast<int>(p[2])) +
-        (command->IsRGBA() ? ", " + std::to_string(static_cast<int>(p[3]))
+        std::to_string(actual_color_texel_.r) + ", " +
```
dj2: Why do we need actual_color_texel_ can't we just copy the failing vector of values in the loop above? No need for the class value.

src/verifier.cc
```diff
@@ -268,22 +453,30 @@ Result Verifier::Probe(const ProbeCommand* command,
...
 
   if (count_of_invalid_pixels) {
-    const uint8_t* p = ptr + row_stride * (first_invalid_j + y) +
-                       texel_stride * (x + first_invalid_i);
+    const auto& component = texel_format->GetComponents().back();
+    float scale_factor_for_error_report = 1.0f;
+    if (component.mode == FormatMode::kUNorm ||
+        component.mode == FormatMode::kSNorm ||
+        component.mode == FormatMode::kSRGB) {
+      scale_factor_for_error_report = 255.0f;
```
dj2: We have the scale factor  for error report here and above why do we need it twice?

src/vulkan/format_data.cc
```diff
@@ -285,15 +285,157 @@ VkFormat ToVkFormat(FormatType type) {
...
+    case VK_FORMAT_R5G5B5A1_UNORM_PACK16:
+    case VK_FORMAT_R5G6B5_UNORM_PACK16:
+    case VK_FORMAT_R8G8_SINT:
+    case VK_FORMAT_R8G8_SNORM:
+    case VK_FORMAT_R8G8_SRGB:
+    case VK_FORMAT_R8G8_SSCALED:
+    case VK_FORMAT_R8G8_UINT:
+    case VK_FORMAT_R8G8_UNORM:
+    case VK_FORMAT_R8G8_USCALED:
+      return 2U;
```
dj2: Can we sort these in increasing order of size, so the 2U's then the 4Us, etc.

src/verifier.cc
```diff
@@ -177,18 +188,204 @@ void SetupToleranceForTexels(const ProbeCommand* command,
...
+
+  return actual_values;
+}
+
+// If component mode of |texel_format| is FormatMode::kUNorm or
+// ::kSNorm or ::kSRGB, scale the corresponding value in |texel|.
+// Note that we do not scale values with FormatMode::kUInt, ::kSInt,
+// ::kUFloat, ::kSFloat.
+std::vector<Value> ScaleTexelValuesIfNeeded(const std::vector<Value>& texel,
+                                            const Format* texel_format) {
```
dj2: Let's rename texel_format to framebuffer_format to make it clear what the format is.

src/verifier.cc
jaebaek: We had offline discussion and determined to keep it.

src/verifier.cc

src/verifier.cc

src/verifier.cc
jaebaek: Removed the above one.

src/verifier.cc

src/vulkan/format_data.cc
jaebaek: done

src/verifier_test.cc
```diff
@@ -250,7 +250,7 @@ TEST_F(VerifierTest, ProbeFrameBufferUInt32) {
                             4 * static_cast<uint32_t>(sizeof(uint32_t)), 1, 1,
                             static_cast<const void*>(&frame_buffer));
   EXPECT_TRUE(r.IsSuccess());
-  EXPECT_STREQ("", r.Error());
+  EXPECT_STREQ("", r.Error().c_str());
```
dj2: This shouldn't be needed. Use:

```
EXPECT_EQ("", r.Error());
```
and it should do string comparisons.

jaebaek: Done

PR #225, opened 2019-01-14
Add test case for accumulated ubo definition

jaebaek:


tests/cases/compute_accumulated_ubo_definition.amber
```diff
@@ -0,0 +1,59 @@
+# Copyright 2018 The Amber Authors.
```
dj2: 2019

jaebaek: done

PR #226, opened 2019-01-15
Vulkan: support index buffer

### LINKED ISSUE #204: Support index buffer for graphics pipeline

jaebaek:
Fixes #204

### DIFF
--- a/src/vulkan/graphics_pipeline.cc
+++ b/src/vulkan/graphics_pipeline.cc
@@ -599,6 +599,9 @@ Result GraphicsPipeline::Draw(const DrawArraysCommand* command,
     if (!r.IsSuccess())
       return r;
 
+    // VkRunner spec says
+    //   "vertexCount will be used as the index count, firstVertex
+    //    becomes the vertex offset and firstIndex will always be zero."
     vkCmdDrawIndexed(command_->GetCommandBuffer(),
                      command->GetVertexCount(),      /* indexCount */
                      instance_count,                 /* instanceCount */
--- a/src/vulkan/graphics_pipeline.cc
+++ b/src/vulkan/graphics_pipeline.cc
@@ -341,12 +341,6 @@ Result GraphicsPipeline::SetVertexBuffer(uint8_t location,
         "GraphicsPipeline::SetVertexBuffer: vertex buffer is nullptr");
   }
 
-  DeactivateRenderPassIfNeeded();
-
-  Result r = command_->BeginIfNotInRecording();
-  if (!r.IsSuccess())
-    return r;
-
   vertex_buffer->SetData(location, format, values);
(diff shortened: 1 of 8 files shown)


src/vulkan/graphics_pipeline.cc
```diff
@@ -571,8 +591,24 @@ Result GraphicsPipeline::Draw(const DrawArraysCommand* command,
...
+      return Result("Vulkan: Draw indexed is used without given indices");
+
+    r = index_buffer_->BindToCommandBuffer(command_->GetCommandBuffer());
+    if (!r.IsSuccess())
+      return r;
+
+    vkCmdDrawIndexed(command_->GetCommandBuffer(),
+                     command->GetVertexCount(),      /* indexCount */
+                     instance_count,                 /* instanceCount */
+                     0,                              /* firstIndex */
```
dj2: Should this be GetFirstVertexIndex()?

jaebaek: No. I was also very confused because of this. VkRunner spec says
```
vertexCount will be used as the index count, firstVertex becomes the vertex offset and firstIndex will always be zero.
```

To explain this, I added a test case as tests/cases/draw_triangle_list_with_index_buffer_and_vertex_offset.amber.
I will add more comments to explain it clearly.

tests/cases/draw_triangle_list_with_index_buffer_and_vertex_offset.amber
```diff
@@ -0,0 +1,95 @@
...
+#         R8   G8     R8  G8  B8
+           0    0    127 127 127
+         127  127    127 127 127
+           0  127    127 127 127
+         127    0    127 127 127
+
+[test]
+clear
+
+#                                "0" means 0th vertex among vertices
```
jaebaek: This should say
```
"0" means the first vertex among vertices which is shown in line 46 i.e., -128 -128    255   0   0.
Note that it uses 6 indices from the first index.
```

tests/cases/draw_triangle_list_with_index_buffer_and_vertex_offset.amber
```diff
@@ -0,0 +1,95 @@
...
+draw arrays indexed TRIANGLE_LIST 0 6
+relative probe rect rgb (0.0, 0.0, 1.0, 1.0) (1.0, 0, 0)
+
+#                                "4" means 4th vertex among vertices
```
jaebaek: This should say
```
"4" means the fifth vertex among vertices which is shown in line 53 i.e., 0 -128      0 255   0.
Note that it uses 6 indices from the first index.
```

jaebaek:
Comments were added. PTAL.

Issue #35, opened 2018-11-14
Dump image data


dj2:
There is an un-implemented command line option to the amber script to dump image data, this should be implemented.

hevrard:
The framebuffer contents is already accessed by the verifier (to probe pixel values) via:
`engine->GetFrameBufferInfo(info); // info.cpu_memory is then a pointer to pixel values`

The verifier then assumes a R8G8B8A8 format, with TODO comments saying more formats should be supported:
https://github.com/google/amber/blob/4336157fe1e4293f4b7b5e5d201571c7b71eff08/src/verifier.cc#L243

So as a first step, we may add a method
`Engine::GetFrameBufferAsR8G8B8A8(std::vector<uint8_t> *pixels)`
(or with a better name) which let the engine take care of any format conversion needed, and fails if the framebuffer format cannot be translated to R8G8B8A8.

Then use this method both in `Verifier::Probe()` and as a basis to dump the image in a PPM file.

Does this sound good?

dj2:
The CL to add all image formats is https://github.com/google/amber/pull/224 being worked on by @jaebaek . I'd prefer if the engine returned Value objects, just to make it clear what the format is, why not do Engine::GetFrameBuffer(std:;vector<Value>* pixels) and for now, just return r8g8b8a1 then we can just update it to do the right thing when the other CL lands.

Then, on the output side, just as the Value what type of data it contains. That way, the writing code should be good to go as soon as the engine returns the other formats.

dj2:
Thinking about this a bit more, it might make sense to return this as a DataBuffer. That way you can have the vector of Values and the data buffer will describe the format the data is in.

dj2:
Fixed by #227

PR #227, opened 2019-01-16
Add proof of concept of image dump in PPM

hevrard:
Add Engine:GetFrameBuffer(FormatBuffer*) to extract framebuffer pixel
values. Use it to read pixels and dump them in PPM format.


hevrard:
This is a first try, with the following caveats:
- the PPM dump should probably be extracted in a function or module, any idea on where to put it?
- The dump image option was originally treated at the top-level samples/amber.cc , but it feels like it should be in src/amber.cc to be able to access the framebuffer values before the engine shutdown. Otherwise we would still need to indicate to src/amber.cc whether to save the framebuffer image or not, to pass it to samples/amber.cc only when image dump is required.
- There are no tests yet :)

Let me know what you think.

hevrard:
Addresses #35

src/amber.cc
```diff
@@ -86,6 +88,57 @@ amber::Result Amber::ExecuteWithShaderData(const amber::Recipe* recipe,
...
+      const uint32_t pixel = data.AsUint32();
+      const char *p = reinterpret_cast<const char *>(&pixel);
+      // Write only RGB channels, PPM does not support alpha
+      image_stream.write(p, 3);
+    }
+
+    image_stream.close();
+  }
+
   return engine->Shutdown();
```
dj2: If we move the engine->Shutdown() to the destructor, then we can add a new call into amber to get the framebuffer instead of having to do it here, yea? We're going to want the ability to write out data buffers as well as framebuffers in the future so would be preferrable to have that code live elsewhere.

src/amber.cc
```diff
@@ -86,6 +88,57 @@ amber::Result Amber::ExecuteWithShaderData(const amber::Recipe* recipe,
...
+      return Result("Only R8G8B8A8 is supported for now");
+    }
+
+    std::ofstream image_stream;
+    image_stream.open(opts.image_file, std::ios::out | std::ios::binary);
+    if (!image_stream.is_open()) {
+      return Result("Cannot open file to write image");
+    }
+
+    // PPM header
```
dj2: If this gets pulled out this would then live in samples/ppm.h/cc or something like that

src/dawn/engine_dawn.h
```diff
@@ -68,6 +68,7 @@ class EngineDawn : public Engine {
   Result DoBuffer(const BufferCommand* cmd) override;
   Result DoProcessCommands() override;
   Result GetFrameBufferInfo(ResourceInfo* info) override;
+  Result GetFrameBuffer(FromatBuffer* pixels) override;
```
jaebaek: `FromatBuffer` --> `FormatBuffer`

src/engine.h
```diff
@@ -156,6 +157,11 @@ class Engine {
   /// commands have occurred e.g., DoClear, DoDrawArrays, DoDrawRect.
   virtual Result GetFrameBufferInfo(ResourceInfo* info) = 0;
 
+  /// Extract the pixel values from the framebuffer into the |pixels|
+  /// DataBuffer storing pixels value as a row-major array of RGBA
+  /// channels. The DatumType indicates the format of each channel.
+  virtual Result GetFrameBuffer(FormatBuffer* pixels) = 0;
```
jaebaek: `class EngineStub : public Engine` in src/executor_test.cc needs to override this method.
Please add `Result GetFrameBuffer(FormatBuffer*) override { return {}; }` to `class EngineStub : public Engine`.

src/dawn/engine_dawn.h
jaebaek: In addition, `Result EngineDawn::GetFrameBuffer(FromatBuffer* pixels)` is declared but does not have definition which results in linking error.

Please put
```
Result EngineDawn::GetFrameBuffer(FormatBuffer*) override {
  return Result("Dawn: EngineDawn::GetFrameBuffer not implemented");
}
```
in .cc file.

src/vulkan/engine_vulkan.cc
```diff
@@ -423,6 +423,47 @@ Result EngineVulkan::GetFrameBufferInfo(ResourceInfo* info) {
...
+Result EngineVulkan::GetFrameBuffer(FormatBuffer* pixels) {
+  assert(pixels);
+
+  ResourceInfo info;
+  GetFrameBufferInfo(&info);
+  if (info.type != ResourceInfoType::kImage) {
+    return Result("Vulkan:GetFrameBuffer() is invalid for non-image framebuffer");
+  }
+
+  // TODO(jaebaek) support other formats
```
jaebaek: // TODO(jaebaek): Support other formats

src/amber.cc
jaebaek: In that case, adding a unittest for this would be nice.
We can use a hard-coded image and run this code for the image to check the output PPM is correct.

hevrard:
Many thanks for your comments! I'll address all this as soon as feasible.

src/amber.cc
hevrard: > If we move the engine->Shutdown() to the destructor, then we can add a new call into amber to get the framebuffer instead of having to do it here, yea?

Just to clarify: you mean the destructor of the Amber object? Then engine must become a field of Amber, and be started/checked to be the right Engine whenever a flavor of Amber::Execute() is called?

Making the Engine survive an Execute call means that all Execute calls have to check that the maybe already started engine is of the type passed in the opts.engine argument of Execute, am I right?

src/amber.cc
hevrard: See my new push for this. An issue is: how do you properly define tests for modules under `samples/`? Also, by hard-coded image if you meant a file, where should this file be / how can the test know the path to this file? Otherwise, store the hard-coded image in the test source as an array of bytes?

I tried my best, happy to hear about your feedback on this.

src/vulkan/engine_vulkan.cc
hevrard: Done

hevrard:
Here goes an update, rebased on recent master (thanks @jaebaek for fixing Dawn issues):
- PPM encoding is now in samples/ppm.[h,cc]
- It comes with ppm_test.cc, but I am not sure how to setup cmake correctly for this test to be able to access the file test_WritePPM_reference.ppm
- See my comments on github about moving the image dump higher: should the engine survive Execute() calls?

jaebaek:
In my opinion, letting `WritePPM` return `std::string` rather directly writing output to file would be better for unittest.
I slightly updated the code [here](https://github.com/jaebaek/amber/tree/image_dump).
Could you please check it and adopt parts of it if it is reasonable change?

We have layering rule that code in `samples/` can use `include/` but cannot use internal objects of `src/`.
In addition, code in `src/` cannot use objects (e.g., variables, functions, classes) of `samples/`.

Since `src/amber.cc` uses `WritePPM()`, I moved `ppm*` files from `samples/` to `src/`.

src/amber.cc
dj2: Yup, I agree it's crappy, heh. I like your idea of using options, though I think we need to extend it a bit as I may want to extract multiple buffers (data buffers or data and frame, etc).

Maybe something like:
```
struct BufferInfo {
   std::string buffer_name;
   std::vector<Value> values;
};
struct Options {
  EngineType engine;
  EngineConfig* config;

  vector<BufferInfo> extractions;
};
```

Then on the client side, you'd push on the bufferInfo objects you want to extract and execute would iterate over the list and write the contents of that buffer into the values list?

Does that seem like it would work?

hevrard: Yes that looks good! I'll give a go at this approach as soon as feasible. Thanks!

hevrard:
New iteration:
- I moved back ppm* from src/ to samples/ : back in this situation, it is not clear where a test for PPM encoding can be added, since there is no test target for files under samples/
- The buffer to extract is now passed as an Amber option. Since this option needs to be modified by the Execute call (to fill in the buffer dimension and data), option becomes a pointer rather than a const reference.
- value.h is promoted to include/amber/ in order to be visible by ppm* files.

Let me know what you think!

dj2:
Looks like the build errors are from the ppm file:

```
/tmpfs/src/github/amber/samples/ppm.cc:27:10: error: use of old-style cast [-Werror,-Wold-style-cast]
  return (uint8_t)(word);
         ^        ~~~~~~
/tmpfs/src/github/amber/samples/ppm.cc:31:10: error: use of old-style cast [-Werror,-Wold-style-cast]
  return (uint8_t)(word >> 8);
         ^        ~~~~~~~~~~~
/tmpfs/src/github/amber/samples/ppm.cc:35:10: error: use of old-style cast [-Werror,-Wold-style-cast]
  return (uint8_t)(word >> 16);
         ^        ~~~~~~~~~~~~
/tmpfs/src/github/amber/samples/ppm.cc:56:21: error: implicit conversion changes signedness: 'uint8_t' (aka 'unsigned char') to 'char' [-Werror,-Wsign-conversion]
    image.push_back(byte0(pixel));  // R
    ~~~~~           ^~~~~~~~~~~~
/tmpfs/src/github/amber/samples/ppm.cc:57:21: error: implicit conversion changes signedness: 'uint8_t' (aka 'unsigned char') to 'char' [-Werror,-Wsign-conversion]
    image.push_back(byte1(pixel));  // G
    ~~~~~           ^~~~~~~~~~~~
/tmpfs/src/github/amber/samples/ppm.cc:58:21: error: implicit conversion changes signedness: 'uint8_t' (aka 'unsigned char') to 'char' [-Werror,-Wsign-conversion]
    image.push_back(byte2(pixel));  // B
    ~~~~~           ^~~~~~~~~~~~
6 errors generated.
```

PR #228, opened 2019-01-16
Vulkan: fix sync bug related to Buffer

### LINKED ISSUE #208: Copy to host or device in Buffer class has several synchronization bugs
Problems:

1. Missing `VkPhysicalDeviceLimits::nonCoherentAtomSize` check for flush / invalidate
2. When `Buffer::is_buffer_host_coherent_ == false` and `Buffer::is_buffer_host_accessible_ == true`,
writes from device --> `Buffer::CopyToHost()` --> `vkQueueSubmit()` --> reads from host
will have unexpected memory contents because `Buffer::CopyToHost()` calls `vkFlushMappedMemoryRanges()` before `vkQueueSubmit()` i.e., writes from device are not submitted.
3. Even when `Buffer::is_buffer_host_coherent_ == true` and `Buffer::is_buffer_host_accessible_ == true`, we need memory barrier to make the writes from host / device **AVAILABLE** to the device / host domain. The spec says
\
"If the memory object does have the VK_MEMORY_PROPERTY_HOST_COHERENT_BIT property flag,
writes to the memory object from the host are automatically made available to the host domain. Similarly,
writes made **available** to the host domain are automatically made visible to the host."
\
What it means is that we must first make writes from the device **AVAILABLE** to the host domain.

Solutions:
1. Simplify the memory type. Allocate only memory type whose property is `VK_MEMORY_PROPERTY_HOST_COHERENT_BIT | VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT`.
2. Use memory barrier for both `Buffer::CopyToHost()` and `Buffer::CopyToDevice()`

jaebaek:
Based on this CL, Vulkan engine always chooses host accessible
and coherent memory for buffer and uses memory barrier for
copy-to-device and copy-to-host.

Fixes #208

### DIFF
--- a/src/vulkan/buffer.cc
+++ b/src/vulkan/buffer.cc
@@ -29,20 +29,22 @@ Result Buffer::Initialize(const VkBufferUsageFlags usage) {
   if (!r.IsSuccess())
     return r;
 
-  AllocateResult allocate_result = AllocateAndBindMemoryToVkBuffer(
-      buffer_, &memory_, VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT, false);
+  AllocateResult allocate_result =
+      AllocateAndBindMemoryToVkBuffer(buffer_, &memory_,
+                                      VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT |
+                                          VK_MEMORY_PROPERTY_HOST_COHERENT_BIT,
+                                      true);
   if (!allocate_result.r.IsSuccess())
     return allocate_result.r;
 
-  if (IsMemoryHostAccessible(allocate_result.memory_type_index)) {
-    is_buffer_host_accessible_ = true;
-    is_buffer_host_coherent_ =
-        IsMemoryHostCoherent(allocate_result.memory_type_index);
-    return MapMemory(memory_);
+  if (!IsMemoryHostAccessible(allocate_result.memory_type_index) ||
+      !IsMemoryHostCoherent(allocate_result.memory_type_index)) {
+    return Result(
+        "Vulkan: Buffer::Initialize() Buffer is not host accessible or not "
+        "host coherent.");
   }
 
-  is_buffer_host_accessible_ = false;
-  return Resource::Initialize();
+  return MapMemory(memory_);
 }
 
 Result Buffer::CreateVkBufferView(VkFormat format) {
@@ -61,29 +63,11 @@ Result Buffer::CreateVkBufferView(VkFormat format) {
 }
 
 Result Buffer::CopyToDevice(VkCommandBuffer command) {
-  if (is_buffer_host_accessible_)
-    return FlushMemoryIfNeeded();
-
-  VkBufferCopy region = {};
-  region.srcOffset = 0;
-  region.dstOffset = 0;
-  region.size = GetSizeInBytes();
-
-  vkCmdCopyBuffer(command, GetHostAccessibleBuffer(), buffer_, 1, &region);
   MemoryBarrier(command);
   return {};
 }
 
 Result Buffer::CopyToHost(VkCommandBuffer command) {
-  if (is_buffer_host_accessible_)
-    return InvalidateMemoryIfNeeded();
-
-  VkBufferCopy region = {};
-  region.srcOffset = 0;
-  region.dstOffset = 0;
-  region.size = GetSizeInBytes();
-
-  vkCmdCopyBuffer(command, buffer_, GetHostAccessibleBuffer(), 1, &region);
   MemoryBarrier(command);
   return {};
 }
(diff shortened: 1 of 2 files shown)


paulthomson:
For the copy-to-device case, I believe a memory barrier should not be necessary. For host writes, followed by vkFlushMappedMemoryRanges (not needed for host coherent memory), followed by a vkQueueSubmit: the host writes should be visible to any device reads via commands in the submitted command buffer, without any barrier. This should be guaranteed. This is due to `6.9. Host Write Ordering Guarantees` in the Vulkan spec.

It is good to be cautious and having a barrier does not hurt, but perhaps it is also good to make use of this fundamental guarantee in case others try to learn from this code? I am unsure what is best! I could also be wrong, but I think the spec is quite clear and concise here.

src/vulkan/buffer.cc
```diff
@@ -61,29 +63,11 @@ Result Buffer::CreateVkBufferView(VkFormat format) {
 }
 
 Result Buffer::CopyToDevice(VkCommandBuffer command) {
-  if (is_buffer_host_accessible_)
-    return FlushMemoryIfNeeded();
```
dj2: Didn't this early exit before?

src/vulkan/buffer.cc
```diff
@@ -61,29 +63,11 @@ Result Buffer::CreateVkBufferView(VkFormat format) {
...
-  region.size = GetSizeInBytes();
-
-  vkCmdCopyBuffer(command, GetHostAccessibleBuffer(), buffer_, 1, &region);
   MemoryBarrier(command);
   return {};
 }
 
 Result Buffer::CopyToHost(VkCommandBuffer command) {
-  if (is_buffer_host_accessible_)
-    return InvalidateMemoryIfNeeded();
```
dj2: Didn't this early exit before?

src/vulkan/buffer.h
```diff
@@ -22,42 +22,38 @@
...
+  // memory |memory_|, this method only conducts memory barrier to
+  // make it available to device domain.
   Result CopyToDevice(VkCommandBuffer command);
 
   // Resource
-  VkDeviceMemory GetHostAccessMemory() const override {
-    if (is_buffer_host_accessible_)
-      return memory_;
-
-    return Resource::GetHostAccessMemory();
```
dj2: Is Resource::GetHostAccessMemory still used or can it be removed?

src/vulkan/buffer.cc
```diff
@@ -103,47 +87,12 @@ void Buffer::Shutdown() {
...
-
+    UnMapMemory(memory_);
     vkFreeMemory(GetDevice(), memory_, nullptr);
   }
 
   if (buffer_ != VK_NULL_HANDLE)
     vkDestroyBuffer(GetDevice(), buffer_, nullptr);
-
-  if (!is_buffer_host_accessible_)
-    Resource::Shutdown();
```
dj2: Do we call Resource::Shutdown elsewwhere?

dneto0:
> It is good to be cautious and having a barrier does not hurt, but perhaps it is also good to make use of this fundamental guarantee in case others try to learn from this code? I am unsure what is best! I could also be wrong, but I think the spec is quite clear and concise here.

We talked about this.  Yes, you're right.  It's not necessary.  We'll keep it because we want Amber to be as dumb as possible w.r.t. synchronization.  We don't care about speed.  But we will add a comment to say something like "This is redundant because of ....  But we prefer to keep it to simplify our own code."

jaebaek:
> For the copy-to-device case, I believe a memory barrier should not be necessary. For host writes, followed by vkFlushMappedMemoryRanges (not needed for host coherent memory), followed by a vkQueueSubmit: the host writes should be visible to any device reads via commands in the submitted command buffer, without any barrier. This should be guaranteed. This is due to `6.9. Host Write Ordering Guarantees` in the Vulkan spec.
> 
> It is good to be cautious and having a barrier does not hurt, but perhaps it is also good to make use of this fundamental guarantee in case others try to learn from this code? I am unsure what is best! I could also be wrong, but I think the spec is quite clear and concise here.

I agree that memory barrier is not actually needed for this case. Thank you for letting me know it. We discuss this issue offline and concluded just adding a comment `This is redundant, but we prefer to leave it in` because we do not focus on performance now and keeping it does not have any problems other than performance.

src/vulkan/buffer.cc
jaebaek: We exist early here with this previous code. This is only the case where the buffer is **host-visible but not host-coherent**.
Now we allocate all buffer with only memory that is **host-visible and host-coherent**.
Now we can simply call memory barrier without exceptions.

src/vulkan/buffer.cc
jaebaek: Ditto.

src/vulkan/buffer.cc
jaebaek: Now `Resource::Shutdown()` does not do anything because now, with this CL, both `host_accessible_memory_` and `host_accessible_buffer_` are `VK_NULL_HANDLE`.

However, to prevent potential bugs later e.g., we update `Resource::Shutdown()` in the future, I put it here.

src/vulkan/buffer.h
jaebaek: `Resource::GetHostAccessMemory()` is still used for image resource. It should not be removed.

PR #232, opened 2019-01-17
Change initializers to make GCC happy.

dj2:
Using the = {} initializer syntax causes missing-field-initializers
warnings for some versions of GCC. This CL converts all uses of = {} to
construct the actual object desired instead.

### DIFF
--- a/src/vulkan/device.h
+++ b/src/vulkan/device.h
@@ -83,9 +83,9 @@ class Device {
   VkInstance instance_ = VK_NULL_HANDLE;
   VkDebugReportCallbackEXT callback_ = VK_NULL_HANDLE;
   VkPhysicalDevice physical_device_ = VK_NULL_HANDLE;
-  VkPhysicalDeviceProperties physical_device_properties_ = {};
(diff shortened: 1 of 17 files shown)


src/vulkan/device.h
```diff
@@ -83,9 +83,9 @@ class Device {
   VkInstance instance_ = VK_NULL_HANDLE;
   VkDebugReportCallbackEXT callback_ = VK_NULL_HANDLE;
   VkPhysicalDevice physical_device_ = VK_NULL_HANDLE;
-  VkPhysicalDeviceProperties physical_device_properties_ = {};
-  VkPhysicalDeviceMemoryProperties physical_memory_properties_ = {};
-  VkPhysicalDeviceFeatures available_physical_device_features_ = {};
+  VkPhysicalDeviceProperties physical_device_properties_;
+  VkPhysicalDeviceMemoryProperties physical_memory_properties_;
+  VkPhysicalDeviceFeatures available_physical_device_features_;
```
jaebaek: I know this works fine but do you think not initializing these is safe?

dj2: I think these are fine, if anything comes up we can fix it.

PR #235, opened 2019-01-18
Dawn: fix compile errors on newer Dawn

jaebaek:


jaebaek:
I did not test this but compile is fine.
Please doublecheck it if you have any way to actually run it.

PR #239, opened 2019-01-21
Fixup vulkan include

dj2:


dj2:
TBR-ing so I can re-run CTS integration test.

PR #240, opened 2019-01-21
VkScript: support probe without rect

### LINKED ISSUE #238: Support probe point for color framebuffer
VkRunner supports not only probing rectangle but also probing a single point.
Currently, if we probe a single point using Amber e.g., `relative probe rgba (0.2, 0.7) (1.0, 0.4, 0.5, 0.2)`, it recognizes it as probing rectangle not probing point.

jaebaek:
Probe command for a color frame buffer without `rect` attribute
means "probe a single pixel". This commit supports it.

Fixes #238


src/verifier_test.cc
```diff
@@ -142,6 +145,58 @@ TEST_F(VerifierTest, ProbeFrameBuffer) {
...
+  probe.SetX(7.0f);
+  probe.SetY(3.0f);
+  probe.SetR(0.2f);
+  probe.SetG(0.8f);
+  probe.SetB(0.25f);
+  probe.SetA(0.5f);
+
+  r = verifier.Probe(&probe, 4, 40, 10, 10,
+                     static_cast<const void*>(frame_buffer));
+  EXPECT_TRUE(r.IsSuccess());
```
dj2: Does this second Probe call verify anything different from the one above?

src/verifier_test.cc
```diff
@@ -142,6 +145,58 @@ TEST_F(VerifierTest, ProbeFrameBuffer) {
...
+
+  probe.SetX(0.0f);
+  probe.SetY(0.0f);
+
+  r = verifier.Probe(&probe, 4, 40, 10, 10,
```
dj2: This can either be removed, or should be moved to it's own test case. What does it tell us that the first Probe doesn't?

tests/cases/clear_color_with_probe_point.amber
```diff
@@ -0,0 +1,40 @@
+# Copyright 2018 The Amber Authors.
```
dj2: 2019

tests/cases/draw_triangle_list_with_probe_point.amber
```diff
@@ -0,0 +1,121 @@
+# Copyright 2018 The Amber Authors.
```
dj2: 2019

src/command.h
```diff
@@ -220,9 +220,15 @@ class ProbeCommand : public Probe {
   ProbeCommand();
   ~ProbeCommand() override;
 
-  void SetWholeWindow() { is_whole_window_ = true; }
+  void SetWholeWindow() {
+    is_whole_window_ = true;
+    is_probe_rect_ = true;
```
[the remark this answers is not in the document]
jaebaek: done

src/verifier_test.cc
jaebaek: Removed

src/verifier_test.cc
jaebaek: removed

tests/cases/clear_color_with_probe_point.amber
jaebaek: done

tests/cases/draw_triangle_list_with_probe_point.amber
jaebaek: done

PR #245, opened 2019-01-24
 [vulkan] Make vkGetInstanceProcAddr a required config param.

dj2:
This CL changes the VulkanConfig to require the vkGetInstanceProcAddr
pointer. This will be used internally to lookup all the used vulkan
methods.


include/amber/amber_vulkan.h
```diff
@@ -43,6 +43,9 @@ struct VulkanEngineConfig : public EngineConfig {
 
   /// The VkQueue to use.
   VkQueue queue;
+
+  /// REQUIRED. The Vulkan instance procedue loader.
```
dneto0: "procedure"

samples/config_helper_vulkan.cc
```diff
@@ -18,6 +18,7 @@
 #include <cassert>
 #include <cstring>
 #include <set>
+#include <vulkan/vulkan.h>
```
dneto0: Hmmm. I normally use " quotes for things not shipped with the compiler.   But others have argued against me.
Coder's choice here.

dneto0:
Recommend fixing the typo, and making sure formatting is acceptable (however headers are arranged).
But otherwise LGTM.

include/amber/amber_vulkan.h
dj2: done. I'd actually fixed it in the next CL in the chain already, heh.

samples/config_helper_vulkan.cc
dj2: I usually go with ""'s for things in the repo (so in include/ src/ or third_party/) and <> otherwise

Issue #206, opened 2018-12-28
Support depth stencil buffer for frame buffer


jaebaek:
See [depth stencil buffer in VkRunner](https://github.com/Igalia/vkrunner/blob/master/examples/depth-buffer.shader_test).

dj2:
Fixed by #248

PR #248, opened 2019-01-28
Vulkan: support depth frame buffer

jaebaek:
Related to #206


tests/cases/draw_triangle_list_with_depth.amber
```diff
@@ -0,0 +1,91 @@
+# Copyright 2018 The Amber Authors.
```
dj2: 2019

tests/cases/draw_triangle_list_with_depth.amber
```diff
@@ -0,0 +1,91 @@
...
+}
+
+[fragment shader]
+#version 430
+
+layout(location = 0) in vec4 frag_color;
+layout(location = 0) out vec4 final_color;
+
+void main() {
+  final_color = frag_color;
```
dj2: This doesn't appear to use the depth buffer? How do you know if it's working?

dj2: Can you add a comment in here that that's what's happening?

jaebaek: done

PR #251, opened 2019-01-29
Fix wrong RGBA report from Verifier

jaebaek:
When Verifier reports probe error, it report wrong RGBA. This CL
reorders actual texel values in RGBA order to fix the bug.

### DIFF
--- a/src/verifier.cc
+++ b/src/verifier.cc
@@ -480,30 +480,29 @@ bool IsTexelEqualToExpected(const std::vector<double>& texel,
   return true;
 }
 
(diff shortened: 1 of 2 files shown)


src/verifier.cc
```diff
@@ -480,6 +480,32 @@ bool IsTexelEqualToExpected(const std::vector<double>& texel,
   return true;
 }
 
+std::vector<double> GetSortedTexelInOrderOfRGBA(
```
dj2: GetTexelInRGBA, sorted doesn't really mean anything in this context I don't think.

Issue #229, opened 2019-01-16
Support pipeline setup based on PipelineData


jaebaek:
Depth/stencil and ColorBlend must be set based on `PipelineData`

PR #253, opened 2019-01-29
Vulkan: support setup based on PipelineData

### LINKED ISSUE #64: ComputeCommand does not need PipelineData
As far as I understand, PipelineData will be used for only graphics pipeline because it only has setups for graphics pipeline e.g., topology, polygon mode, cull mode.

I think we can remove it from ComputeCommand.

jaebaek:
Compute pipeline does not need PipelineData. This commit removes
PipelineData from ComputeCommand and supports graphics pipeline
setup based on PipelineData. Since PipelineData contains information
for stencil test, it also supports stencil test.

Fixes #64, #229
Related to #206

### DIFF
--- a/src/command.cc
+++ b/src/command.cc
@@ -68,27 +68,18 @@ ProbeSSBOCommand* Command::AsProbeSSBO() {
   return static_cast<ProbeSSBOCommand*>(this);
 }
 
-DrawRectCommand::DrawRectCommand(
-    PipelineData data,
-    bool is_pipeline_data_different_from_previous_one)
-    : Command(Type::kDrawRect),
-      data_(data),
-      is_pipeline_data_different_from_previous_one_(
-          is_pipeline_data_different_from_previous_one) {}
+DrawRectCommand::DrawRectCommand(PipelineData data)
+    : Command(Type::kDrawRect), data_(data) {}
 
 DrawRectCommand::~DrawRectCommand() = default;
 
-DrawArraysCommand::DrawArraysCommand(
-    PipelineData data,
-    bool is_pipeline_data_different_from_previous_one)
-    : Command(Type::kDrawArrays),
-      data_(data),
-      is_pipeline_data_different_from_previous_one_(
-          is_pipeline_data_different_from_previous_one) {}
+DrawArraysCommand::DrawArraysCommand(PipelineData data)
+    : Command(Type::kDrawArrays), data_(data) {}
 
 DrawArraysCommand::~DrawArraysCommand() = default;
 
-ComputeCommand::ComputeCommand() : Command(Type::kCompute) {}
+ComputeCommand::ComputeCommand(PipelineData data)
+    : Command(Type::kCompute), data_(data) {}
 
 ComputeCommand::~ComputeCommand() = default;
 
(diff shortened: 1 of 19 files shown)


jaebaek:
I will add two more test cases for stencil tests, but C++ code is ready to review. PTAL.

src/command.cc
```diff
@@ -68,18 +68,27 @@ ProbeSSBOCommand* Command::AsProbeSSBO() {
   return static_cast<ProbeSSBOCommand*>(this);
 }
 
-DrawRectCommand::DrawRectCommand(PipelineData data)
-    : Command(Type::kDrawRect), data_(data) {}
+DrawRectCommand::DrawRectCommand(
+    PipelineData data,
+    bool is_pipeline_data_different_from_previous_one)
```
dj2: I don't think this bool should be part of the commands as it doesn't mean anything to the command. It should be up to the backend to figure this out automatically.

src/command.cc
```diff
@@ -68,18 +68,27 @@ ProbeSSBOCommand* Command::AsProbeSSBO() {
...
+    : Command(Type::kDrawArrays),
+      data_(data),
+      is_pipeline_data_different_from_previous_one_(
+          is_pipeline_data_different_from_previous_one) {}
 
 DrawArraysCommand::~DrawArraysCommand() = default;
 
-ComputeCommand::ComputeCommand(PipelineData data)
-    : Command(Type::kCompute), data_(data) {}
+ComputeCommand::ComputeCommand() : Command(Type::kCompute) {}
```
dj2: If this doesn't need pipeline data let's remove it in a separate CL which just does that.

src/vulkan/graphics_pipeline.cc
```diff
@@ -89,6 +89,282 @@ VkPrimitiveTopology ToVkTopology(Topology topology) {
...
+    case CompareOp::kGreaterOrEqual:
+      return VK_COMPARE_OP_GREATER_OR_EQUAL;
+    case CompareOp::kAlways:
+      return VK_COMPARE_OP_ALWAYS;
+  }
+  assert(false && "Vulkan::Unknown CompareOp");
+  return VK_COMPARE_OP_NEVER;
+}
+
+VkBool32 ToVkBool32(bool true_or_false) {
```
dj2: I think we can remove this one and just do it inline.

src/vulkan/graphics_pipeline.cc
```diff
@@ -582,8 +925,16 @@ Result GraphicsPipeline::Draw(const DrawArraysCommand* command,
   if (!r.IsSuccess())
     return r;
 
+  if (pipeline_ != VK_NULL_HANDLE &&
+      command->IsPipelineDataDifferentFromPreviousOne()) {
+    r = ResetPipeline();
```
dj2: Any reason to just not always do this?

jaebaek: I think creating new graphics pipeline for every draw commands would cost a lot, because graphics pipeline is involved with my setups.
If we do not have different PipelineData, we do not need to create new one.

Do you think it would be better to create new graphics pipeline for every draw commands regardless of performance at this moment?

PR #255, opened 2019-01-31
Fixup to make clang happy

dj2:


dj2:
TBRing to fixup building with clang.

PR #258, opened 2019-02-04
Fix compiler error for release build with GCC 7.3

### LINKED ISSUE #257: amber fails to compile in release mode.
When I build the release version of amber with gcc 7.3, I get a couple errors:

```
../samples/config_helper_vulkan.cc: In member function ‘void sample::ConfigHelperVulkan::CreateVulkanInstance()’:
../samples/config_helper_vulkan.cc:675:24: error: variable ‘instance_info’ set but not used [-Werror=unused-but-set-variable]
   VkInstanceCreateInfo instance_info = {};
                        ^~~~~~~~~~~~~
../samples/config_helper_vulkan.cc: In member function ‘void sample::ConfigHelperVulkan::CreateVulkanDevice(const VkPhysicalDeviceFeatures&, const std::vector<std::__cxx11::basic_string<char> >&)’:
../samples/config_helper_vulkan.cc:752:22: error: variable ‘info’ set but not used [-Werror=unused-but-set-variable]
   VkDeviceCreateInfo info = {};
                      ^~~~
../samples/config_helper_vulkan.cc: In member function ‘void sample::ConfigHelperVulkan::ChooseVulkanPhysicalDevice(const VkPhysicalDeviceFeatures&, const std::vector<std::__cxx11::basic_string<char> >&)’:
../samples/config_helper_vulkan.cc:714:26: error: ‘count’ is used uninitialized in this function [-Werror=uninitialized]
   physical_devices.resize(count);
   ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~

```

This seems to come from [here](https://github.com/google/amber/blob/master/samples/config_helper_vulkan.cc#L683) where real work is in an assert.  That is a problem because that work will not be done in release mode.

There same happens [here](https://github.com/google/amber/blob/master/samples/config_helper_vulkan.cc#L767), and [here](https://github.com/google/amber/blob/master/samples/config_helper_vulkan.cc#L712).

Those are the ones that I found because of compiler errors.  A check should be done for others as well.

jaebaek:
Calling important functions within assert() causes build / runtime
errors. We must use assert() only when reporting the error and it
must not contain critical logic.

Fixes #257


samples/config_helper_vulkan.cc
```diff
@@ -680,8 +680,10 @@ void ConfigHelperVulkan::CreateVulkanInstance() {
   instance_info.enabledExtensionCount = 1U;
   instance_info.ppEnabledExtensionNames = &kExtensionForValidationLayer;
 
-  assert(vkCreateInstance(&instance_info, nullptr, &vulkan_instance_) ==
-         VK_SUCCESS);
+  if (vkCreateInstance(&instance_info, nullptr, &vulkan_instance_) !=
+      VK_SUCCESS) {
+    assert(false && "Sample: vkCreateInstance fail");
```
dj2: This needs to return an error. In Release mode, the code will just continue as though everything is fine.

samples/config_helper_vulkan.cc
```diff
@@ -706,14 +708,18 @@ void ConfigHelperVulkan::CreateDebugReportCallback() {
...
     const std::vector<std::string>& required_extensions) {
-  uint32_t count;
+  uint32_t count = 0;
   std::vector<VkPhysicalDevice> physical_devices;
 
-  assert(vkEnumeratePhysicalDevices(vulkan_instance_, &count, nullptr) ==
-         VK_SUCCESS);
+  if (vkEnumeratePhysicalDevices(vulkan_instance_, &count, nullptr) !=
+      VK_SUCCESS) {
+    assert(false && "Sample: vkEnumeratePhysicalDevices fail");
```
dj2: Same with these ones.

dj2:
I think we have to do something like https://github.com/google/amber/pull/260

samples/config_helper_vulkan.cc
jaebaek: done

samples/config_helper_vulkan.cc
jaebaek: done

jaebaek:
Since #260 replaced `assert()` to `Result`, this commit needs only simple changes.
PTAL.

PR #260, opened 2019-02-05
Remove some asserts, use return Results

### LINKED ISSUE #257: amber fails to compile in release mode.

dj2:
Fixes #257  where assets were wrapping required calls.

### DIFF
--- a/samples/config_helper_vulkan.cc
+++ b/samples/config_helper_vulkan.cc
@@ -16,7 +16,6 @@
 
 #include <vulkan/vulkan.h>
 #include <algorithm>
-#include <cassert>
 #include <cstring>
 #include <set>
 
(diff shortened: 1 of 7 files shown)


samples/config_helper_vulkan.cc
```diff
@@ -764,26 +659,46 @@ void ConfigHelperVulkan::CreateVulkanDevice(
...
 
-  CreateVulkanInstance();
-  CreateDebugReportCallback();
-  ChooseVulkanPhysicalDevice(required_vulkan_features, required_extensions);
-  CreateVulkanDevice(required_vulkan_features, required_extensions);
   vkGetDeviceQueue(vulkan_device_, vulkan_queue_family_index_, 0,
                    &vulkan_queue_);
 
-  std::unique_ptr<amber::EngineConfig> cfg_holder =
+  *cfg_holder =
```
dneto0: I would have gone for cfg_holder->reset( the new'd pointer ).

But this is fine.

PR #266, opened 2019-02-06
[vulkan] Always require vulkan to be provided to library.

dj2:
This CL requires that all of the device and instance information for the vulkan
engine are provided. The Amber library does not create the device internally
anymore.


dj2:
Thoughts? The CTS does this already and we converted Samples app when we did the CTS integration. This removes a lot of code that is duplicated between vulkan engine and samples/

dj2:
The comment at the top of the structure says they are all required. Would you prefer if it labeled each field individually?

I'll do the readme update in a followup.

src/vulkan/engine_vulkan.cc
```diff
@@ -69,21 +69,17 @@ Result EngineVulkan::Initialize(EngineConfig* config,
...
-      return Result("Vulkan::Initialize queue handle is null.");
-
-    device_ = MakeUnique<Device>(
-        vk_config->instance, vk_config->physical_device,
-        vk_config->available_features, vk_config->available_extensions,
-        vk_config->queue_family_index, vk_config->device, vk_config->queue);
-  } else {
-    device_ = MakeUnique<Device>();
-  }
+  if (vk_config->device != VK_NULL_HANDLE)
```
jaebaek: This must be `vk_config->device == VK_NULL_HANDLE`.

dj2: good catch. done.

PR #268, opened 2019-02-07
Vkscript: fix matrix stride bug

### LINKED ISSUE #267: Wrong matrix stride bug causes errors when we use "mat*" types for ssbo and uniform commands
```
# Fill mat3x3 using mat3
# MatrixStride == 16
ssbo 0:0 subdata mat3 0 1.0 0.0 0.0 \
                        0.0 1.0 0.0 \
                        0.0 0.0 1.0
# MatrixStride == 16
uniform mat3 0 1.0 0.0 0.0 \
               0.0 1.0 0.0 \
               0.0 0.0 1.0
```

In the above example, matrix stride must be 16 bytes but Amber processes it as 12 bytes.

jaebaek:
Fixes #267


jaebaek:
Change row stride for matrices based on [Vulkan spec](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#interfaces-resources-layout)

tests/cases/compute_mat2x2.amber
```diff
@@ -0,0 +1,76 @@
+# Copyright 2018 The Amber Authors.
```
dj2: 2019

tests/cases/compute_mat2x3.amber
```diff
@@ -0,0 +1,76 @@
+# Copyright 2018 The Amber Authors.
```
dj2: 2019

tests/cases/compute_mat3x2.amber
```diff
@@ -0,0 +1,95 @@
+# Copyright 2018 The Amber Authors.
```
dj2: 2019

tests/cases/compute_mat3x3.amber
```diff
@@ -0,0 +1,82 @@
+# Copyright 2018 The Amber Authors.
```
dj2: 2019 (and others)

src/vkscript/command_parser.cc
```diff
@@ -470,13 +470,36 @@ Result CommandParser::ProcessClear() {
 
 Result CommandParser::ParseValues(const std::string& name,
                                   const DatumType& type,
-                                  std::vector<Value>* values) {
+                                  std::vector<Value>* values,
+                                  bool use_std430_layout = false) {
```
dj2: Remove the = false. I'm not a fan of default values, makes it harder to tell when a param isn't needed anymore.

src/vkscript/command_parser.cc
```diff
@@ -470,13 +470,36 @@ Result CommandParser::ProcessClear() {
...
   assert(values);
 
+  uint32_t row_index = 0;
   auto token = tokenizer_->NextToken();
+  size_t seen = 0;
   while (!token->IsEOL() && !token->IsEOS()) {
     Value v;
 
+    // When |type| is a multi-column matrix, fill more values for its
+    // memory layout.
```
dj2: Can you add why we fill more values here?

src/vkscript/command_parser.cc
```diff
@@ -470,13 +470,36 @@ Result CommandParser::ProcessClear() {
...
+      if (use_std430_layout || type.RowCount() == 3U) {
+        if ((type.IsFloat() || type.IsDouble())) {
+          v.SetDoubleValue(0);
+        } else {
+          v.SetIntValue(0);
+        }
+        // For std430 layout, the stride of each row must be
+        // 4 * element size in bytes when row count is not 1.
+        // For std140 layout, the stride of each row must be
+        // 4 * element size in bytes when row count is 3 or 4.
```
dj2: In std140, if row size is 2 can the stride be 16? Can we just always make the row stride 16 and not worry if it's std430 or 140?

src/vkscript/command_parser.cc
jaebaek: done

src/vkscript/command_parser.cc
jaebaek: No.
If it is std140 and the row size is 2, its stride will be 8.
Please check this condition `use_std430_layout || type.RowCount() == 3U`. If it is std140 i.e., `use_std430_layout == false`, it is true only when `type.RowCount() == 3U`. In other words, if it is std140, only matrix with 3 elements in a row matters.

tests/cases/compute_mat2x2.amber
jaebaek: done

tests/cases/compute_mat2x3.amber
jaebaek: done

tests/cases/compute_mat3x2.amber
jaebaek: done

tests/cases/compute_mat3x3.amber
jaebaek: done

src/vkscript/command_parser.cc
dj2: Right, but even in std140, can we make the row size 2 and the stride 16, or the row size 4 and then then stride is 16 anyway? Basically, can we force these to have the same stride even if we don't have too?

jaebaek: When glslang compiles a shader with a matrix whose row size is 2, it always sets the stride 8.
It sets the stride 16 only for matrix with 3 or 4 rows.

If a user correctly writes a vkrunner script, we expect `mat.*` types in SSBO command matches `mat.*` types in shaders.
Thus, in std140, we must set the stride 8 for the row size 2.
We must set the stride 16 for the row size 3 or 4.

PR #269, opened 2019-02-07
Vulkan: support buffer dump

### LINKED ISSUE #36: Dump buffer data
There is an un-implemented flag to the amber script to dump buffer data, this should be implemented.

jaebaek:
Fixes #36


samples/amber.cc
```diff
@@ -303,6 +305,31 @@ int main(int argc, const char** argv) {
       image_file << image;
       image_file.close();
     }
+
+    if (!options.buffer_filename.empty()) {
+      std::ofstream buffer_file;
+      buffer_file.open(options.buffer_filename, std::ios::out);
+      if (!buffer_file.is_open()) {
+        std::cerr << "Cannot open file for image dump: ";
```
dj2: buffer dump

src/amber.cc
```diff
@@ -96,6 +98,36 @@ amber::Result Amber::ExecuteWithShaderData(const amber::Recipe* recipe,
...
     }
+
+    if (!dump_buffer && buffer_info.buffer_name == "buffer")
+      dump_buffer = true;
+  }
+
+  if (!dump_buffer)
+    return engine->Shutdown();
+
+  opts->extractions.erase(
```
dj2: The rest of this method seems overly complicated. When requesting a buffer dump, we should ask for the buffer name to be dumped to be provided on the command line, and give the ability to request multiple buffers. We then just dump that specific buffer. That code can go in the loop above which dumps framebuffer.

src/engine.h
```diff
@@ -162,6 +170,11 @@ class Engine {
                                    const uint32_t binding,
                                    ResourceInfo* info) = 0;
 
+  /// Copy the contents of all resources bound to descriptors and get
+  /// their resource information e.g., size for buffer, width, height,
+  /// depth for image.
+  virtual std::vector<ResourceInfo> GetAllDescriptorInfo() = 0;
```
dj2: I think we want to get just a single buffer. If we have one small one to extract and a huge second one there is no reason to extract the huge buffer.

dj2:
We're not trying to match vkrunner in command lines, just in the input script format. So, if we're different it doesn't matter. 

We have the -B flag already which is used to specify the buffers to write, The description is probably wrong and it should just be the buffer name to write as in amberscript all buffers are named. Then, you can give a file name and a series of -B flags to write out those specific buffers. All buffers get written to the same file.

jaebaek:
I removed `GetAllDescriptorInfo()` and let it dump given buffer via `-B`.
PTAL.

src/amber.cc
```diff
@@ -25,6 +27,64 @@
...
 
 namespace amber {
+namespace {
+
+// Parse |buffer_id| and return desciptor set and binding using |desc_set|
+// and |binding|. |buffer_id| must be non-empty string. It must be a single
+// non-negative integer or two those integers separated by ':'. For example,
+// ":0", "1", and "2:3" are valid strings for |buffer_id|, but "", "-4",
+// ":-5", ":", "a", and "b:c" are invalid strings for |buffer_id|.
+amber::Result ParseDescriptorSetAndBinding(const std::string& buffer_id,
```
dj2: This is complicated enough it needs tests. Maybe move it out to a helper and do it there? You could also do something like:
```
Tokenizer t(buffer_id);
auto token = t.nextToken();
if (token.IsInteger()) {
  if (token.AsInt32() < 0)
    return error;

  uint32_t val = token.AsUint32();
  token = t.nextToken();
  if (token.IsEOS() || token.IsEOL()) {
    *desc_set = 0;
    *binding = val;
  }
 *desc_set = val;
}
if (!token.IsString() || token.AsString() != ':')
  return error;

token = t.nextToken();
if (!token.IsInteger() || token.AsInt32() < 0)
  return error;

*binding = token->AsUint32();
```

jaebaek: done

src/descriptor_set_and_binding_parser.h
```diff
@@ -0,0 +1,51 @@
...
+  /// |buffer_id|.
+  Result Parse(const std::string& buffer_id);
+
+  /// Return descriptor set that is the result of Parse().
+  uint32_t GetDescriptorSet() const { return descriptor_set_; }
+
+  /// Return binding that is the result of Parse().
+  uint32_t GetBinding() const { return binding_; }
+
+ protected:
```
dj2: private

afd:
I'm actively using the functionality this PR provides - would it be possible to get this merged into master now?

afd:
The amber help says:

"  -B [<desc set>:]<binding>     -- Descriptor set and binding of buffer to write.
                                   Default is [0:]0."

However, I find that I don't get anything dumped when I provide -b but no -B, whereas I get buffer 0 dumped when I pass -b and -B 0.  So I guess this default is not getting set.

PR #270, opened 2019-02-07
Vulkan: support disabling validation layer

### LINKED ISSUE #233: Add an option to disable vulkan validation
I recently updated to 3e4f538eb4e54cf5ec97f66f026bb00809661940 and it seems validation is now on by default. That's fine, but I would like a way to disable it via a command line option.

jaebaek:
Fixes #233


jaebaek:
This must be submitted after submitting #266

PR #272, opened 2019-02-07
Add flag to set vulkan version.

dj2:
This CL adds a flag to the sample app to allow specifing the version of
vulkan to use in the apiVersion field.

### DIFF
--- a/samples/amber.cc
+++ b/samples/amber.cc
@@ -35,6 +35,8 @@ struct Options {
   std::string image_filename;
   std::string buffer_filename;
   int64_t buffer_binding_index = 0;
+  uint32_t engine_major = 1;
+  uint32_t engine_minor = 0;
   bool parse_only = false;
   bool disable_validation_layer = false;
   bool show_summary = false;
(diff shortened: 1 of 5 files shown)


dj2:
@alan-baker

samples/amber.cc
```diff
@@ -116,9 +119,32 @@ bool ParseArgs(const std::vector<std::string>& args, Options* opts) {
...
       opts->show_help = true;
+    } else if (arg == "-v") {
+      ++i;
+      if (i >= args.size()) {
+        std::cerr << "Missing value for -v argument." << std::endl;
+        return false;
+      }
+      const std::string& ver = args[i];
+
+      size_t dot_pos = 0;
```
dneto0: gosh, I would reach for sscanf first.

  int major=0, minor=0;
  const auto num_conversions = sscanf(args[i], "%d.%d%c", &major, minor);
  if ( num_conversions == 3 ) { ... error: bad suffix }
  if ( num_conversions < 2 ) { error: invalid parse.}
  else ... now you have major and minor..

But hey, I'm old school.

dj2: Heh, but this is fancy and cool.

PR #273, opened 2019-02-07
Allow setting SPIR-V target environment.

dj2:
This CL adds an option to the amber sample application to set the SPIR-V
target environment. The environment name matches the names accepted by
the other spirv-tools applications.

The envionment is set into the script and used by the shader compiler.

### DIFF
--- a/include/amber/amber.h
+++ b/include/amber/amber.h
@@ -57,6 +57,8 @@ struct Options {
   EngineType engine;
   /// Holds engine specific configuration. Ownership stays with the caller.
   EngineConfig* config;
+  /// The SPIR-V environment to target.
+  uint32_t spv_env;
   /// Lists the buffers to extract at the end of the execution
   std::vector<BufferInfo> extractions;
 };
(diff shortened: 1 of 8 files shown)


dj2:
This will require rolling spirv-tools once the CL lands to expose spvParseTargetEnv.

@alan-baker

include/amber/amber.h
```diff
@@ -57,6 +57,8 @@ struct Options {
   EngineType engine;
   /// Holds engine specific configuration. Ownership stays with the caller.
   EngineConfig* config;
+  /// The SPIR-V environment to target.
+  uint32_t spv_env;
```
dj2: We use uint32_t throughout instead of spv_target_env because amber can be compiled without spirv-tools. So, we can't use the spirv-tools env. We wait until the last minute to cast the uint32_t to the spv_target_env.

alan-baker: You could store the environment string and defer parsing until compiling the shader where you already have the SPIRV-Tools requirement. You might parse it once per shader though in that case.

dj2: That's nicer, done. Also rolled DEPS to pickup new spirv-tools and spirv-headers.

PR #280, opened 2019-02-11
Fix clang 5.0 compile error

### LINKED ISSUE #278: Amber does not build with clang 5.0.
I tried compiling amber with clang-5.0, and I get this error:

```
../samples/config_helper_vulkan.cc:736:27: error: zero as null pointer constant [-Werror,-Wzero-as-null-pointer-constant]
  if (vulkan_instance_ != VK_NULL_HANDLE)
                          ^~~~~~~~~~~~~~
                          nullptr
/usr/include/vulkan/vulkan.h:49:24: note: expanded from macro 'VK_NULL_HANDLE'
#define VK_NULL_HANDLE 0
```

jaebaek:
Fixes #278


samples/CMakeLists.txt
```diff
@@ -31,6 +31,13 @@ target_include_directories(amber PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/..")
 set_target_properties(amber PROPERTIES OUTPUT_NAME "amber")
 target_link_libraries(amber libamber ${VULKAN_LIB})
 
+if(${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang")
+  # vulkan/vulkan.h defines VK_NULL_HANDLE as 0u and that also serves as a null pointer.
+  # Disable Clang's warning that will alwaays fire on that.  This is required to build
+  # with XCode 10.
+  target_compile_options(amber PRIVATE -Wno-zero-as-null-pointer-constant)
```
dj2: Where is the error coming from? I'd rather do these as targeted disables if possible.

jaebaek: Vulkan related handles are pointers i.e., VkInstance. Because of `-Wzero-as-null-pointer-constant` in clang, initializing VkInstance with `VK_NULL_HANDLE` which is actually `0u` has compile errors i.e., "error: zero as null pointer constant [-Werror,-Wzero-as-null-pointer-constant]".

This is known issue and we solved it by giving `-Wno-zero-as-null-pointer-constant` as a compile option.

I cannot understand what you meant by `targeted disables`. Could you please give me more information how I can make it better?

dj2: By targeted disables I mean with a clang pragma line like in include/amber/vulkan_header.h. You need this just for config_helper_vulkan.{h|cc} right. So, I'd rather not turn this off for everything in samples/amber because those files have this weirdness.

You can wrap the pragmas around the sections where it's an issue to just turn the error off in a smaller bit of code, so place around the whole header and the Shutdown method.

jaebaek: done

PR #284, opened 2019-02-12
Generate default pipeline in VkScript.

dj2:
This CL adds the code to generate a default pipeline and set the various
buffers/shaders into that pipeline from VkScript.

### DIFF
--- a/src/executor_test.cc
+++ b/src/executor_test.cc
@@ -294,7 +294,7 @@ logicOp)";
   auto color_frame_format = ToStub(engine.get())->GetColorFrameFormat();
   auto depth_frame_format = ToStub(engine.get())->GetDepthFrameFormat();
   EXPECT_EQ(FormatType::kB8G8R8A8_UNORM, color_frame_format);
-  EXPECT_EQ(FormatType::kD32_SFLOAT_S8_UINT, depth_frame_format);
+  EXPECT_EQ(FormatType::kUnknown, depth_frame_format);
 }
 
 TEST_F(VkScriptExecutorTest, ExecutesRequiredExtensions) {
@@ -327,7 +327,7 @@ VK_KHR_variable_pointers)";
   auto color_frame_format = ToStub(engine.get())->GetColorFrameFormat();
   auto depth_frame_format = ToStub(engine.get())->GetDepthFrameFormat();
   EXPECT_EQ(FormatType::kB8G8R8A8_UNORM, color_frame_format);
-  EXPECT_EQ(FormatType::kD32_SFLOAT_S8_UINT, depth_frame_format);
+  EXPECT_EQ(FormatType::kUnknown, depth_frame_format);
 }
 
 TEST_F(VkScriptExecutorTest, ExecutesRequiredFrameBuffers) {
@@ -388,7 +388,7 @@ fence_timeout 12345)";
   auto color_frame_format = ToStub(engine.get())->GetColorFrameFormat();
   auto depth_frame_format = ToStub(engine.get())->GetDepthFrameFormat();
   EXPECT_EQ(FormatType::kB8G8R8A8_UNORM, color_frame_format);
-  EXPECT_EQ(FormatType::kD32_SFLOAT_S8_UINT, depth_frame_format);
+  EXPECT_EQ(FormatType::kUnknown, depth_frame_format);
 }
 
 TEST_F(VkScriptExecutorTest, ExecutesRequiredAll) {
--- a/src/executor_test.cc
+++ b/src/executor_test.cc
@@ -293,8 +293,8 @@ logicOp)";
 
   auto color_frame_format = ToStub(engine.get())->GetColorFrameFormat();
   auto depth_frame_format = ToStub(engine.get())->GetDepthFrameFormat();
-  EXPECT_EQ(FormatType::kUnknown, color_frame_format);
-  EXPECT_EQ(FormatType::kUnknown, depth_frame_format);
+  EXPECT_EQ(FormatType::kB8G8R8A8_UNORM, color_frame_format);
+  EXPECT_EQ(FormatType::kD32_SFLOAT_S8_UINT, depth_frame_format);
 }
 
 TEST_F(VkScriptExecutorTest, ExecutesRequiredExtensions) {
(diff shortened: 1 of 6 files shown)


src/vkscript/parser.cc
```diff
@@ -176,6 +183,52 @@ Result Parser::Parse(const std::string& input) {
   return {};
 }
 
+Result Parser::GenerateDefaultPipeline(SectionParser* section_parser) {
```
jaebaek: `const SectionParser& section_parser` would be better?

src/vkscript/parser.cc
```diff
@@ -204,7 +257,12 @@ Result Parser::ProcessShaderBlock(const SectionParser::Section& section) {
   shader->SetFormat(section.format);
   shader->SetData(section.contents);
 
-  Result r = script_->AddShader(std::move(shader));
+  Result r = script_->GetPipeline(kDefaultPipelineName)
```
jaebaek: Here the pipeline name is hard-coded as `kDefaultPipelineName`.
Does it mean we will have only the default pipeline?

src/pipeline.cc
```diff
@@ -31,8 +31,8 @@ Pipeline::ShaderInfo::~ShaderInfo() = default;
 
 const char* Pipeline::kDefaultColorBufferFormat = "B8G8R8A8_UNORM";
 const char* Pipeline::kDefaultDepthBufferFormat = "D32_SFLOAT_S8_UINT";
```
jaebaek: VkRunner assumes that there is no depth buffer in default.
I agree with the idea that no depth buffer in default would be less unexpected behavior because some rendering result would be different because of depth buffer.

Is there any special reason for this?
(sorry for not claiming this when you wrote Amber script doc)

src/vkscript/parser.cc
dj2: I did that originally, but then I run into const'ness issues later on. This makes it work for now we can dig into const'ness later.

src/vkscript/parser.cc
dj2: I have a feeling vkscript will only have one pipeline, similar to how it works now. AmberScript will have multiple pipelines.

src/pipeline.cc
dj2: So, you don't think we should have a depth buffer in amber either? I can remove it in VkScript, but do you think the general user would expect there to be a depth buffer no no depth buffer? @dneto0 do you have any idea what the expected would be?

src/vkscript/parser.cc
jaebaek: I do not have strong opinion on this so it is fine but the following patch works fine on my laptop.
Please adopt it if you think it is fine:

```
diff --git a/src/vkscript/parser.cc b/src/vkscript/parser.cc
index 1e0415c..1e88e4e 100644
--- a/src/vkscript/parser.cc
+++ b/src/vkscript/parser.cc
@@ -170,7 +170,7 @@ Result Parser::Parse(const std::string& input) {
   if (!r.IsSuccess())
     return r;
 
-  r = GenerateDefaultPipeline(&section_parser);
+  r = GenerateDefaultPipeline(section_parser);
   if (!r.IsSuccess())
     return r;
 
@@ -183,10 +183,10 @@ Result Parser::Parse(const std::string& input) {
   return {};
 }
 
-Result Parser::GenerateDefaultPipeline(SectionParser* section_parser) {
+Result Parser::GenerateDefaultPipeline(const SectionParser& section_parser) {
   // Generate a pipeline for VkScript.
   PipelineType pipeline_type = PipelineType::kCompute;
-  for (const auto& section : section_parser->Sections()) {
+  for (const auto& section : section_parser.Sections()) {
     if (!SectionParser::HasShader(section.section_type))
       continue;
 
diff --git a/src/vkscript/parser.h b/src/vkscript/parser.h
index a48e605..70e30d1 100644
--- a/src/vkscript/parser.h
+++ b/src/vkscript/parser.h
@@ -38,7 +38,7 @@ class Parser : public amber::Parser {
 
  private:
   std::string make_error(const Tokenizer& tokenizer, const std::string& err);
-  Result GenerateDefaultPipeline(SectionParser* section_parser);
+  Result GenerateDefaultPipeline(const SectionParser& section_parser);
   Result ProcessSection(const SectionParser::Section& section);
   Result ProcessShaderBlock(const SectionParser::Section& section);
   Result ProcessRequireBlock(const SectionParser::Section& section);
diff --git a/src/vkscript/section_parser.h b/src/vkscript/section_parser.h
index 3acecab..3d898c3 100644
--- a/src/vkscript/section_parser.h
+++ b/src/vkscript/section_parser.h
@@ -50,7 +50,7 @@ class SectionParser {
   ~SectionParser();
 
   Result Parse(const std::string& data);
-  const std::vector<Section>& Sections() { return sections_; }
+  const std::vector<Section>& Sections() const { return sections_; }
 
   Result SplitSectionsForTesting(const std::string& data) {
     return SplitSections(data);
```

dj2: Hah, for some reason I thought Sections was const already, so when I got the build error thought something else was going wrong. Thanks for digging into it, will update this CL.

dj2: done, thanks.

src/pipeline.cc
dj2: Updated VkScript to not generate a depth buffer by default. Left amberscript to generate one. We can decide if that's the right route for amberscript in the future.

dneto0: I expect no depth buffer by default.

dj2: K, this CL is already updated. I'll make the amber change in a separate CL as it's unrelated to the rest of this CL.

src/executor_test.cc
```diff
@@ -293,7 +293,7 @@ logicOp)";
 
   auto color_frame_format = ToStub(engine.get())->GetColorFrameFormat();
   auto depth_frame_format = ToStub(engine.get())->GetDepthFrameFormat();
-  EXPECT_EQ(FormatType::kUnknown, color_frame_format);
+  EXPECT_EQ(FormatType::kB8G8R8A8_UNORM, color_frame_format);
```
dneto0: I'm surprised this is not kR8G8B8A8_UNORM.  (i,e. rgba order instead of bgra)

src/pipeline.cc
dneto0: Agree depth/stencil buffer should not be used by default.
Vkrunner says about depthstencil directive:   "Without it no depth-stencil buffer will be created."

src/executor_test.cc
dj2: That's the default vkrunner uses: https://github.com/Igalia/vkrunner/blob/master/vkrunner/vr-script.c#L2466

src/pipeline.cc
dj2: Yup, I've done that for VkScript already. Will handle updating AmberScript to not have a default depth buffer in a followup.

PR #286, opened 2019-02-12
 Execute from pipeline, not script.

dj2:
This CL changes the executor to read information from the pipeline
instead of using the script object.


dj2:
This needs to be tested on a Linux machine before landing. I have 8 tests on my mac which fail with and without the patch, so it at least seems to be the same but we need to verify the Linux run passes all tests before landing.

dj2:
For context, the next CL in the chain is: https://github.com/dj2/amber/commit/87cb488e838716c7301c9a1f9725dbbafeb58567 which moves all this into CreatePipeline. The SetShader and SetBuffer methods go away and the engines set themselves up as they need based on what the pipeline contains.

src/executor.cc
```diff
@@ -32,65 +32,73 @@ Result Executor::Execute(Engine* engine,
...
+    if (!r.IsSuccess())
+      return r;
+  }
 
-    Result r = engine->SetBuffer(
-        buf->GetBufferType(), buf->GetLocation(),
-        buf->IsFormatBuffer() ? buf->AsFormatBuffer()->GetFormat() : Format(),
-        buf->GetData());
+  if (pipeline->GetDepthBuffer().buffer) {
+    const auto& info = pipeline->GetDepthBuffer();
```
dneto0: Question: What happens if the first pipeline has a depth buffer but the second one doesn't ?  When is the engine supposed to reset its state?

dj2: VkScript only has one pipeline. There is no concept so I create a fake one. So, for vkscript this will work fine.

For AmberScript we need to update the engine code to handle multiple pipelines so each setBuffer and setShader has to work per-pipeline not globally. I'm leaving that for a future enhancement as, I believe, this is the minimal change to get VkScript to work and we can update to make AmberScript work.

For AmberScript each command RUN, etc, has the pipeline attached so the engine will know what state needs to be set to run a given command. I'm going to fake that in VkScript as well and attach the fake pipeline to each command, just in a future CL.

Issue #277, opened 2019-02-08
Memory leaks in amber


s-perron:
`EngineConfig` does not have a virtual destructor, so `VulkanEngineConfig` do not get destroyed properly.

In `PushConstant::PushConstant`, an array of uint8_t is allocated, but the pointer is kept in a unique_ptr<uint8_t>.

I found these using ASAN.  There is another leak reported, but it is very confusing.  I cannot tell where it comes from.

dj2:
@dneto0 has a CL for the virtual destructor.

@jaebaek can you look into the PushConstant code when you get back? Can we use a vector instead of an array?

dneto0:
My CL got into trouble.  I've stopped looking at it.

dj2:
Fixed.

PR #287, opened 2019-02-12
Add virtual destructor to engine configs

dneto0:
Partially address #277

### DIFF
--- a/include/amber/amber_dawn.h
+++ b/include/amber/amber_dawn.h
@@ -22,6 +22,8 @@ namespace amber {
 
 /// Configuration for the Dawn engine.
 struct DawnEngineConfig : public EngineConfig {
+  virtual ~DawnEngineConfig() override;
+
   /// The Dawn Device to use for running tests.
   ::dawn::Device* device;
 };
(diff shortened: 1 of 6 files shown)


include/amber/amber_dawn.h
```diff
@@ -22,6 +22,8 @@ namespace amber {
 
 /// Configuration for the Dawn engine.
 struct DawnEngineConfig : public EngineConfig {
+  virtual ~DawnEngineConfig() = default;
```
dj2: ~DawnEngineConfig() override = default;

include/amber/amber_vulkan.h
```diff
@@ -27,6 +27,8 @@ namespace amber {
 /// Configuration for the Vulkan Engine. The following are all required when
 /// when using the vulkan backend.
 struct VulkanEngineConfig : public EngineConfig {
+  virtual ~VulkanEngineConfig() = default;
```
dj2: ditto

include/amber/amber_dawn.h
dneto0: done.

dneto0:
Rebased.  Also put the destructor definition for each of the three classes into a .cc  in the right lib.  That should fix the clang complaints about not having a home for the vtable.

dj2:
I'm going to close this so it drops out of the PR queue since it's not being worked on at the moment.

PR #289, opened 2019-02-13
 Move pipeline setup out of executor

dj2:
This CL moves the code to setup shaders and buffers out of the executor
and into the CreatePipeline method. This way each engine can determine
how to setup the data needed.


dj2:
Thoughts? Does this seem like it's better as an engine concern but with the possibility of duplication?

PR #291, opened 2019-02-13
More casts to make CTS happy.

dj2:


dj2:
TBRing to try to pass CTS.

PR #295, opened 2019-02-14
Initialize the impl_ field of RecipeImpl to null, to avoid problems d…

### LINKED ISSUE #294: Amber segfaults after printing error when given an invalid name.
Run amber on the attached file to reproduce.
[bad.shader_test.txt](https://github.com/google/amber/files/2864725/bad.shader_test.txt)

afd:
…eleting it if it has not been initialized.

Fixes #294

### DIFF
--- a/src/recipe.cc
+++ b/src/recipe.cc
@@ -19,7 +19,7 @@ RecipeImpl::RecipeImpl() = default;
 
 RecipeImpl::~RecipeImpl() = default;
 
-Recipe::Recipe() = default;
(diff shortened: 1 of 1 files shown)


afd:
I had wanted to include a test to check that the Amber program I attached to the issue is rejected by Amber but not with a segfault, but I couldn't see a natural place to add such a test.

dj2:
Landed this so the fix is in and we can figure out how to do a test case.  Can you repro it in a unit test somehow?

PR #297, opened 2019-02-18
Fix Android plain executable: force Vulkan, disable validation layers

hevrard:


hevrard:
Kokoro build with NDK r15 fails with a series of errors like:
`../src/vulkan/graphics_pipeline.cc:252:14: error: 'VK_BLEND_OP_ZERO_EXT' was not declared in this scope`

This is apparently due to r15 vulkan headers which are quite out of date and lack these definitions. The update to NDK r16 (currently the latest available on kokoro instances) fixes the build.

PR #298, opened 2019-02-18
Add image export to farbfeld format

hevrard:
Currently image export is done in PPM format, which unfortunately lacks support of alpha channel. Alpha channel support is crucial for graphicsfuzz experiments.

This change adds a possible export to farbfeld format, which both supports alpha channel and is trivial to implement, bypassing the need for a third party library. For more info, see https://tools.suckless.org/farbfeld/faq/

EDIT: PNG export would be nice as PNG is supported virtually everywhere. I prepared this farbfeld export because I assumed adding a third-party PNG library (e.g. lodepng https://github.com/lvandeve/lodepng) as an Amber dependency would be frowned upon. Yet if you can see a way of easily adding PNG export support, please let me know!

### DIFF
--- a/samples/farbfeld.cc
+++ b/samples/farbfeld.cc
@@ -1,0 +1,83 @@
+// Copyright 2019 The Amber Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "samples/farbfeld.h"
+
+#include <cassert>
+
+#include "amber/result.h"
+#include "amber/value.h"
+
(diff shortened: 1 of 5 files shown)


dj2:
I think I'd prefer png to farbfeld as I can't find much on image editor support for farbfeld (can I open it in preview on Mac or what software do I need on Linux?)

I don't have an issue with adding the extra dependency to the samples/ app.

hevrard:
See #301 for PNG export.

samples/farbfeld.cc
```diff
@@ -0,0 +1,83 @@
...
+#include <cassert>
+
+#include "amber/result.h"
+#include "amber/value.h"
+
+namespace farbfeld {
+
+namespace {
+
+char byte0(uint32_t word) {
```
dneto0: I'd use uint8_t instead of char.

samples/farbfeld.cc
```diff
@@ -0,0 +1,83 @@
...
+std::pair<amber::Result, std::string> ConvertToFarbfeld(
+    uint32_t width,
+    uint32_t height,
+    const std::vector<amber::Value>& values) {
+  assert(values.size() == width * height);
+
+  // Farbfeld format details: https://tools.suckless.org/farbfeld/
+
+  // Farbfeld header
+  std::string image = "farbfeld";
```
dneto0: Recommend adding image.reserve(8 +4 +4 + width * height * 4 * 2);

PR #299, opened 2019-02-19
Fixup image buffer dump

dj2:
This CL updates the image dumping code to the new default framebuffer format of B8G8R8A8_UNORM.

### DIFF
--- a/samples/ppm.cc
+++ b/samples/ppm.cc
@@ -52,10 +52,10 @@ std::pair<amber::Result, std::string> ConvertToPPM(
   // Write PPM data
   for (amber::Value value : values) {
     const uint32_t pixel = value.AsUint32();
-    // We assume R8G8B8A8_UINT here:
-    image.push_back(byte0(pixel));  // R
+    // We assume B8G8R8A8_UNORM here:
+    image.push_back(byte2(pixel));  // R
     image.push_back(byte1(pixel));  // G
-    image.push_back(byte2(pixel));  // B
+    image.push_back(byte0(pixel));  // B
     // PPM does not support alpha channel
   }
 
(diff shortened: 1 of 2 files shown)


dj2:
@hevrard I believe this should fix the issue you encountered.

samples/ppm.cc
```diff
@@ -52,10 +52,10 @@ std::pair<amber::Result, std::string> ConvertToPPM(
   // Write PPM data
   for (amber::Value value : values) {
     const uint32_t pixel = value.AsUint32();
-    // We assume R8G8B8A8_UINT here:
-    image.push_back(byte0(pixel));  // R
+    // We assume B8G8R8A8_UNORM here:
+    image.push_back(byte2(pixel));  // R
```
hevrard: The comment is out of sync here, it should read BGRA

samples/ppm.cc
```diff
@@ -52,10 +52,10 @@ std::pair<amber::Result, std::string> ConvertToPPM(
...
     image.push_back(byte1(pixel));  // G
-    image.push_back(byte2(pixel));  // B
+    image.push_back(byte0(pixel));  // B
```
hevrard: idem comment out of sync

samples/ppm.cc
dj2: Hah, good catch. Updated the top comment and missed all the inline ones.

dj2: Actually, I think these are right. We're writting the R value here, not the B value. It just happens that R is in byte2.

Does that make sense?

hevrard: Woops my bad, I got confused, the comments are correct indeed!

PR #300, opened 2019-02-19
Fixup sample app buffer parsing

dj2:
This CL fixes the -B flag to default to 0:0 if not provided. It also
fixes the code to use : as the separator instead of ,


dj2:
@afd I believe this should fixup your default -B issue. It also corrects the parsing to look for : instead of ,.

dj2:
TBR-ing to fix the incorrect parsing.

PR #301, opened 2019-02-19
Add image dump in PNG

hevrard:
This change adds support to dump image to PNG format, it relies on the third party 'lodepng' library (zlib license).

Maybe PPM support can be dropped subsequently, but it would break backward compatibility with vkrunner.

Note that in the cmake we add `third_party/` as an include directory in order to please the linter about the `import` line of lodepng, which the linter insists to have the directory in it.

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -25,7 +25,7 @@ deps = {
       vars['glslang_revision'],
 
   'third_party/lodepng': vars['lvandeve_git'] + '/lodepng.git@' +
-      vars['glslang_revision'],
+      vars['lodepng_revision'],
 
   'third_party/shaderc': vars['google_git'] + '/shaderc.git@' +
       vars['shaderc_revision'],
--- a/DEPS
+++ b/DEPS
@@ -3,10 +3,12 @@ use_relative_paths = True
 vars = {
   'google_git':  'https://github.com/google',
   'khronos_git': 'https://github.com/KhronosGroup',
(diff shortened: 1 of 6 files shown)


DEPS
```diff
@@ -22,6 +24,9 @@ deps = {
   'third_party/glslang': vars['khronos_git'] + '/glslang.git@' +
       vars['glslang_revision'],
 
+  'third_party/lodepng': vars['lvandeve_git'] + '/lodepng.git@' +
+      vars['glslang_revision'],
```
dj2: This should be lodepng_revision not glslang_revision

samples/png.cc
```diff
@@ -0,0 +1,77 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "samples/png.h"
+
+#include <cassert>
+
+#include "amber/result.h"
+#include "amber/value.h"
+#include "lodepng/lodepng.h"
```
dj2: #include "third_party/lodepng/lodepng.h" ? If you do that, can you drop the adding of third_party/ to the include path?

dj2:
You're probably going to have to move the building of lodepng into the third_party/CmakeLists.txt file, have it build a library and then you can disable all of the warnings it triggers similar to how we do other third_party dependencies.

dneto0:
This is superceded by #302 which has been merged.

PR #302, opened 2019-02-20
Add image dump in PNG

dj2:
This change adds support to dump image to PNG format, it relies on the
third party 'lodepng' library (zlib license).

This CL is based on Pull #301 by @hevrard.


dj2:
@hevrard I think this should work ....

samples/amber.cc
```diff
@@ -325,10 +326,18 @@ int main(int argc, const char** argv) {
 
     if (!options.image_filename.empty()) {
       std::string image;
+      bool usePNG = options.image_filename.compare(
+          options.image_filename.length() - 5, options.image_filename.length(),
```
dneto0: What if the string is very short, e.g. length is 2, then 2-5 is large negative that is then converted to unsigned....  out of bounds?

samples/png.cc
```diff
@@ -0,0 +1,81 @@
...
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wweak-vtables"
+#include "third_party/lodepng/lodepng.h"
+#pragma clang diagnostic pop
+
+namespace png {
+
+namespace {
+
+unsigned char byte0(uint32_t word) {
```
dneto0: Normally I'd ask for uint8_t here. But lodepng wants a vector of unsigned char.  So I'll let this pass.

samples/png.cc
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  std::vector<unsigned char> png;
+  unsigned error = lodepng::encode(png, data, width, height);
+  if (error) {
+    return std::make_pair(amber::Result("lodepng::encode() returned non-zero"),
+                          nullptr);
+  }
+
+  std::string image;
```
dneto0: Would recommend this instead:
std::string image(png.data(), png.size())

samples/amber.cc
dj2: done

samples/png.cc
dj2: ack

samples/png.cc
dj2: std::string(png.begin(), png.end()) works

PR #303, opened 2019-02-20
Add -ps option to allow compiling shaders.

dj2:
This CL adds a -ps option to the sample app to allow parsing the script
and executing the shader compiler. The script is not run against the
engine (and the engine is not created).

### DIFF
--- a/src/executor.cc
+++ b/src/executor.cc
@@ -48,7 +48,8 @@ Result Executor::CompileShaders(const amber::Script* script,
 
 Result Executor::Execute(Engine* engine,
                          const amber::Script* script,
-                         const ShaderMap& shader_map) {
+                         const ShaderMap& shader_map,
+                         ExecutionType executionType) {
   engine->SetEngineData(script->GetEngineData());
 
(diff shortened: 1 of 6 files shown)


dj2:
@afd Does this do what you were looking for?

src/executor.cc
```diff
@@ -40,22 +58,12 @@ Result Executor::Execute(Engine* engine,
...
-
-    shader_info.SetData(std::move(data));
-  }
+  Result r = CompileShaders(script, shader_map);
+  if (!r.IsSuccess())
+    return r;
 
-  Result r = engine->CreatePipeline(pipeline);
+  auto* pipeline = script->GetPipelines()[0].get();
+  r = engine->CreatePipeline(pipeline);
```
dneto0: This is where the driver compiler kicks in.   So I don't think this is what @afd  wanted.

This MR only does the "front end" compilation: from shader in the script down to SPIR-V.

dj2: How about now?

PR #304, opened 2019-02-20
Differentiate between instance and device extensions.

dj2:
This CL adds the necessary code to differentiate that
VK_KHR_get_physical_device_properties2 is an instance extension and not
a device extension.


dj2:
@alan-baker this is is 1/4 of the variable pointers CLs.

PR #306, opened 2019-02-20
Add sample example.

### LINKED ISSUE #292: Provide link to sample in README
The Amber README has a section called "Amber Sample", but this section doesn't actually point the reader to a sample script file they could try.

After:

"
The build will generate an out/Debug/amber executable which can be used to run amber scripts. The script can be used as out/Debug/amber <path to amber file>. Where, currently, the amber file is in the VkScript format.
"
it would be nice to have an example command pointing to some specific amber file in the repo, with an expected output.

dj2:
This CL adds an example to the README of running the sample app and
documents the expected output.

Fixes #292.


dj2:
@afd does this work?

PR #307, opened 2019-02-20
Flip the sample app to be less quiet.

dj2:
The sample app outputting just an error code by default is confusing to
users. The -s flag was originally created to turn on the summary mode.
This CL removes -s from the help (but it still parses) and adds a -q
flag instead. The quiet flag is used by the test runner to suppress the
summary output.


dj2:
@alan-baker

samples/amber.cc
```diff
@@ -147,8 +148,11 @@ bool ParseArgs(const std::vector<std::string>& args, Options* opts) {
       opts->parse_only = true;
     } else if (arg == "-d") {
       opts->disable_validation_layer = true;
+    // -s is an old option which is still recognized but does nothing.
     } else if (arg == "-s") {
       opts->show_summary = true;
```
alan-baker: Maybe remove `show_summary` (since it seems unused) and make this `quiet = false` (or ignore)?

dj2: good point, done.

samples/amber.cc
```diff
@@ -151,8 +151,11 @@ bool ParseArgs(const std::vector<std::string>& args, Options* opts) {
       opts->pipeline_create_only = true;
     } else if (arg == "-d") {
       opts->disable_validation_layer = true;
+      // -s is an old option which is still recognized but does nothing.
```
dneto0: Well, it does something: It reverses the -q option...

dj2: done

PR #308, opened 2019-02-20
 Allow querying vkGetPhysicalProperties2

dj2:
This CL adds the code to query for physical properties2 if the
VK_KHR_get_physical_device_properties2 extension is required.

### DIFF
--- a/samples/config_helper_vulkan.cc
+++ b/samples/config_helper_vulkan.cc
@@ -607,6 +607,13 @@ amber::Result ConfigHelperVulkan::CreateVulkanInstance(
     }
   }
 
+  // Determine if VkPhysicalDeviceProperties2KHR should be used
+  for (auto& ext : required_extensions) {
+    if (ext == "VK_KHR_get_physical_device_properties2") {
+      use_physical_device_features2_ = true;
+    }
+  }
+
   std::vector<const char*> required_extensions_in_char;
   std::transform(
       required_extensions.begin(), required_extensions.end(),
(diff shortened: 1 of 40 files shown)


dj2:
@alan-baker 2/4

samples/config_helper_vulkan.cc
```diff
@@ -662,7 +669,20 @@ amber::Result ConfigHelperVulkan::ChooseVulkanPhysicalDevice(
...
+    if (use_physical_device_features2_) {
+      VkPhysicalDeviceFeatures2KHR features2 = {};
+      features2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2_KHR;
+      features2.pNext = nullptr;
+
+      auto vkGetPhysicalDeviceFeatures2KHR =
+          reinterpret_cast<PFN_vkGetPhysicalDeviceFeatures2KHR>(
+              vkGetInstanceProcAddr(vulkan_instance_,
+                                    "vkGetPhysicalDeviceFeatures2KHR"));
+      vkGetPhysicalDeviceFeatures2KHR(physical_devices[i], &features2);
```
dneto0: Will this code even link if physical-device-features2 is not supported?

dj2: As long as the structures are in the header file it's all good. For the actual function call, we use the getInstanceProcAddr call, so that should be fine as long as the PFN_vkGetPhysicalDeviceFeatures2KHR  in the header as well.

I used the KHR version for everything to make sure it was backwards compatible.

PR #309, opened 2019-02-21
 Hold required features as strings.

dj2:
This CL keeps the required features as strings as long as possible. This
will allow extending the list of features with additional entries.


dj2:
Actually, I think this CL is silly. I'm going to close and rebase the vptr Cl on top of master and make it work with the feature enum.

dj2:
Actually, this does make sense, heh.

dj2:
The reason it makes sense is the current way has a lot of conversions due to the recipe API.

 * parser does string -> feature
 * script does feature -> string (through GetRequiredFeatures in the recipe)
 * device does string -> feature to get back to features again

If we just hold strings we just pass them along ....

PR #313, opened 2019-02-21
 Allow specifiying variable pointers in requirements.

dj2:
This CL adds the two requirements:
 * VariablePointerFeatures.variablePointers
 * VariablePointerFeatures.variablePointersStorageBuffer

 Variable pointers requires VK_KHR_get_physical_properties2 and the
 variable pointer extension are both requested.


dj2:
@alan-baker

samples/config_helper_vulkan.cc
```diff
@@ -757,14 +791,33 @@ amber::Result ConfigHelperVulkan::CreateDeviceWithFeatures1(
...
+    // No dot means this is a features1 feature.
+    if (feature.find_first_of('.') == std::string::npos) {
+      feature1_names.push_back(feature);
+      continue;
+    }
+
+    if (feature == kVariablePointers)
+      variable_pointers_feature_.variablePointers = VK_TRUE;
+    else if (feature == kVariablePointersStorageBuffer)
+      variable_pointers_feature_.variablePointersStorageBuffer = VK_TRUE;
```
dj2: Comment from @alan-baker:

"The spec says if you turn on VK_KHR_variable_pointers you MUST support variablePointersStorageBuffer so I'd default that to true if the extension or VariablePointerFeatures.variablePointers is mentioned"

Do we want the implicit setting to true, or should we make the script have to be explicit?

samples/config_helper_vulkan.cc
```diff
@@ -183,7 +187,7 @@ amber::Result NamesToVulkanFeatures(
     } else if (name == "inheritedQueries") {
       features->inheritedQueries = VK_TRUE;
     } else {
-      return amber::Result("Sample: Unknown Vulkan feature");
+      return amber::Result("Sample: Unknown Vulkan feature: " + name);
```
dneto0: this compiles!?
TIL

I expected the LHS of the + to be a std::string

samples/config_helper_vulkan.cc
dneto0: I'm ok with forcing the script to be explicit.  I think.

alan-baker: What about VariablePointerFeatures.variablePointers implying both variables?

samples/config_helper_vulkan.cc
dj2: amber::Result takes a `const std::string&` as input. So, it's probably converting the first const char* to std::string then doing the add.

samples/config_helper_vulkan.cc
dj2: I'm leaning towards explicit so yo have to specify both. It makes the test a lot clearer and removes spooky action at a distance.

PR #316, opened 2019-02-21
Output requested images/buffers on probe failure.

### LINKED ISSUE #314: Output image even if probes fail.
It seems like amber does not output the image (`-i` option) when one of the probes fails.  It would be useful to be able to see the image when things fail, so I can more easily see what is wrong.

dj2:
This Cl updates the sample app to allow dumping of image and buffer data
even if the probes fail to execute.

Fixes #314.

### DIFF
--- a/samples/amber.cc
+++ b/samples/amber.cc
@@ -336,7 +336,8 @@ int main(int argc, const char** argv) {
     if (!result.IsSuccess()) {
       std::cerr << file << ": " << result.Error() << std::endl;
       failures.push_back(file);
-      continue;
+      // Note, we continue after failure to allow dumping the buffers which may
+      // give clues as to the failure.
     }
 
     if (!options.image_filename.empty()) {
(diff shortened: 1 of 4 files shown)


s-perron:
With this change, if I have `-i p.ppm` on the command line and the script fails because the validation layer complains, I do not see the error message because there is a segmenation fault:

```
➜  entrypoint git:(master) ✗ /usr/local/google/home/stevenperron/cts/amber/build.release/amber -s geom_ssbo_entry_point.vkscript
geom_ssbo_entry_point.vkscript: Invalid shader: error: line 66: 2nd operand of TypePointer: operand 12 requires one of these extensions: SPV_KHR_storage_buffer_storage_class SPV_KHR_variable_pointers 
  %_ptr_StorageBuffer_input_buffer = OpTypePointer StorageBuffer %input_buffer



Summary of Failures:
  geom_ssbo_entry_point.vkscript

Summary: 0 pass, 1 fail
➜  entrypoint git:(master) ✗ /usr/local/google/home/stevenperron/cts/amber/build.release/amber -s geom_ssbo_entry_point.vkscript -i p.ppm
[1]    158235 segmentation fault  /usr/local/google/home/stevenperron/cts/amber/build.release/amber -s  -i p.pp
➜  entrypoint git:(master) ✗ 
```

dj2:
K, the pipeline creation failure should be handled now. We output the extractor error, and not the dump error as I think that makes the most sense.

PR #317, opened 2019-02-21
[vulkan] Don't store pipeline_layout or pipeline.

dj2:
Currently the pipeline_layout and pipeline _may_ get destroyed between
compute/draw calls. As such, there is a bunch of code to check and
re-create if needed.

This Cl changes the code to no longer store pipeline_ or
pipeline_layout_ in the vulkan/pipepline object. Instead, both the
layout and pipeline are created each draw/compute call.


dj2:
This is working on my machine without validation errors. The code overall seems simpler, thoughts?

PR #319, opened 2019-02-22
Build with local vulkan data if requested

dj2:
This CL adds a flag AMBER_USE_LOCAL_VULKAN which if set when running cmake will have amber use the vulkan header and loaders out of the third_party directory. This should let amber try to build on a system without vulkan so the build can be tested on the bots.


dj2:
@dneto0 @paulthomson thoughts? I had to change the find python setting in the loader cmake file to build. Otherwise, this seems to build the libambervulkanengine on my Mac without MoltenVK.

paulthomson:
This looks great to me! Thanks for doing this. I have not yet tested on Windows, but I will do this soon.

paulthomson:
OK I think it works on Windows. I used:

```
cmake .. -G Ninja -DUSE_LOCAL_VULKAN=ON -DPYTHON_EXECUTABLE:FILEPATH=C:\python3\python.exe
cmake --build .
```

I ended up hitting the "assert" identifier not found error, but that is fixed in another PR. 

I tried a Release build and I hit some failure in building the vulkan loader: 

```
Failed to write the updated manifest to the resource of file "shaderc-online-compile.exe".
The process cannot access the file because it is being used by another process.
```

But this seems like an issue of the vulkan loader build, not amber.

jaebaek:
Sorry for being late on reviewing this .. I am testing it on Mac and Windows. In Windows, I met lots of error to run CMake. @paulthomson do you have ever met an issue related to "rc"?

jaebaek:
I think it works fine on Mac.

paulthomson:
I don't believe I got an issue mentioning "rc". This was using Windows 10, CMake 3.13, Visual Studio 2017.

dj2:
Looks like this is a python issue. We need to get all our stuff updated to python 3 and then the loader CMakefile should be happy I think.

paulthomson:
Cool! Thanks for working on this!

dj2:
I think this is working .....

kokoro/scripts/windows/build.bat
```diff
@@ -1,64 +1,66 @@
...
+@echo on
+
+set BUILD_ROOT=%cd%
+set SRC=%cd%\github\amber
+set BUILD_TYPE=%1
+
+set PATH=C:\python27;%PATH%
+
+cd %SRC%
+python tools\git-sync-deps
```
dneto0: i wonder if you can use "python2" here and not have to play games with the path.

samples/config_helper_vulkan.cc
```diff
@@ -860,9 +869,9 @@ void ConfigHelperVulkan::DumpPhysicalDeviceInfo() {
 
   std::cout << std::endl;
   std::cout << "Physical device properties:" << std::endl;
-  std::cout << "  apiVersion: " << VK_VERSION_MAJOR(api_version) << "."
-            << VK_VERSION_MINOR(api_version) << "."
-            << VK_VERSION_PATCH(api_version) << std::endl;
+  std::cout << "  apiVersion: " << static_cast<uint32_t>(api_version >> 22)
```
dneto0: awesome.

kokoro/scripts/windows/build.bat
dj2: Maybe? We have a fix for git-sync-deps in shaderc so porting that over is probably the right solution in the long run.

samples/config_helper_vulkan.cc
dj2: Yup, old style cast build warning, heh.

PR #320, opened 2019-02-26
Fix debug build

hevrard:
Fixes debug builds, which currently fails with errors like:

```
../../src/vulkan/resource.cc:227:3: error: ‘assert’ was not declared in this scope
   assert(memory_type_index);
   ^~~~~~
```

### DIFF
--- a/src/vulkan/image.cc
+++ b/src/vulkan/image.cc
@@ -14,6 +14,7 @@
 
 #include "src/vulkan/image.h"
 
+#include <cassert>
 #include <limits>
 
 #include "src/vulkan/command_buffer.h"
--- a/src/vulkan/image.cc
+++ b/src/vulkan/image.cc
@@ -14,7 +14,6 @@
 
 #include "src/vulkan/image.h"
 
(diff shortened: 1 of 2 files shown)


jaebaek:
Thank you for working on it!

hevrard:
My pleasure :)

It is surprising that such build errors got into master branch, given that the CI builds should have failed.

@jaebaek any idea how it managed to sneak in, and how to avoid this in the future?

paulthomson:
Because the CI does not build the vulkan backend. #319 is the first step to fixing this.

jaebaek:
Right, it is because CL does not build the vulkan backend.

I just realized that it would be better just removing `assert()` and replacing it to returning `Result("...")` after checking error.

i.e.,

In src/vulkan/image.cc
```
  if (!memory_type_index)
    return Result("Vulkan: Image::AllocateAndBindMemoryToVkImage memory_type_index is nullptr");
```

hevrard:
Updated and replaced assertions by Result()

PR #321, opened 2019-03-04
Add Amber::AreAllRequirementsSupported

dneto0:
It determines if the engine satisifes all requirements declared
in the recipe.

### DIFF
--- a/src/amber.cc
+++ b/src/amber.cc
@@ -68,15 +68,10 @@ Result CreateEngineAndCheckRequirements(const Recipe* recipe,
   if (!script)
     return Result("Recipe must contain a parsed script");
 
-  const auto saved_env = script->GetSpvTargetEnv();
-  auto restore_env = [&script, &saved_env]() {
-    script->SetSpvTargetEnv(saved_env);
-  };
   script->SetSpvTargetEnv(opts->spv_env);
 
   auto engine = Engine::Create(opts->engine);
   if (!engine) {
-    restore_env();
     return Result("Failed to create engine");
   }
 
@@ -85,15 +80,12 @@ Result CreateEngineAndCheckRequirements(const Recipe* recipe,
   Result r = engine->Initialize(opts->config, script->GetRequiredFeatures(),
                                 script->GetRequiredInstanceExtensions(),
                                 script->GetRequiredDeviceExtensions());
-  if (!r.IsSuccess()) {
-    restore_env();
+  if (!r.IsSuccess())
     return r;
-  }
 
   *engine_ptr = std::move(engine);
   *script_ptr = script;
 
-  restore_env();
   return r;
 }
 }  // namespace
--- a/src/amber.cc
+++ b/src/amber.cc
@@ -57,11 +57,10 @@ namespace {
 // Returns a failing result if anything fails.  Otherwise pass the created
 // engine out through |engine_ptr| and the script via |script|.  The |script|
 // pointer is borrowed, and should not be freed.
-Result CreateEngineAndCheckRequirements(
-    const Recipe* recipe,
-    Options* opts,
-    std::unique_ptr<Engine>* engine_ptr,
-    Script** script_ptr) {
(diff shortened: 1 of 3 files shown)


dneto0:
I looked at making a unit test for this, by substituting my own engine that returns a specified Result object.  Unfortunately the Amber object really wants to create the engine object, and amber::Engine is opaque to the public API.  So I stopped.  

I'd like to talk about whether injection of the Engine object in the Options object is worth the complexity in the public API.

src/amber.cc
```diff
@@ -51,16 +51,16 @@ amber::Result Amber::Parse(const std::string& input, amber::Recipe* recipe) {
...
-  return ExecuteWithShaderData(recipe, opts, map);
-}
-
-amber::Result Amber::ExecuteWithShaderData(const amber::Recipe* recipe,
-                                           Options* opts,
-                                           const ShaderMap& shader_data) {
+// Create an engine initialize it, and check the recipe's requirements.
+// Returns a failing result if anything fails.  Otherwise pass the created
+// engine out through |engine_ptr| and the script via |script|.
+static Result CreateEngineAndCheckRequirements(
```
dj2: Move this into the anonymous namespace

src/amber.cc
```diff
@@ -51,16 +51,16 @@ amber::Result Amber::Parse(const std::string& input, amber::Recipe* recipe) {
...
-  ShaderMap map;
-  return ExecuteWithShaderData(recipe, opts, map);
-}
-
-amber::Result Amber::ExecuteWithShaderData(const amber::Recipe* recipe,
-                                           Options* opts,
-                                           const ShaderMap& shader_data) {
+// Create an engine initialize it, and check the recipe's requirements.
+// Returns a failing result if anything fails.  Otherwise pass the created
+// engine out through |engine_ptr| and the script via |script|.
```
dj2: Add a:

The |script| pointer should not be freed.

or something?

src/amber.cc
```diff
@@ -72,9 +72,38 @@ amber::Result Amber::ExecuteWithShaderData(const amber::Recipe* recipe,
   if (!engine)
     return Result("Failed to create engine");
 
+  // Engine initialization checks requirements.  Current backends don't do
+  // much else.  Refactor this if they end up doing to much here.
   Result r = engine->Initialize(opts->config, script->GetRequiredFeatures(),
                                 script->GetRequiredInstanceExtensions(),
                                 script->GetRequiredDeviceExtensions());
+  if (r.IsSuccess()) {
```
dj2: We typically check the failure case:
```
if (!r.IsSuccess())
  return r;

*engine_ptr = std::move(engine);
*script_ptr = script;
return {};
```

src/amber.cc
dneto0: Better comment added.

src/amber.cc
```diff
@@ -51,39 +51,52 @@ amber::Result Amber::Parse(const std::string& input, amber::Recipe* recipe) {
...
+                                        std::unique_ptr<Engine>* engine_ptr,
+                                        Script** script_ptr) {
   if (!recipe)
     return Result("Attempting to check an invalid recipe");
 
   Script* script = static_cast<Script*>(recipe->GetImpl());
   if (!script)
     return Result("Recipe must contain a parsed script");
 
+  const auto saved_env = script->GetSpvTargetEnv();
```
dj2: What is this for? Do you expect the environment to change somehow? This seems like a complicated way of just calling script->SetSpvTargetEnv(saved_env) at the various usage sites. ...

dneto0: I wanted to avoid having to mention anything about the mutation of the recipe.
Without this, the recipe's script actually changes state, which is potentially surprising.   I will avod the saving and instead enforce setting in the usage sites.

dj2: I think we can just add a note in the comment and allow it for now. I don't think it will cause an issue at the moment.

PR #322, opened 2019-03-04
Fixup version dump script.

dj2:
This CL fixes the version dump script to correct the paths which were
missing a separator character.


dj2:
TBRing to get this fixed.

PR #325, opened 2019-03-04
Log Vulkan API calls with the new VKLOG() macro

### LINKED ISSUE #324: Log all calls in vulkan engine
We'd like to be able to log all calls in the vulkan engine so if something fails, we have a complete log of where execution terminated.

This will probably involve wrapping each vulkan call into a macro. We just have to make sure we return the vulkan result so Amber can error out as needed.

hevrard:
Fix #324  

The VKLOG() macro pseudo-code is:
`#define VKLOG(expr) ( printf(#expr) , (expr) )`

It uses the C comma operator to make sure the VKLOG(vkCall(...)) expression returns the value of the vkCall() inside it, if any (some vkCalls return void). In addition, it prints the stripped file name and line number. This macro is meaningful only if AMBER_LOG_VULKAN is defined at cmake time, otherwise it just evaluates its argument.

An example of log:

```
device.cc:450 vkGetPhysicalDeviceProperties(physical_device_, &physical_device_properties_)
device.cc:453 vkGetPhysicalDeviceMemoryProperties(physical_device_, &physical_memory_properties_)
command_pool.cc:34 vkCreateCommandPool(device_->GetDevice(), &pool_info, nullptr, &pool_)
engine_vulkan.cc:532 vkGetPhysicalDeviceFormatProperties(physical_device, format, &properties)
engine_vulkan.cc:252 vkCreateShaderModule( device_->GetDevice(), &info, nullptr, &shader)
engine_vulkan.cc:252 vkCreateShaderModule( device_->GetDevice(), &info, nullptr, &shader)
command_buffer.cc:37 vkAllocateCommandBuffers( device_->GetDevice(), &command_info, &command_)
command_buffer.cc:44 vkCreateFence(device_->GetDevice(), &fence_info, nullptr, &fence_)
graphics_pipeline.cc:419 vkCreateRenderPass(device_->GetDevice(), &render_pass_info, nullptr, &render_pass_)
image.cc:72 vkCreateImage(device_->GetDevice(), &image_info_, nullptr, &image_)
image.cc:284 vkGetImageMemoryRequirements(device_->GetDevice(), image, &requirement)
[etc]
```

The logging is currently in place for all API calls under src/vulkan/, not for API calls in /samples, we would need to make vklog.h/cc accessible in samples for this.

What do you think of the approach?

Is it relevant to log the Vulkan calls of the samples?

### DIFF
--- a/src/vulkan/CMakeLists.txt
+++ b/src/vulkan/CMakeLists.txt
@@ -32,6 +32,14 @@ set(VULKAN_ENGINE_SOURCES
     vertex_buffer.cc
 )
 
+if (${AMBER_LOG_VULKAN})
+  set(VULKAN_ENGINE_SOURCES
+    ${VULKAN_ENGINE_SOURCES}
+    vklog.cc
+    )
+  add_definitions(-DAMBER_LOG_VULKAN)
+endif()
+
 add_library(libamberenginevulkan ${VULKAN_ENGINE_SOURCES})
 amber_default_compile_options(libamberenginevulkan)
 set_target_properties(libamberenginevulkan PROPERTIES
--- a/src/vulkan/buffer.cc
+++ b/src/vulkan/buffer.cc
@@ -16,6 +16,7 @@
 
 #include "src/vulkan/command_buffer.h"
 #include "src/vulkan/device.h"
+#include "src/vulkan/vklog.h"
 
 namespace amber {
 namespace vulkan {
@@ -74,9 +75,9 @@ Result Buffer::CreateVkBufferView(VkFormat format) {
   buffer_view_info.format = format;
   buffer_view_info.offset = 0;
   buffer_view_info.range = VK_WHOLE_SIZE;
-  if (device_->GetPtrs()->vkCreateBufferView(device_->GetDevice(),
-                                             &buffer_view_info, nullptr,
-                                             &view_) != VK_SUCCESS) {
+  if (VKLOG(device_->GetPtrs()->vkCreateBufferView(
+          device_->GetDevice(), &buffer_view_info, nullptr, &view_)) !=
+      VK_SUCCESS) {
     return Result("Vulkan::Calling vkCreateBufferView Fail");
   }
 
@@ -103,24 +104,26 @@ void Buffer::CopyFromBuffer(CommandBuffer* command, const Buffer& src) {
   region.dstOffset = 0;
   region.size = src.GetSizeInBytes();
 
-  device_->GetPtrs()->vkCmdCopyBuffer(command->GetCommandBuffer(), src.buffer_,
-                                      buffer_, 1, &region);
+  VKLOG(device_->GetPtrs()->vkCmdCopyBuffer(command->GetCommandBuffer(),
+                                            src.buffer_, buffer_, 1, &region));
   MemoryBarrier(command);
 }
 
 void Buffer::Shutdown() {
   if (view_ != VK_NULL_HANDLE) {
-    device_->GetPtrs()->vkDestroyBufferView(device_->GetDevice(), view_,
-                                            nullptr);
+    VKLOG(device_->GetPtrs()->vkDestroyBufferView(device_->GetDevice(), view_,
+                                                  nullptr));
   }
 
   if (memory_ != VK_NULL_HANDLE) {
     UnMapMemory(memory_);
-    device_->GetPtrs()->vkFreeMemory(device_->GetDevice(), memory_, nullptr);
+    VKLOG(device_->GetPtrs()->vkFreeMemory(device_->GetDevice(), memory_,
+                                           nullptr));
   }
 
   if (buffer_ != VK_NULL_HANDLE)
-    device_->GetPtrs()->vkDestroyBuffer(device_->GetDevice(), buffer_, nullptr);
+    VKLOG(device_->GetPtrs()->vkDestroyBuffer(device_->GetDevice(), buffer_,
+                                              nullptr));
 
   Resource::Shutdown();
 }
--- a/src/vulkan/command_buffer.cc
+++ b/src/vulkan/command_buffer.cc
@@ -16,6 +16,7 @@
 
 #include "src/vulkan/command_pool.h"
 #include "src/vulkan/device.h"
+#include "src/vulkan/vklog.h"
 
 namespace amber {
 namespace vulkan {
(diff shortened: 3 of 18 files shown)


src/vulkan/buffer.cc
```diff
@@ -103,24 +104,26 @@ void Buffer::CopyFromBuffer(CommandBuffer* command, const Buffer& src) {
...
   }
 
   if (memory_ != VK_NULL_HANDLE) {
     UnMapMemory(memory_);
-    device_->GetPtrs()->vkFreeMemory(device_->GetDevice(), memory_, nullptr);
+    VKLOG(device_->GetPtrs()->vkFreeMemory(device_->GetDevice(), memory_,
+                                           nullptr));
   }
 
   if (buffer_ != VK_NULL_HANDLE)
```
dj2: nit: add {}'s

Macros in if's without {}'s scare me ....

src/vulkan/command_buffer.cc
```diff
@@ -81,29 +82,30 @@ Result CommandBuffer::SubmitAndReset(uint32_t timeout_ms) {
...
-      static_cast<uint64_t>(timeout_ms) * 1000ULL * 1000ULL /* nanosecond */);
+      static_cast<uint64_t>(timeout_ms) * 1000ULL * 1000ULL /* nanosecond */));
   if (r == VK_TIMEOUT)
     return Result("Vulkan::Calling vkWaitForFences Timeout");
   if (r != VK_SUCCESS)
     return Result("Vulkan::Calling vkWaitForFences Fail");
 
-  if (device_->GetPtrs()->vkResetCommandBuffer(command_, 0) != VK_SUCCESS)
+  if (VKLOG(device_->GetPtrs()->vkResetCommandBuffer(command_, 0)) !=
+      VK_SUCCESS)
```
dj2: nit: add {}s

src/vulkan/engine_vulkan.cc
```diff
@@ -143,7 +144,8 @@ Result EngineVulkan::Shutdown() {
   for (auto it = modules_.begin(); it != modules_.end(); ++it) {
     auto vk_device = device_->GetDevice();
     if (vk_device != VK_NULL_HANDLE && it->second != VK_NULL_HANDLE)
```
dj2: nit {}s

src/vulkan/image.cc
```diff
@@ -122,15 +124,17 @@ Result Image::CreateVkImageView() {
 
 void Image::Shutdown() {
   if (view_ != VK_NULL_HANDLE) {
-    device_->GetPtrs()->vkDestroyImageView(device_->GetDevice(), view_,
-                                           nullptr);
+    VKLOG(device_->GetPtrs()->vkDestroyImageView(device_->GetDevice(), view_,
+                                                 nullptr));
   }
 
   if (image_ != VK_NULL_HANDLE)
```
dj2: {}s

src/vulkan/image.cc
```diff
@@ -122,15 +124,17 @@ Result Image::CreateVkImageView() {
...
-    device_->GetPtrs()->vkDestroyImage(device_->GetDevice(), image_, nullptr);
+    VKLOG(device_->GetPtrs()->vkDestroyImage(device_->GetDevice(), image_,
+                                             nullptr));
 
   if (memory_ != VK_NULL_HANDLE)
```
dj2: {}s

src/vulkan/vklog.cc
```diff
@@ -0,0 +1,60 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "src/vulkan/vklog.h"
+
+// Use C string library for preprocessor strings like __FILE__
+#include <string.h>
```
dj2: #include &lt;cstring> ?

dj2:
I don't think the samples calls are that interesting in this case as it's just setting up the engine? If you don't need it, let's skip it for now, we can always add later.

src/vulkan/CMakeLists.txt
```diff
@@ -32,6 +32,14 @@ set(VULKAN_ENGINE_SOURCES
     vertex_buffer.cc
 )
 
+if (${AMBER_LOG_VULKAN})
```
dj2: I think we want to add an option(AMBER_LOG_VULKAN ...) into the top level CMakeLists.txt file (similar to AMBER_SKIP_SAMPLES etc)

dj2:
I was trying to think if we could do this in a less invasive way by creating wrapper functions in device.cc when we generate all the function pointer calls but I can't think of any good way to get __FILE__ and __LINE__ to be correct in that case.

hevrard:
> I was trying to think if we could do this in a less invasive way by creating wrapper functions in device.cc when we generate all the function pointer calls but I can't think of any good way to get **FILE** and **LINE** to be correct in that case.

Well, FILE and LINE are not strictly required. Were you thinking of a tweak of AMBER_VK_FUNC in device.cc?

This came to my mind also, but I'm not certain about how to achieve this. For each API primitive, we basically want a function with the same signature that would just print the function name before calling it (having a print of the result value before returning would be nice, but we can do without it). Right now the macro just assigns the function returned by getInstancePointer(), is it possible to wrap this in a lambda that adds the logging without having to define the signature of the API primitive being wrapped?

Maybe with some variadic lambda / variadic template magic, which I am very not familiar with?

dj2:
So, we have another possible use for this type of logging, we'd like to be able to add timing information for calls in the log as well. This would let us do things like wrap the dispatch command for a compute and output how long it command took.

I think that will require us to go the route of replacing the function pointers, I haven't found a nice way to do that yet. As far as I can see (in c++11) we'll probably have to write out each of the lambda's we want by hand device.cc. I can't see a way to generate it from a macro at this point.

Thoughts on a better way? I can look at updating the device code to use the lambda wrappers tomorrow if you want, then you can add your logging code in afterwards. Or, feel free to poke it yourself.

dneto0:
I don't have anything to add.  :-)

dj2:
I'm also thinking now, we want this as a runtime flag, not compile time. I'm guessing you want to be able to run the thing, and if it hangs, run it again in the slower logging mode? Probably the same with getting timing information, run the slower mode after you know it works.

hevrard:
Yes! I didn't dare going all the way to generate functions from the header def, but this is great as it enables both logging and timing (which is quite interesting for us too), and whatever we need in the future. Also, overloading the function pointers directly avoids wrapping calls in a macro, which is error-prone. So I close this PR and will comment on #328 for other ideas.

PR #328, opened 2019-03-05
[vulkan] Generate wrapper functions for vulkan methods.

dj2:
This CL adds a update_vk_wrappers.py script which is executed during the
build. The script will generate src/vk-wrappers.h and
src/vk-wrappers.inc files into the binary directory. The .h file lists
the struct entries for the VulkanPtr struct in device.h.

The vk-wrappers.inc file contains the code to load the vulkan function
pointers. This file is included into device.cc.

If the vk.xml file is found, the wrappers will be lambdas and the struct
entries will be std::function's. If vk.xml is not found the struct entries are
the PFN_* methods themselves. This should allow us to work with the
CTS as we do now as the vk.xml file is not available during CTS build.

The wrapper code will either just setup the function pointer or it will
generate a lambda which captures the function pointer. We can then
update the lambda to do things like logging or timing as required.

Issue #324.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ third_party/lodepng
 third_party/shaderc
 third_party/spirv-tools
 third_party/spirv-headers
+third_party/vulkan-headers
 .vs
 
 # Vim swap files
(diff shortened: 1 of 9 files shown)


hevrard:
I like this approach!

With the idea of a runtime flag to enable between "slow logging-timing mode", we can think of passing such a flag to Device::LoadVulkanPointers(), which would load either the plain function (in regular mode) or the logging-lambda (in slow mode). Thus, we can have a run-time flag that leads to virtually zero overhead when the regular mode is used (just a conditional at API function loading time, not at API function call time). How does this sound?

dj2:
Filed #330 to track adding the runtime switch for function pointer vs lambda.

PR #331, opened 2019-03-05
Allow multiple pipelines to execute.

dj2:
This CL updates amber (and the vulkan backend) to work with multiple
pipelines. The vulkan backend is changed to store a PipelineData
structure with the colour/depth formats, vertex data and the vulkan
pipeline for a given amber::Pipeline.

Each command is changed to take the pipeline it is associated with,
those pipelines are used when executing the commands.


src/vulkan/engine_vulkan.cc
```diff
@@ -159,108 +148,135 @@ Result EngineVulkan::Shutdown() {
...
-                         info.buffer->GetData());
+    Result r = SetBuffer(
+        pipeline, info.type, static_cast<uint8_t>(info.location),
+        info.buffer->AsFormatBuffer()->GetFormat(), info.buffer->GetData());
     if (!r.IsSuccess())
       return r;
   }
 
   if (pipeline->GetDepthBuffer().buffer) {
     const auto& info = pipeline->GetDepthBuffer();
```
jaebaek: This is confusing. Better not using `info` because we used it in `auto& info = pipeline_map_[pipeline];`

src/vulkan/engine_vulkan.cc
```diff
@@ -159,108 +148,135 @@ Result EngineVulkan::Shutdown() {
...
+  info.depth_frame_format = MakeUnique<Format>();
+  info.depth_frame_format->SetFormatType(FormatType::kUnknown);
 
   // Handle Image and Depth buffers early so they are available when we create
   // the pipeline.
   for (const auto& info : pipeline->GetColorAttachments()) {
-    Result r = SetBuffer(info.type, static_cast<uint8_t>(info.location),
-                         info.buffer->AsFormatBuffer()->GetFormat(),
-                         info.buffer->GetData());
+    Result r = SetBuffer(
```
jaebaek: This is confusing. Better not using `info` because we used it in `auto& info = pipeline_map_[pipeline];`

src/vulkan/engine_vulkan.cc
```diff
@@ -159,108 +148,135 @@ Result EngineVulkan::Shutdown() {
...
 
+  info.vk_pipeline = std::move(vk_pipeline);
+
   for (const auto& info : pipeline->GetVertexBuffers()) {
-    Result r = SetBuffer(info.type, static_cast<uint8_t>(info.location),
-                         info.buffer->IsFormatBuffer()
-                             ? info.buffer->AsFormatBuffer()->GetFormat()
-                             : Format(),
-                         info.buffer->GetData());
+    Result r =
```
jaebaek: Same here

src/vulkan/engine_vulkan.cc
dj2: I have a followup CL which fixes all of these.

PR #333, opened 2019-03-06
Fix and tweak build instructions in README

paulthomson:
Minor fix and tweaks to the README. Emphasize up-front that a newer version of CMake is needed for automatic discovery of the Vulkan SDK. Use `https://` for the git clone so a GitHub key is not required. Add additional commands needed for an Android NDK build.


dj2:
Awesome, thanks.

PR #334, opened 2019-03-06
Add Delegate class to enable hooking log function around graphics API calls

hevrard:
This adds a Delegate which enables to define hook functions that will
be called around each graphics API calls.

In practice, a pointer to the delegate object is passed at engine
creation and the Vulkan engine eventually pass it to
LoadVulkanPointers() which calls the functions produced by
update_vk_wrappers.py. This enable to choose at function load time
whether to load a straighforward wrapper to the API command, or to
load a lambda that surround the API command with calls to the delegate
methods.

The delegate currently offers only a log() method, called before the
actual API command, but other delegate methods may be added later on.

The --verbose flag enables the delegate to produce a log of API calls.

### DIFF
--- a/include/amber/amber.h
+++ b/include/amber/amber.h
@@ -52,6 +52,17 @@ struct BufferInfo {
   std::vector<Value> values;
 };
 
+/// Delegate class for various hook functions
+class Delegate {
+ public:
+  virtual ~Delegate();
+
+  /// Log the given message
+  virtual void Log(const std::string& message) = 0;
+  /// Tells whether to log the graphics API calls
+  virtual bool LogGraphicsCalls() const = 0;
+};
+
 struct Options {
   /// Sets the engine to be created. Default Vulkan.
   EngineType engine;
@@ -63,6 +74,8 @@ struct Options {
   std::vector<BufferInfo> extractions;
   /// Terminate after creating the pipelines.
   bool pipeline_create_only;
+  /// Delegate implementation
+  Delegate* delegate;
 };
 
 /// Main interface to the Amber environment.
--- a/samples/amber.cc
+++ b/samples/amber.cc
@@ -45,6 +45,7 @@ struct Options {
   bool quiet = false;
   bool show_help = false;
   bool show_version_info = false;
+  bool log_graphics_calls = false;
   amber::EngineType engine = amber::kEngineTypeVulkan;
   std::string spv_env;
 };
@@ -64,6 +65,7 @@ const char kUsage[] = R"(Usage: amber [options] SCRIPT [SCRIPTS...]
   -e <engine>               -- Specify graphics engine: vulkan, dawn. Default is vulkan.
   -v <engine version>       -- Engine version (eg, 1.1 for Vulkan). Default 1.0.
   -V, --version             -- Output version information for Amber and libraries.
+  --log-graphics-calls      -- Log graphics API calls (only for Vulkan so far).
   -h                        -- This help text.
 )";
 
@@ -157,6 +159,8 @@ bool ParseArgs(const std::vector<std::string>& args, Options* opts) {
       opts->quiet = false;
     } else if (arg == "-q") {
       opts->quiet = true;
+    } else if (arg == "--log-graphics-calls") {
+      opts->log_graphics_calls = true;
     } else if (arg.size() > 0 && arg[0] == '-') {
       std::cerr << "Unrecognized option " << arg << std::endl;
       return false;
@@ -202,6 +206,25 @@ std::string ReadFile(const std::string& input_file) {
 
   return std::string(data.begin(), data.end());
 }
+
+class SampleDelegate : public amber::Delegate {
+ public:
+  SampleDelegate() : log_graphics_calls_(false) {}
+  ~SampleDelegate() override = default;
+
+  void Log(const std::string& message) override {
+    std::cout << message << std::endl;
+  }
+
+  bool LogGraphicsCalls() const override { return log_graphics_calls_; }
+
+  void SetLogGraphicsCalls(bool log_graphics_calls) {
+    log_graphics_calls_ = log_graphics_calls;
+  }
+
+ private:
+  bool log_graphics_calls_;
+};
 
 }  // namespace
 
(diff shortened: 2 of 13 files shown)


include/amber/amber.h
```diff
@@ -52,6 +52,12 @@ struct BufferInfo {
   std::vector<Value> values;
 };
 
+/// Delegate class to hook functions around graphics API commands
+class Delegate {
+ public:
+  virtual void log(std::string command) = 0;
```
dj2: const std::string& command

dj2: const std::string& command

samples/amber.cc
```diff
@@ -203,6 +207,13 @@ std::string ReadFile(const std::string& input_file) {
   return std::string(data.begin(), data.end());
 }
 
+class Delegate : public amber::Delegate {
+ public:
+  void log(std::string command) {
+    std::cout << "Graphics API call: " << command << std::endl;
```
dj2: Do we need "Graphics API call: " in here or can we just log the command?

samples/amber.cc
```diff
@@ -203,6 +207,13 @@ std::string ReadFile(const std::string& input_file) {
   return std::string(data.begin(), data.end());
 }
 
+class Delegate : public amber::Delegate {
+ public:
+  void log(std::string command) {
```
dj2: void log(const std::string& command) override

tools/update_vk_wrappers.py
```diff
@@ -86,20 +87,43 @@ def gen_wrappers(methods, xml):
     content += "    return Result(\"Vulkan: Unable to "
     content += "load {} pointer\");\n".format(method)
     content += "  }\n"
-    content += "  ptrs_.{} = [ptr](".format(method)
+
+    # if delegate is not null ...
+    content += "  if (delegate != nullptr) {\n"
```
dj2: What if we make this:
```
if (delegate && delegate->LogGraphicsCalls()) {
```

Add that second method to the delegate, then we add that second method to the delegate and always set it in the sample app? We're possibly going to want delegate for things like loading files so we don't want to just use the existence here. Having the method saves plumbing the flag all the way here.

src/vulkan/device.cc
```diff
@@ -354,20 +355,22 @@ Device::Device(VkInstance instance,
 
 Device::~Device() = default;
 
-Result Device::LoadVulkanPointers(
-    PFN_vkGetInstanceProcAddr getInstanceProcAddr) {
+Result Device::LoadVulkanPointers(PFN_vkGetInstanceProcAddr getInstanceProcAddr,
+                                  Delegate* delegate) {
+  (void)delegate;  // work around -Werror=unused-parameter
```
dj2: What about if we stick right here something like:
```
if (delegate && delegate->LogGraphicsCalls())
  delegate->Log("Loading Vulkan Pointers");
```
Then the delegate is used, and we get the context of these calls as well?

samples/amber.cc
```diff
@@ -64,6 +65,7 @@ const char kUsage[] = R"(Usage: amber [options] SCRIPT [SCRIPTS...]
   -e <engine>               -- Specify graphics engine: vulkan, dawn. Default is vulkan.
   -v <engine version>       -- Engine version (eg, 1.1 for Vulkan). Default 1.0.
   -V, --version             -- Output version information for Amber and libraries.
+  --verbose                 -- Log graphics API calls made by the Vulkan engine.
```
dj2: --log-graphics-calls ? Which we then set into the delegate directly and return with the LogGraphicsCalls method?

include/amber/amber.h
```diff
@@ -52,6 +52,12 @@ struct BufferInfo {
...
+};
```
dj2: See below, I think we should add a:
```
  bool LogGraphicsCalls() const { ... }
```

And a setting so the sample app can set that value into the delegate. (Which also means we need a virtual destructor as well)

samples/amber.cc
hevrard: This was to convince myself that the log I saw was from the delegate function, let's just print the command indeed.

hevrard:
Many thanks for the comments, I tried to implement them all, but I may not have got the virtual destructors right—bare with me as I learn those C++ subtleties.

include/amber/amber.h
```diff
@@ -52,6 +52,17 @@ struct BufferInfo {
   std::vector<Value> values;
 };
 
+/// Delegate class for various hook functions
+class Delegate {
+ public:
+  virtual ~Delegate() = default;
```
dj2: This will fail on the bots due to a warning being treated as error. You need to put the definition into the src/amber.cc file. So just virtual ~Delegate(); here and later Just Delegate::~Delegate() = default; in the .cc

samples/Android.mk
```diff
@@ -14,6 +14,8 @@
 
 LOCAL_PATH:= $(call my-dir)
 
+cmd-strip :=
```
dj2: This seems unrelated?

samples/amber.cc
```diff
@@ -268,6 +291,13 @@ int main(int argc, const char** argv) {
   amber_options.engine = options.engine;
   amber_options.spv_env = options.spv_env;
   amber_options.pipeline_create_only = options.pipeline_create_only;
+  amber_options.delegate = nullptr;
+
+  Delegate delegate;
```
dj2: Move the delegate stuff up to 289, then you don't have to set the nullptr into amber_options.delegate you can just set the delegate

samples/Android.mk
hevrard: Woops!

include/amber/amber.h
hevrard: I tried what you said before, but it leads to the following error at link time (Linux gcc):

`samples/CMakeFiles/amber.dir/amber.cc.o:(.data.rel.ro+0x10): undefined reference to `typeinfo for amber::Delegate'`

Chatting with Ally and Ben, it turned out the issue was that `-fno-rtti` is used for all library code, but not the sample.

See the change on the sample CMakeLists.txt which fixes this.

hevrard:
In addition to the review fixes, I renamed the Delegate class in samples/ to be SampleDelegate, for the sake of clarity. I also added a note to clarify why logging via a Vulkan layer is not an option for us, as in the future one may wonder why we didn't rely on a layer.

PR #335, opened 2019-03-06
[amberscript] Implement RUN for compute and DRAW_RECT.

dj2:
This CL adds parsing of the RUN commands for compute and DRAW_RECT
instances to AmberScript.

### DIFF
--- a/src/amberscript/parser.cc
+++ b/src/amberscript/parser.cc
@@ -956,11 +956,9 @@ Result Parser::ParseRun() {
 
     script_->AddCommand(std::move(cmd));
     return ValidateEndOfStatement("RUN command");
-  } else {
-    return Result("invalid token in RUN command: " + token->AsString());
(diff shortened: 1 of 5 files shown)


src/amberscript/parser.cc
```diff
@@ -846,5 +848,118 @@ Result Parser::ParseBufferInitializerData(DataBuffer* buffer) {
...
+    r = token->ConvertToDouble();
+    if (!r.IsSuccess())
+      return r;
+    cmd->SetHeight(token->AsFloat());
+
+    script_->AddCommand(std::move(cmd));
+    return ValidateEndOfStatement("RUN command");
+  }
+
+  if (token->AsString() == "DRAW_ARRAY") {
```
dneto0: This goes a little beyond what's in the commit message, but I'll allow it. :-)

PR #338, opened 2019-03-07
[vulkan] fix shadow variable

dj2:


dj2:
This is breaking my local vulkan. TBR'ing to see if I can get that CL to green.

PR #351, opened 2019-03-11
WIP: Add --log-graphics-call-time flag

### LINKED ISSUE #342: Log timing information if requested
Now that we have logging of all the vulkan calls if desired, we should add in a flag so you can get timing information for each vulkan call if desired.

This should probably be another flag as it adds overhead, but logging timing would imply logging in general.

hevrard:
Fix #342 

Provide a timestamp in nanoseconds in front of each graphics API call log.

This is a work-in-progress draft to discuss the following:
- Should GetTimestamp() be implemented in samples/ (like it is in this draft), or is it OK to put the implementation under src/ ?
- Currently we prefix each API call with a timestamp, was the idea to actually return the API call *duration*?
- Are you OK with the refactoring of vulkan wrapper generation using a python template string? (I can extract this as a preliminary, isolated PR if you prefer?)

### DIFF
--- a/include/amber/amber.h
+++ b/include/amber/amber.h
@@ -61,6 +61,10 @@ class Delegate {
   virtual void Log(const std::string& message) = 0;
   /// Tells whether to log the graphics API calls
   virtual bool LogGraphicsCalls() const = 0;
+  /// Tells whether to log the duration of graphics API calls
+  virtual bool LogGraphicsCallsTime() const = 0;
+  /// Returns the current timestamp in nanoseconds
+  virtual uint64_t GetTimestampNs() const = 0;
 };
 
 struct Options {
--- a/samples/timestamp.cc
+++ b/samples/timestamp.cc
@@ -1,0 +1,69 @@
+// Copyright 2019 The Amber Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "samples/timestamp.h"
+
+#include <cassert>
+
+#if defined(_WIN32) || defined(_WIN64)
+#define SAMPLE_PLATFORM_WINDOWS 1
+#define SAMPLE_PLATFORM_POSIX 0
+#elif defined(__linux__) || defined(__APPLE__)
+#define SAMPLE_PLATFORM_POSIX 1
+#define SAMPLE_PLATFORM_WINDOWS 0
+#endif
+
+#if SAMPLE_PLATFORM_WINDOWS
+#include <windows.h>
+#elif SAMPLE_PLATFORM_POSIX
+#include <time.h>
+#else
+#error "Unknown platform"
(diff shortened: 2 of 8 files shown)


dj2:
* For GetTimestamp, it's up to you. Having it in samples/ is nice because a client may have their own clock they want the numbers against, so it's more flexible.

 * I think _duration_ is what we really want. Amber isn't designed to be fast, so the start times won't give you a lot of useful information, I think.

 * Yes please. Extract in a separate Cl and we'll land that first. It was on my list, but my python skills are terrible, heh.

include/amber/amber.h
```diff
@@ -61,6 +61,10 @@ class Delegate {
   virtual void Log(const std::string& message) = 0;
   /// Tells whether to log the graphics API calls
   virtual bool LogGraphicsCalls() const = 0;
+  /// Tells whether to log the time at graphics API calls
+  virtual bool LogGraphicsCallsTime() const = 0;
+  /// Returns the current timestamp
```
dj2: Returns the current timestamp in nanoseconds

samples/timestamp.cc
```diff
@@ -0,0 +1,42 @@
...
+#elif AMBER_PLATFORM_POSIX
+#include <time.h>
+#endif
+
+namespace timestamp {
+
+long SampleGetTimestamp() {
+  long timestamp = 0L;
+#if AMBER_PLATFORM_WINDOWS
+  assert(false && "TODO: implement GetTimestamp() on Windows");
```
dj2: https://pdfium.googlesource.com/pdfium/+/refs/heads/master/core/fxcrt/fx_random.cpp#55  is what you want I think. Looks like there is no include required, just use GetSystemTime here

src/sleep.cc
```diff
@@ -14,7 +14,7 @@
 
 #include "src/sleep.h"
 
-#include "src/platform.h"
+#include "amber/platform.h"
```
dj2: I don't think we want to push this into the public API. I'd be more inclined to duplicate it into samples/ if that's where you need it.

tools/update_vk_wrappers.py
```diff
@@ -80,51 +80,49 @@ def gen_wrappers(methods, xml):
...
+    arguments = ', '.join(str(x) for x in param_names)
+    return_type = data['return_type']
+    return_variable = ''
+    call_prefix = ''
+    if return_type != 'void':
+      return_variable = 'ret'
+      call_prefix = return_type + ' ' + return_variable + ' = '
+
+    template = Template(R'''{
+  PFN_${method} ptr = reinterpret_cast<PFN_${method}>(getInstanceProcAddr(instance_, "${method}"));
```
dj2: This is so much better, thank you.

src/sleep.cc
hevrard: Agreed.

samples/timestamp.cc
hevrard: I think we might prefer QueryPerformanceCounter (see https://docs.microsoft.com/en-us/windows/desktop/SysInfo/acquiring-high-resolution-time-stamps) , as GetSystemTime precision is milliseconds at best, and it is affected by change of system time (e.g. summer time, etc).

samples/timestamp.cc
```diff
@@ -0,0 +1,69 @@
...
+#define SAMPLE_PLATFORM_POSIX 1
+#define SAMPLE_PLATFORM_WINDOWS 0
+#endif
+
+#if SAMPLE_PLATFORM_WINDOWS
+#include <windows.h>
+#elif SAMPLE_PLATFORM_POSIX
+#include <time.h>
+#else
+#error "Unknow platform"
```
dj2: Unknown

PR #352, opened 2019-03-11
 Add buffer to probe command.

dj2:
This CL extends the ProbeCommand to contain the buffer to probe against.
With VkScript there is a generated image buffer which is used. For
AmberScript the buffer is provided in the EXPECT command.


dj2:
Rebasing on top of buf_type cl .... will re-open

PR #353, opened 2019-03-11
[vulkan] Set correct colour attachment locations.

dj2:
This CL fixes the colour attachment code to set the colour attachment
into the requested location, not the order they were attached. This
allows setting the locations out of order.

Note, the locations must be a complete set of numbers starting from 0.

### DIFF
--- a/src/vulkan/frame_buffer.cc
+++ b/src/vulkan/frame_buffer.cc
@@ -51,7 +51,7 @@ Result FrameBuffer::Initialize(
       if (info->location >= color_attachments_.size())
         return Result("color attachment locations must be sequential from 0");
       if (seen_idx[info->location] != -1) {
-        return Result("can not use duplicate attachment location: " +
+        return Result("duplicate attachment location: " +
                       std::to_string(info->location));
       }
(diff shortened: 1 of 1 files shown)


src/vulkan/frame_buffer.cc
```diff
@@ -44,18 +45,31 @@ Result FrameBuffer::Initialize(
...
+    uint32_t max_attachment_idx = 0;
+    for (auto* info : color_attachments_)
+      max_attachment_idx = std::max(max_attachment_idx, info->location);
 
-    Result r = color_images_.back()->Initialize(
-        VK_IMAGE_USAGE_TRANSFER_SRC_BIT | VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT);
-    if (!r.IsSuccess())
-      return r;
+    if ((max_attachment_idx + 1) != color_attachments_.size())
+      return Result("color attachment locations must be sequential from 0");
```
hevrard: This does not check whether all color attachment locations are distinct, you may want to check this as well?

e.g., three color attachments with locations 0, 2, 2 will pass this test.

dj2: done.

PR #354, opened 2019-03-12
Use template for Vulkan wrappers generation

hevrard:


hevrard:
As discussed in #351

dj2:
Awesome. Thank you very much for this. It's readable now, heh.

PR #355, opened 2019-03-12
Correctly store BufferType in buffer.

dj2:
Previously the buffer type existed in both the Buffer and
Pipeline::BufferInfo. These buffer types could diverge, which could
cause issues with thinking the buffer is of an incorrect type.

This CL removes the buffer type from BufferInfo and always uses the
buffer to retrieve the type.

### DIFF
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -218,8 +218,9 @@ Result Pipeline::AddColorAttachment(Buffer* buf, uint32_t location) {
 Result Pipeline::SetDepthBuffer(Buffer* buf) {
   if (depth_buffer_.buffer != nullptr)
     return Result("can only bind one depth buffer in a PIPELINE");
+  if (buf->GetBufferType() != BufferType::kDepth)
+    return Result("expected a depth buffer");
 
-  depth_buffer_.type = BufferType::kDepth;
   depth_buffer_.buffer = buf;
   depth_buffer_.width = fb_width_;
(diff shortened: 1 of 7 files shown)


src/pipeline.cc
```diff
@@ -218,8 +218,9 @@ Result Pipeline::AddColorAttachment(Buffer* buf, uint32_t location) {
 Result Pipeline::SetDepthBuffer(Buffer* buf) {
   if (depth_buffer_.buffer != nullptr)
     return Result("can only bind one depth buffer in a PIPELINE");
+  if (buf->GetBufferType() != BufferType::kDepth)
+    return Result("expected a depth buffer");
```
jaebaek: Why don't we give more debugging information e.g., buffer name?

src/pipeline.cc
```diff
@@ -244,10 +245,11 @@ Result Pipeline::AddVertexBuffer(Buffer* buf, uint32_t location) {
     if (vtex.buffer == buf)
       return Result("vertex buffer may only be bound to a PIPELINE once");
   }
+  if (buf->GetBufferType() != BufferType::kVertex)
+    return Result("expected a vertex buffer");
```
jaebaek: ditto

src/pipeline.cc
dj2: I believe in both of these cases it should have the correct line number. I'll try to write a test for this later.

PR #358, opened 2019-03-13
Remove GetDescriptorInfo and store in Buffer object.

dj2:
This Cl stores the relevant data into the amber::Buffer and removes the
need to call GetDescriptorInfo to retrieve the data.


src/vkscript/command_parser.cc
```diff
@@ -686,6 +705,22 @@ Result CommandParser::ProcessUniform() {
...
     }
 
+    {
+      // Generate an internal buffer for this binding if needed.
+      auto set = cmd->GetDescriptorSet();
+      auto binding = cmd->GetBinding();
+
+      auto* buffer = pipeline_->GetBufferForBinding(set, binding);
+      if (!buffer) {
+        auto b = MakeUnique<Buffer>(BufferType::kStorage);
```
jaebaek: Change `BufferType::kStorage` to `BufferType::kUniform`

dj2: Good catch, thanks.

PR #366, opened 2019-03-13
 [vulkan] remove Shutdown methods.

dj2:
This CL moves Shutdown methods to destructors where possible.


src/vulkan/graphics_pipeline.cc
```diff
@@ -930,18 +930,15 @@ Result GraphicsPipeline::Draw(const DrawArraysCommand* command,
...
 
 void GraphicsPipeline::Shutdown() {
   DeactivateRenderPassIfNeeded();
 
-  if (index_buffer_)
-    index_buffer_->Shutdown();
-
+  index_buffer_ = nullptr;
   Pipeline::Shutdown();
-  frame_->Shutdown();
```
hevrard: c++ learner question: here you transform `frame_->Shutdown()` to `frame_ = nullptr`, does this force the call fo frame_ destructor?

dj2: Yea, frame_ is a std::unique_ptr<FrameBuffer> by setting the unique_ptr to null it will destroy the FrameBuffer object that it currently has and store the nullptr.

Note, this wouldn't work if frame_ wasn't a unique_ptr.

Issue #345, opened 2019-03-08
Parse EXPECT commands


dj2:

dj2:
https://github.com/google/amber/pull/411 has most of the EXPECT commands listed. Just EQ_FLOAT and NE_FLOAT are missing.

dj2:
#444 added EXPECT EQ_BUFFER.

dj2:
I've removed the EQ_FLOAT and NE_FLOAT for now as I'm not sure we need them. Closing this as complete.

PR #367, opened 2019-03-14
[AmberScript] Implement EXPECT EQ_RGB(A)? parsing

dj2:
This CL adds the EXPECT parsing for EQ_RGB and EQ_RGBA expectations.
This is enough to allow a clear.amber script to be written.

Issue #345.

### DIFF
--- a/src/amberscript/parser.cc
+++ b/src/amberscript/parser.cc
@@ -1008,8 +1008,8 @@ Result Parser::ParseExpect() {
   if (token->IsEOS() || token->IsEOL())
     return Result("missing BUFFER in EXPECT command");
   if (!token->IsString() || token->AsString() != "BUFFER") {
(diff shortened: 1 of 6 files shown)


src/amberscript/parser.cc
```diff
@@ -989,5 +993,119 @@ Result Parser::ParseClear() {
...
+      if (info.buffer == buffer) {
+        found = true;
+        break;
+      }
+    }
+    if (!found)
+      return Result("buffer not in pipeline for EXPECT command");
+
+    auto probe = MakeUnique<ProbeCommand>(pipeline, buffer);
+    probe->SetX(x);
```
dneto0: MSVC is complaining about possible loss of data on the conversion from uint32_t to float.
Here and later on.

PR #371, opened 2019-03-15
Implement AmberScript COPY command

hevrard:
I am not sure about the proper implementation in executor, in particular, how should we check that the destination buffer is clean? how should we properly clean it?

### DIFF
--- a/src/amberscript/parser.cc
+++ b/src/amberscript/parser.cc
@@ -76,6 +76,8 @@ Result Parser::Parse(const std::string& data) {
       r = ParseBuffer();
     } else if (tok == "CLEAR") {
       r = ParseClear();
+    } else if (tok == "COPY") {
+      r = ParseCopy();
     } else if (tok == "EXPECT") {
       r = ParseExpect();
     } else if (tok == "PIPELINE") {
@@ -91,7 +93,7 @@ Result Parser::Parse(const std::string& data) {
       return Result(make_error(r.Error()));
   }
 
-  // Generate any  needed color and depth attachments. This is done before
+  // Generate any needed color and depth attachments. This is done before
   // validating in case one of the pipelines specifies the framebuffer size
   // it needs to be verified against all other pipelines.
   for (const auto& pipeline : script_->GetPipelines()) {
@@ -1176,6 +1178,61 @@ Result Parser::ParseExpect() {
   }
 
   return ValidateEndOfStatement("EXPECT command");
+}
+
+Result Parser::ParseCopy() {
+  auto token = tokenizer_->NextToken();
+  if (token->IsEOL() || token->IsEOS())
+    return Result("missing buffer name after COPY");
+  if (!token->IsString())
+    return Result("invalid buffer name after COPY");
+
+  auto name = token->AsString();
+  if (name == "TO")
+    return Result("missing buffer name between COPY and TO");
+
+  Buffer* buffer_from = script_->GetBuffer(name);
+  if (!buffer_from)
+    return Result("COPY origin buffer was not declared");
+
+  token = tokenizer_->NextToken();
+  if (token->IsEOL() || token->IsEOS())
+    return Result("missing 'TO' after COPY and buffer name");
+  if (!token->IsString())
+    return Result("expected 'TO' after COPY and buffer name");
+
+  name = token->AsString();
+  if (name != "TO")
+    return Result("expected 'TO' after COPY and buffer name");
+
+  token = tokenizer_->NextToken();
+  if (token->IsEOL() || token->IsEOS())
+    return Result("missing buffer name after TO");
+  if (!token->IsString())
+    return Result("invalid buffer name after TO");
+
+  name = token->AsString();
+  Buffer* buffer_to = script_->GetBuffer(name);
+  if (!buffer_to)
+    return Result("COPY destination buffer was not declared");
+
+  if (buffer_to->GetBufferType() == amber::BufferType::kUnknown) {
+    // Set destination buffer to mirror origin buffer
+    buffer_to->SetBufferType(buffer_from->GetBufferType());
+    buffer_to->SetWidth(buffer_from->GetWidth());
+    buffer_to->SetHeight(buffer_from->GetHeight());
+    buffer_to->SetSize(buffer_from->GetSize());
+  }
+
+  if (buffer_from->GetBufferType() != buffer_to->GetBufferType())
+    return Result("cannot COPY between buffers of different types");
+  if (buffer_from == buffer_to)
+    return Result("COPY origin and destination buffers are identical");
+
+  auto cmd = MakeUnique<CopyCommand>(buffer_from, buffer_to);
+  script_->AddCommand(std::move(cmd));
+
+  return ValidateEndOfStatement("COPY command");
 }
 
 }  // namespace amberscript
--- a/src/buffer.cc
+++ b/src/buffer.cc
@@ -217,7 +217,6 @@ Result Buffer::CopyTo(Buffer* buffer) const {
     return Result("Buffer::CopyBaseFields() buffers have a different height");
   if (buffer->size_ != size_)
     return Result("Buffer::CopyBaseFields() buffers have a different size");
-  buffer->values_.clear();
   buffer->values_ = values_;
   return {};
 }
--- a/src/buffer.cc
+++ b/src/buffer.cc
@@ -15,8 +15,6 @@
 #include "src/buffer.h"
 
 #include <cassert>
-#include <cstdlib>
-#include <cstring>
 
 namespace amber {
 namespace {
--- a/src/buffer.cc
+++ b/src/buffer.cc
@@ -15,6 +15,8 @@
 #include "src/buffer.h"
 
 #include <cassert>
+#include <cstdlib>
+#include <cstring>
 
 namespace amber {
 namespace {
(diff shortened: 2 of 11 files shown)


src/amberscript/parser.cc
```diff
@@ -60,14 +60,18 @@ Result Parser::Parse(const std::string& data) {
       r = ParseRun();
     } else if (tok == "SHADER") {
       r = ParseShaderBlock();
+    } else if (tok == "CLEAR") {
```
dj2: I think this was a bad merge, clear was moved up to line 53 to put these in sorted order (can you move COPY up there as well?)

src/executor.cc
```diff
@@ -111,6 +111,14 @@ Result Executor::Execute(Engine* engine,
...
     } else if (cmd->IsClearStencil()) {
       r = engine->DoClearStencil(cmd->AsClearStencil());
+    } else if (cmd->IsCopy()) {
+      auto copy = cmd->AsCopy();
+      if (copy->GetBufferTo()->GetMemPtr() != nullptr) {
+        // FIXME: clean the buffer?
+        copy->GetBufferTo()->SetMemPtr(nullptr);
+      }
+      std::vector<Value> data = copy->GetBufferFrom()->GetData();
+      copy->GetBufferTo()->SetData(std::move(data));
```
dj2: I think we want to put this method into buffer itself, then executor just does `Result r = buffer_from->CopyTo(buffer_to);`

Then in the buffer code you need to do something like:

````
// TODO(hevrard) Return error if buffer_to mem_ptr_ is backed by GPU memory
free(buffer_to->mem_ptr_);
buffer_to_->values_.clear();
bufer_to_values_ = values_;
if (mem_ptr_ != nullptr) {
  malloc(buffer_to->mem_ptr_, width * height * format byte size)
  std::memcpy(buffer_to->mem_ptr_, ....);
}
return {}
```

The GetMemPtr is a bit tricky as it returns either the mem_ptr_ or the values data buffer, and so we can't use it directly to do the copy.

This also means you'll need to be sure we're cleaning up mem_ptr_ correctly in this case. I think I assume'd it would get cleaned up through the GPU code?

src/amberscript/parser.cc
```diff
@@ -1115,5 +1119,57 @@ Result Parser::ParseExpect() {
...
+  if (!token->IsString())
+    return Result("invalid buffer name after TO");
+
+  name = token->AsString();
+
+  Buffer* buffer_to = script_->GetBuffer(name);
+  if (buffer_to == nullptr) {
+    return Result("COPY destination buffer was not declared");
+  }
+
```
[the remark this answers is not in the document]
hevrard: Good point!

src/command.h
```diff
@@ -194,6 +198,19 @@ class ComputeCommand : public Command {
   uint32_t z_ = 0;
 };
 
+class CopyCommand : public Command {
+ public:
+  explicit CopyCommand(Buffer* buffer_from, Buffer* buffer_to);
```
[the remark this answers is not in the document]
hevrard: It seems like all command constructors have an explicit keyword? e.g.

`explicit ProbeCommand(Pipeline* pipeline, Buffer* buffer);`

src/executor.cc
hevrard: Yeah I spotted that GetMemPtr is a bit tricky. Thanks for the guidance here.

So a buffer stores things in either data_, values_, or mem_ptr_? Is this exclusive, i.e., only one of these three is allowed at a time?

src/command.h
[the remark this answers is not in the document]
dj2: Yea, they got changed to include the pipeline and I forgot to remove the explicit. It doesn't hurt, it just doesn't do anything.

src/executor.cc
dj2: To be on the safe side, I'd just copy all of them. In general only one should be set, but if we don't need to make that assumption, I'd prefer not too.

hevrard:
Now the copy properly checks that buffer dimensions are compatible, and copy relevant fields. I still need to check whether mem_ptr_ is properly cleaned in GPU code.

As you will notice, it looks like we have to cast to Data/Format buffers in order to get the relevant size in bytes for each of them.

More generally, the tests/case/copy.amber script used to fail with:
```
36: cannot COPY between buffers of different types
```
Because the destination is of type "unknown" since it is not used in the pipeline. The parser now sets the destination buffer type, if it is unknown, to match the origin buffer type.

Now the test fails with:
```
37: buffer not in pipeline for EXPECT command
```

The EXPECT logic seems to limit itself to buffers used in a pipeline, shall we extend this to buffers used as COPY destinations?

hevrard:
OK just saw #377, which should remove the current EXPECT error I guess.

dj2:
Yea, the EXPECT pipeline requirement was needed to make it work, but has now been removed as I refactored the code. If you update on top of master hopefully that error should go away.

src/buffer.cc
```diff
@@ -30,16 +33,104 @@ FormatBuffer* Buffer::AsFormatBuffer() {
...
+}
+
+Result Buffer::CopyTo(Buffer* buffer) const {
+  Result result = CopyBaseFieldsTo(buffer);
+  if (!result.IsSuccess())
+    return result;
+
+  if (buffer->mem_ptr_) {
+    // TODO(hevrard) Return error if buffer_to mem_ptr_ is backed by GPU memory
+    std::free(buffer->mem_ptr_);
```
dj2: You can remove the if wrapper, it's safe to call free on a nullptr.

src/buffer.cc
```diff
@@ -30,16 +33,104 @@ FormatBuffer* Buffer::AsFormatBuffer() {
...
+        "DataBuffer::CopyTo() destination buffer is not a data buffer");
+
+  auto data_buffer = buffer->AsDataBuffer();
+
+  if (data_buffer->datum_type_.GetType() != datum_type_.GetType())
+    return Result("DataBuffer::CopyTo() buffers have a different datum type");
+  if (data_buffer->GetSizeInBytes() != GetSizeInBytes())
+    return Result("DataBuffer::CopyTo() buffers have a different byte size");
+
+  if (data_buffer->mem_ptr_) {
```
dj2: ditto

src/buffer.cc
```diff
@@ -30,16 +33,104 @@ FormatBuffer* Buffer::AsFormatBuffer() {
...
+
+  auto format_buffer = buffer->AsFormatBuffer();
+
+  if (format_buffer->GetFormat().GetFormatType() != GetFormat().GetFormatType())
+    return Result(
+        "FormatBuffer::CopyTo() buffers have a different format type");
+  if (format_buffer->GetSizeInBytes() != GetSizeInBytes())
+    return Result("FormatBuffer::CopyTo() buffers have a different byte size");
+
+  if (format_buffer->mem_ptr_) {
```
dj2: ditto

src/buffer.cc
dj2: Actually, you could add a help in Buffer which does the mem_ptr_ copy. Just send in the size from the overridden CopyTo method. Removes the duplicated code.

src/executor.cc
```diff
@@ -111,6 +111,17 @@ Result Executor::Execute(Engine* engine,
       r = engine->DoClearDepth(cmd->AsClearDepth());
     } else if (cmd->IsClearStencil()) {
       r = engine->DoClearStencil(cmd->AsClearStencil());
+    } else if (cmd->IsCopy()) {
+      auto copy = cmd->AsCopy();
+      auto buffer_from = copy->GetBufferFrom();
+      auto buffer_to = copy->GetBufferTo();
+      if (buffer_from->IsDataBuffer() && buffer_to->IsDataBuffer()) {
```
dj2: I don't think you need to do this. C++ should pick the overridden method by default so just doing `buffer_from->CopyTo(buffer_to)` should be fine I believe.

src/buffer.cc
hevrard: OK, I wasn't sure what is the best practice for Amber.

src/executor.cc
hevrard: Thanks, and apologies for all these newbie bumps, I managed to stick to C so far in my career so I am a bit walking on eggs when dealing with C++ inheritance and overriding.

src/buffer.cc
hevrard: Good point!

hevrard:
Thanks again for the feedback. I applied comments, rebased on master and added documentation.

The `copy_format_buffer.amber` test now fails because RUN is not implemented, I guess. I tried to use data buffers in `copy_data_buffer.amber`, but the EXPECT implementation is lacking.

`mem_ptr_` seems to be set only in:
```
src/vulkan/frame_buffer.cc:80:      info->buffer->SetMemPtr(color_images_.back()->HostAccessibleMemoryPtr());
```
In general, Buffer objects seems to be used only in Pipeline inside a BufferInfo struct. As there does not appear to exist any method that remove a buffer from the pipeline, I guess we cannot have dandling pointers in Buffer.mem_ptr_. I am not sure how to be able to check thoroughly whether the mem_ptr_ points to GPU memory?

dj2:
PR #383 fixes the format copy command. I'm working on adding support for the EQ command in EXPECT calls and I'm working to remove the data_ pointer in the data buffers which should fixup your data format test.

Once #383 lands, I think you can rebase this on top of that and verify your copy format test works. Then rename the copy data test be .expect_fail.amber and then I can rename it to just .amber when I get the EQ and data CLs landed.

dj2:
I have a CL up which deletes the data_ pointer from buffer and a second one ready which deletes the mem_ptr_ from buffer. Once they land, you just have to copy the values_ vector.

dj2:
Ok, everything has landed and you should be able to simplify this to just copy the values_ vector in the base buffer class. Overrides aren't needed anymore.

hevrard:
Thanks for this! I'm busy on other fronts but will take care of this tomorrow.

src/buffer.h
```diff
@@ -93,6 +93,9 @@ class Buffer {
     return reinterpret_cast<const T*>(values_.data());
   }
 
+  /// Copy the buffer values to an other one
+  virtual Result CopyTo(Buffer* buffer) const;
```
dj2: Nit: remove virtual as we don't override this anymore

src/buffer.cc
```diff
@@ -210,6 +210,18 @@ FormatBuffer* Buffer::AsFormatBuffer() {
...
 }
 
+Result Buffer::CopyTo(Buffer* buffer) const {
+  if (buffer->width_ != width_)
+    return Result("Buffer::CopyBaseFields() buffers have a different width");
+  if (buffer->height_ != height_)
+    return Result("Buffer::CopyBaseFields() buffers have a different height");
+  if (buffer->size_ != size_)
+    return Result("Buffer::CopyBaseFields() buffers have a different size");
+  buffer->values_.clear();
```
dj2: nit: clear() isn't needed as the buffer will get removed when we assign it below.

PR #377, opened 2019-03-17
 Remove pipeline from probe commands.

### LINKED ISSUE #368: Remove pipeline from EXPECT commands
Currently to get expect commands to work I added a PIPELINE parameter. This should not be needed but is a hold over from how the system used to work. We need to verify that the Probe and ProbeSSBO calls do not need to call into the pipeline and, if so, remove those calls. This will then let us remove PIPELINE from EXPECT.

The calls into the pipeline from the executor were due to late copying of data, I believe. All data should be copied at the end of each compute and draw command so the amber::Buffers should always have the full set of data needed to probe.

dj2:
This CL updates the probe commands to contain the buffers needed instead
of providing the pipeline.

Fixes #368.


src/vkscript/command_parser_test.cc
```diff
@@ -3830,7 +3830,7 @@ TEST_F(CommandParserTest, ToleranceWithCommas) {
 }
 
 TEST_F(CommandParserTest, ProbeSSBOWithTolerance) {
-  std::string data = R"(
+  std::string data = R"(ssbo 3:6 2
```
hevrard: Most tests start with a new line right after 'R"(' ?

src/vkscript/command_parser_test.cc
```diff
@@ -3887,7 +3887,8 @@ probe all rgba 0.2 0.3 0.4 0.5)";
 }
 
 TEST_F(CommandParserTest, ProbeSSBOWithDescriptorSet) {
-  std::string data = "probe ssbo vec3 3:6 2 >= 2.3 4.2 1.2";
+  std::string data = R"(ssbo 3:6 2
```
hevrard: Newline after 'R"(' ?

src/vkscript/command_parser_test.cc
```diff
@@ -3920,7 +3921,8 @@ TEST_F(CommandParserTest, ProbeSSBOWithDescriptorSet) {
 }
 
 TEST_F(CommandParserTest, ProbeSSBOWithFloats) {
-  std::string data = "probe ssbo vec3 6 2 >= 2.3 4.2 1.2";
+  std::string data = R"(ssbo 6 2
```
hevrard: Newline after R"( ?

src/vkscript/command_parser_test.cc
```diff
@@ -3953,8 +3955,9 @@ TEST_F(CommandParserTest, ProbeSSBOWithFloats) {
 }
 
 TEST_F(CommandParserTest, MultiProbeSSBOWithFloats) {
-  std::string data =
-      "probe ssbo vec3 6 2 >= 2.3 4.2 1.2\nprobe ssbo vec3 6 2 >= 2.3 4.2 1.2";
+  std::string data = R"(ssbo 6 2
```
hevrard: Newline after R"( ?

src/vkscript/command_parser_test.cc
```diff
@@ -3986,7 +3989,8 @@ TEST_F(CommandParserTest, MultiProbeSSBOWithFloats) {
 }
 
 TEST_F(CommandParserTest, ProbeSSBOWithInts) {
-  std::string data = "probe ssbo i16vec3 6 2 <= 2 4 1";
+  std::string data = R"(ssbo 6 2
```
hevrard: Newline after R"( ?

src/vulkan/compute_pipeline.cc
```diff
@@ -116,6 +116,14 @@ Result ComputePipeline::Compute(uint32_t x, uint32_t y, uint32_t z) {
       command_->GetCommandBuffer(), VK_PIPELINE_BIND_POINT_COMPUTE, pipeline);
   device_->GetPtrs()->vkCmdDispatch(command_->GetCommandBuffer(), x, y, z);
 
+  r = command_->End();
```
hevrard: This change looks unrelated? Good fix though.

src/vulkan/graphics_pipeline.cc
```diff
@@ -819,7 +815,19 @@ Result GraphicsPipeline::ClearBuffer(const VkClearValue& clear_value,
 
   DeactivateRenderPassIfNeeded();
 
-  return frame_->CopyColorImagesToHost(command_.get());
+  r = frame_->CopyColorImagesToHost(command_.get());
```
hevrard: Is this change related to pipeline removal from probe commands?

src/vulkan/compute_pipeline.cc
dj2: It's needed in order to finish the command and have the data available for probe. Without this we'd need the pipeline in order to submit the command correctly.

src/vulkan/graphics_pipeline.cc
dj2: ditto

PR #382, opened 2019-03-18
 [vulkan] Wrap render pass into RAII object.

dj2:
This Cl wraps the begin/end render pass calls inside a RAII object.
This removes the possibility of having called begin render pass without
subsequently calling end.

### DIFF
--- a/src/vulkan/graphics_pipeline.h
+++ b/src/vulkan/graphics_pipeline.h
@@ -49,6 +49,9 @@ class GraphicsPipeline : public Pipeline {
       const std::vector<VkPipelineShaderStageCreateInfo>&);
   ~GraphicsPipeline() override;
 
+  // Pipeline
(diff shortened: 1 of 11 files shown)


src/vulkan/graphics_pipeline.h
```diff
@@ -66,7 +69,8 @@ class GraphicsPipeline : public Pipeline {
 
   Result Draw(const DrawArraysCommand* command, VertexBuffer* vertex_buffer);
 
-  const FrameBuffer* GetFrame() const { return frame_.get(); }
+  VkRenderPass GetRenderPass() const { return render_pass_; }
+  FrameBuffer* GetFrame() const { return frame_.get(); }
```
dneto0: Preferable to call this GetFrameBuffer.  But that can wait.

PR #386, opened 2019-03-19
 [vulkan] Add CommandBuffer guard.

dj2:
This Cl wraps the CommandBuffer object and verifies the begin/submit
methods are called properly.

### DIFF
--- a/src/vulkan/command_buffer.cc
+++ b/src/vulkan/command_buffer.cc
@@ -14,8 +14,6 @@
 
 #include "src/vulkan/command_buffer.h"
 
-#include <cassert>
-
 #include "src/vulkan/command_pool.h"
 #include "src/vulkan/device.h"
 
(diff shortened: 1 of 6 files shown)


src/vulkan/command_buffer.cc
```diff
@@ -106,9 +96,28 @@ Result CommandBuffer::SubmitAndReset(uint32_t timeout_ms) {
...
 
+CommandBufferGuard::CommandBufferGuard(CommandBuffer* buffer) : buffer_(buffer) {
+  assert(!buffer_->guarded_);
+
+  buffer_->guarded_ = true;
+  result_ = buffer_->BeginRecording();
+}
+
+CommandBufferGuard::~CommandBufferGuard() {
+  assert(submitted_);
```
dneto0: This isn't the right thing to do if the flow has stopped before submission.
E.g. if any of the intermediate things fail with an error result, then this will assert out rather than report the error.

src/vulkan/compute_pipeline.cc
```diff
@@ -90,23 +90,25 @@ Result ComputePipeline::Compute(uint32_t x, uint32_t y, uint32_t z) {
...
 
-  BindVkDescriptorSets(pipeline_layout);
+    BindVkDescriptorSets(pipeline_layout);
 
-  r = RecordPushConstant(pipeline_layout);
-  if (!r.IsSuccess())
-    return r;
+    r = RecordPushConstant(pipeline_layout);
+    if (!r.IsSuccess())
+      return r;
```
dneto0: Example, if this recording fails, then the guard will be in non-submitted state and it will assert out.

src/vulkan/command_buffer.cc
dj2: Doh, fixed.

Issue #54, opened 2018-11-15
Vulkan: Remove asserts from the codebase


dj2:
The vulkan engine currently asserts in places where things should never happen. We should probably return in these cases so that amber can be used as a library if desired.

dj2:
This is pretty much done. There are a few remaining but I think they all make sense.

PR #388, opened 2019-03-20
[vulkan] Cleanup some asserts

dj2:
This CL removes some of the asserts in the vulkan engine and returns
results instead.

Issue #54.

### DIFF
--- a/src/vulkan/engine_vulkan.cc
+++ b/src/vulkan/engine_vulkan.cc
@@ -15,7 +15,6 @@
 #include "src/vulkan/engine_vulkan.h"
 
 #include <algorithm>
-#include <cassert>
 #include <set>
 #include <utility>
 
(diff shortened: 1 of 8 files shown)


src/vulkan/engine_vulkan.cc
```diff
@@ -32,27 +31,32 @@ namespace {
...
-      return VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT;
+      *ret = VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT;
+      break;
     case kShaderTypeCompute:
-      return VK_SHADER_STAGE_COMPUTE_BIT;
-    case kShaderTypeMulti:
-      // It's an error if this arrives here ...
+      *ret = VK_SHADER_STAGE_COMPUTE_BIT;
       break;
+    case kShaderTypeMulti:
```
dneto0: what the heck is multi?   the header file isn't very illuminating.  But that's for another CL

dj2: It's a spirv module with multiple shaders in it.So you can have a vertex and fragment shaders in a single module. so, a multi shader module, heh

PR #389, opened 2019-03-20
[vulkan] Move GetFrameBuffer out of engine code.

dj2:
This CL moves the GetFrameBuffer code up into the Amber object and
removes from the engine code. There should be no engine specific work
involved in retrieving the framebuffer.

### DIFF
--- a/src/amber.cc
+++ b/src/amber.cc
@@ -28,6 +28,41 @@
 #include "src/vkscript/parser.h"
 
 namespace amber {
+namespace {
+
+const FormatType kDefaultFramebufferFormat = FormatType::kB8G8R8A8_UNORM;
+
+Result GetFrameBuffer(Buffer* buffer, std::vector<Value>* values) {
+  values->clear();
+
+  // TODO(jaebaek): Support other formats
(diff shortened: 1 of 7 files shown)


src/amber.cc
```diff
@@ -28,6 +28,41 @@
...
+
+  const auto texel_stride = buffer->AsFormatBuffer()->GetTexelStride();
+  const auto row_stride = buffer->AsFormatBuffer()->GetRowStride();
+
+  for (uint32_t y = 0; y < buffer->GetHeight(); ++y) {
+    for (uint32_t x = 0; x < buffer->GetWidth(); ++x) {
+      Value pixel;
+
+      const uint8_t* ptr_8 = cpu_memory + (row_stride * y) + (texel_stride * x);
+      const uint32_t* ptr_32 = reinterpret_cast<const uint32_t*>(ptr_8);
```
dneto0: I was tempted to ask for an assertion that the byte offset was a multiple of 4.... but we don't even know that the base pointer is at an address aligned to 4bytes. So we're just going to rely on the caller being good.

PR #407, opened 2019-03-22
[vulkan] Remove memory pointer from Buffer.

dj2:
This CL removes the memory pointer from the buffer object and copies the
host accessible memory into the value vector.

### DIFF
--- a/src/buffer.h
+++ b/src/buffer.h
@@ -85,21 +85,8 @@ class Buffer {
   /// size of the data provided.
   virtual Result SetData(std::vector<Value>&& data) = 0;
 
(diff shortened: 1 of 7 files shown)


src/buffer.h
```diff
@@ -85,21 +85,8 @@ class Buffer {
   /// size of the data provided.
   virtual Result SetData(std::vector<Value>&& data) = 0;
 
-  void SetMemPtr(void* ptr) { mem_ptr_ = ptr; }
-
-  const void* GetMemPtr() const {
```
dneto0: Law of Demeter would encourage you to keep this method.  But meh.

PR #408, opened 2019-03-22
Initialize amber structures.

dj2:
This CL adds initializers for the amber::Options and amber::BufferInfo
structures.


samples/amber.cc
```diff
@@ -399,7 +399,7 @@ int main(int argc, const char** argv) {
       auto pos = options.image_filename.find_last_of('.');
       bool usePNG = pos != std::string::npos &&
                     options.image_filename.substr(pos + 1) == "png";
-      for (amber::BufferInfo buffer_info : amber_options.extractions) {
+      for (amber::BufferInfo& buffer_info : amber_options.extractions) {
```
dneto0: this could probably be const?

PR #411, opened 2019-03-23
Add EXPECT EQ

dj2:


src/amberscript/parser.cc
```diff
@@ -995,6 +1019,41 @@ Result Parser::ParseClear() {
...
+                           const DatumType& type,
+                           std::vector<Value>* values) {
+  assert(values);
+
+  auto token = tokenizer_->NextToken();
+  while (!token->IsEOL() && !token->IsEOS()) {
+    Value v;
+
+    if ((type.IsFloat() || type.IsDouble())) {
+      if (!token->IsInteger() && !token->IsDouble()) {
```
dneto0: IsInteger -> IsFloat

dj2: The token doesn't have an IsFloat, just IsDouble. We ConvertToDouble a couple of lines down, so if we have an Integer or a Double then we're good to go. If we get anything other then a number then things go wrong.

dneto0: ok.  I was wrong.

src/amberscript/parser.h
```diff
@@ -64,6 +65,10 @@ class Parser : public amber::Parser {
   Result ParseClear();
   Result ParseExpect();
 
+  Result ParseValues(const std::string& name,
```
dneto0: Method comment please.  What is the name for?  Are we appending to *values or completely overwriting.

dj2: done

PR #415, opened 2019-03-26
[vulkan] Remove vulkan/format_data

dj2:
This Cl removes the free methods in vulkan/format_data and attaches them
to the classes where they make the most sense.


src/vulkan/transfer_image.cc
```diff
@@ -46,15 +45,15 @@ const VkImageCreateInfo kDefaultImageInfo = {
...
 
 TransferImage::TransferImage(Device* device,
-                             VkFormat format,
+                             const Format& format,
                              VkImageAspectFlags aspect,
                              uint32_t x,
                              uint32_t y,
                              uint32_t z)
-    : Resource(device, x * y * z * VkFormatToByteSize(format)),
+    : Resource(device, x * y * z * format.GetByteSize()),
```
dneto0: Hmm. I'm surprised this doesn't need a row-stride...

src/vulkan/vertex_buffer_test.cc
```diff
@@ -202,6 +207,7 @@ TEST_F(VertexBufferTest, R16G16B16A16_SNORM) {
   values[3].SetIntValue(255);
 
   auto format = MakeUnique<Format>();
+  format->SetFormatType(FormatType::kR16G16B16A16_UNORM);
```
dneto0: Should be SNORM

src/vulkan/vertex_buffer_test.cc
```diff
@@ -280,6 +288,7 @@ TEST_F(VertexBufferTest, R16G11B10_SFLOAT) {
   values[2].SetDoubleValue(0.1171875);
 
   auto format = MakeUnique<Format>();
+  format->SetFormatType(FormatType::kR32G32_SFLOAT);  // big enough to cover.
```
dneto0: Needs a B component.

src/vulkan/vertex_buffer_test.cc
```diff
@@ -312,6 +321,7 @@ TEST_F(VertexBufferTest, R10G16B11_SFLOAT) {
   values[2].SetDoubleValue(14.0);
 
   auto format = MakeUnique<Format>();
+  format->SetFormatType(FormatType::kR32G32_SFLOAT);  // big enough to cover.
```
dneto0: Needs a B component

src/vulkan/vertex_buffer_test.cc
```diff
@@ -344,6 +354,7 @@ TEST_F(VertexBufferTest, R11G16B10_SFLOAT) {
   values[2].SetDoubleValue(0.1171875);
 
   auto format = MakeUnique<Format>();
+  format->SetFormatType(FormatType::kR32G32_SFLOAT);  // big enough to cover.
```
dneto0: Needs a B component.

src/format.cc
```diff
@@ -22,4 +22,152 @@ Format::Format(const Format&) = default;
...
+    case FormatType::kX8_D24_UNORM_PACK32:
+    case FormatType::kR8G8B8A8_SINT:
+    case FormatType::kR8G8B8A8_SNORM:
+    case FormatType::kR8G8B8A8_SRGB:
+    case FormatType::kR8G8B8A8_SSCALED:
+    case FormatType::kR8G8B8A8_UINT:
+    case FormatType::kR8G8B8A8_UNORM:
+    case FormatType::kR8G8B8A8_USCALED:
+      return 4U;
+    case FormatType::kD32_SFLOAT_S8_UINT:
```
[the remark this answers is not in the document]
dj2: Ack

src/vulkan/transfer_image.cc
dj2: This is the same as previously so it seems to be working correctly.

src/vulkan/vertex_buffer_test.cc
dj2: done

src/vulkan/vertex_buffer_test.cc
dj2: These don't actually match the components. We purposely make a weird component below. I just have to have a format for things to work correctly. Since there is no matching format, I made one that is at least close in byte size.

src/vulkan/vertex_buffer_test.cc
dj2: same

src/vulkan/vertex_buffer_test.cc
dj2: same

src/vulkan/transfer_image.cc
jaebaek: @dneto0 is it possible that the actually created VkImage has a different number of texels in a row (not `x`) when I requested creating an image whose size is `x * y * z`?

jaebaek: If this is true, I think we need another CL to handle this case (or adding a TODO here?)

dj2: This is just a refactoring CL. If there is an issue here can we please file a bug with a repro case? This is currently working, so without a repro it's hard to tell if there is an issue or not.

PR #416, opened 2019-03-27
[amberscript] Add CLEAR_COLOR parsing.

### LINKED ISSUE #344: Parse CLEAR_COLOR command

dj2:
This Cl adds CLEAR_COLOR parsing into the AmberScript parser. The EXPECT
EQ was also fixed to use the correct 0-255 range instead of 0.0-1.0
range for colour expectations.

Fixes #344.


src/amberscript/parser.cc
```diff
@@ -1235,5 +1237,66 @@ Result Parser::ParseCopy() {
...
+    return Result("missing pipeline name for CLEAR_COLOR command");
+
+  auto* pipeline = script_->GetPipeline(token->AsString());
+  if (!pipeline) {
+    return Result("unknown pipeline for CLEAR_COLOR command: " +
+                  token->AsString());
+  }
+  if (!pipeline->IsGraphics()) {
+    return Result(
+        "CLEAR_COLOR command requires graphics pipeline, got compute");
```
hevrard: This conditional + message make the assumption that a pipeline is either graphics or compute, which may not be future proof? That is nit-picking, but maybe the message should just say "CLEAR_COLOR command argument is not a graphics pipeline" ?

src/command.h
```diff
@@ -418,6 +419,7 @@ class ClearColorCommand : public PipelineCommand {
   explicit ClearColorCommand(Pipeline* pipeline);
   ~ClearColorCommand() override;
 
+  // Colours are stored in the range 0.0-1.0
```
hevrard: Nit: `the range 0.0-1.0` -> `the range 0.0 - 1.0` (add spaces to exactly match the previous comment)

src/amberscript/parser.cc
dj2: Good point. Fixed other instances as well.

src/command.h
dj2: done

PR #420, opened 2019-03-27
Hide pipeline buffer formats

dj2:
The default colour formats are only used in the pipeline class so this
CL moves them into the anonymous namespace.


src/pipeline.cc
```diff
@@ -21,6 +21,15 @@
...
 
 namespace amber {
+namespace {
+
+const char* kDefaultColorBufferFormat = "B8G8R8A8_UNORM";
+const char* kDefaultDepthBufferFormat = "D32_SFLOAT_S8_UINT";
+
+}  // namespace
+
+const char* Pipeline::kGeneratedColorBuffer = "framebuffer";
```
hevrard: Should this const be used all over the place? I spotted uses of the string "framebuffer" here and there:

```
samples/amber.cc:379:    buffer_info.buffer_name = "framebuffer";
samples/amber.cc:403:        if (buffer_info.buffer_name == "framebuffer") {
samples/amber.cc:439:          if (buffer_info.buffer_name == "framebuffer")
src/amber.cc:192:    if (buffer_info.buffer_name == "framebuffer") {
src/pipeline.cc:34:const char* Pipeline::kGeneratedColorBuffer = "framebuffer";
src/vkscript/parser.cc:204:    } else if (str == "framebuffer") {
```

Not sure how/whether to export this to samples/ though.

dj2: Fixed the ones in src and added a kGeneratedColorBuffer in the samples/amber.cc anonymous namespace. Not exported, but still reduces the string count.

PR #421, opened 2019-03-27
[vkscript] report error on invalid ssbo subdata offset.

### LINKED ISSUE #296: Amber gives a helpful error when an SSBO probe's offset is bad, but not when a bad offset is used to set an SSBO value
Amber gives a nice error when I run it on [this file](https://github.com/google/amber/files/2864971/good_error.amber.txt), telling me that I'm probing using an offset that is not a multiple of the data type size.

But for [this file](https://github.com/google/amber/files/2864970/bad_no_error.amber.txt) I get no equivalent error when I set an SSBO value using an offset that is not a multiple of the size of the type.

I think it would be great to have an error in both case, as thinking of offsets in terms of element counts rather than bytes is an easy mistake to make.

dj2:
This CL adds an error message if the requested offset is not a multiple
of the data type. This also updates the DatumType code to handle std140
layouts where we have to add an extra element for 3 item arrays.

Currently all things are considered to be in std140.

Fixes #296.


dj2:
This seems to be working with all scripts we have. But, I'm still hazy on what has to be std140 and what is std430 by default, so it's possible there is an issue here I'm not seeing. If so, we should add a script that exposes the problem.

PR #434, opened 2019-03-28
[amberscript] add DEVICE_FEATURE.

### LINKED ISSUE #347: Add feature detection
Extend the grammar and parsing to handle required features.

dj2:
This CL adds the DEVICE_FEATURE command which allows requesting specific
features be enabled.

Fixes #347.


tests/cases/draw_rectangles.amber
```diff
@@ -0,0 +1,77 @@
...
+DEVICE_FEATURE vertexPipelineStoresAndAtomics
+
+SHADER vertex vtex_shader GLSL
+#version 430
+
+layout(location = 0) in vec4 position;
+layout(location = 0) out vec4 frag_color;
+
+layout(set = 0, binding = 0) buffer block1 {
+  vec4 in_color;
```
dneto0: Slightly confusing as this is only read.  However, I think it's not marked NonWritable, so that exercises the device feature. So this is ok.

PR #437, opened 2019-03-30
Remove accidentally added file

dj2:
This was an example of a possible design direction that got added accidentally. We aren't going with BUFFER_UPDATE so this is incorrect.


dj2:
TBR=dneto@

PR #440, opened 2019-04-01
Update Dawn config file

sarahM0:
- Update Amber to catch up with Dawn changes
- Update Dawn config file to work with both Vulkan and Metal backends
- Update DoClear


samples/config_helper_dawn.cc
```diff
@@ -43,12 +43,19 @@ amber::Result ConfigHelperDawn::CreateConfig(
   auto r = dawn::CreateMetalDevice(&dawn_instance_, &dawn_device_);
   if (!r.IsSuccess())
     return r;
-#else
-  return amber::Result("Can't make Dawn engine config");
+#else  // assuming VULKAN
+  dawn_instance_.DiscoverDefaultAdapters();
+  for (dawn_native::Adapter& adapter : dawn_instance_.GetAdapters()) {
```
dj2: Does this work for metal as well? Can we move the #if inside this for() and have it detect vulkan/metal and d3d in the future?

dneto0: Hm. Good question.  When the Dawn-over-Metal support was first added, the initial device bringup required access to an Objective-C API, and so I had to hide it away in a .mm file.  (Well that's the way I remember it, there might have been a different way...)
Since then Dawn's native backend bringup APIs have changed and maybe we only need the ::dawn_native::BackendType enum values that Dawn itself defines. In other words maybe the whole dawn_device_metal.mm file can go away?

sarahM0: I tried it with Metal but I get a segmentation fault, so it doesn't work right now. I'm going to leave it as it is, @dj2 is going to file a bug for it.

dj2:
Is this CL or #454 the current one?

PR #441, opened 2019-04-02
[vulkan] simplify image transition code.

dj2:
This CL splits the ChangeLayout method into a method to change to Draw
mode and a method to change to Probe mode. The framebuffer is always
transitioned to Draw mode after initialization. The RenderPassGuard now
transitions the framebuffer into draw layout on initialize and out of
draw layout on shutdown.

### DIFF
--- a/src/vulkan/frame_buffer.cc
+++ b/src/vulkan/frame_buffer.cc
@@ -43,7 +43,8 @@ FrameBuffer::~FrameBuffer() {
   }
 }
 
-Result FrameBuffer::Initialize(VkRenderPass render_pass,
+Result FrameBuffer::Initialize(CommandBuffer* command_buffer,
+                               VkRenderPass render_pass,
                                const Format& depth_format) {
   std::vector<VkImageView> attachments;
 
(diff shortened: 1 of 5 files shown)


src/vulkan/frame_buffer.cc
```diff
@@ -156,24 +127,46 @@ Result FrameBuffer::ChangeFrameImageLayout(CommandBuffer* command,
...
+    return;
+
   for (auto& img : color_images_) {
-    ChangeFrameImageLayout(command, FrameImageState::kProbe);
-    img->CopyToHost(command);
+    auto barrier = img->CreateBarrier(VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL,
+                                      VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL);
+    img->ImageBarrier(command, barrier,
+                      VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT,
+                      VK_PIPELINE_STAGE_TRANSFER_BIT);
```
hevrard: This code assumes ChangeFrameToProbeLayout() is called only when frame_image_layout is either kProbe or kClearOrDraw. While that may be the case now, wouldn't it be good and more future proof to somehow treat the kInit case (or any other cases)?

src/vulkan/transfer_image.cc
```diff
@@ -34,12 +34,11 @@ const VkImageCreateInfo kDefaultImageInfo = {
     1,                                   /* arrayLayers */
     VK_SAMPLE_COUNT_1_BIT,               /* samples */
     VK_IMAGE_TILING_OPTIMAL,             /* tiling */
-    VK_IMAGE_USAGE_TRANSFER_SRC_BIT |
-        VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, /* usage */
-    VK_SHARING_MODE_EXCLUSIVE,               /* sharingMode */
-    0,                                       /* queueFamilyIndexCount */
-    nullptr,                                 /* pQueueFamilyIndices */
-    VK_IMAGE_LAYOUT_UNDEFINED,               /* initialLayout */
+    0,                                   /* usage */
```
hevrard: Is the change of the 'usage' arg from flags to 0 made on-purpose? (I'm not sure about the implications here, just seems like it could be a typo?)

src/vulkan/frame_buffer.cc
dj2: This is an excellent point, and one I encountered 3 patch sets down. I have a CL coming which stores the current image layout in the image and allows doing any transition instead of just ones we thought of.

src/vulkan/transfer_image.cc
dj2: Yes, we always set usage, so having the default value here was confusing as it never got used.

PR #443, opened 2019-04-03
[amberscript] Add REPEAT command

### LINKED ISSUE #425: Add REPEAT to AmberScript
We want the abilty to run a given draw command multiple times. To that end, add:

```
REPEAT n
  ...
END
```

dj2:
This CL adds the REPEAT directive to AmberScript. This allows running a
given set of commands `n` times.

Fixes #425.


docs/amber_script.md
```diff
@@ -383,7 +401,7 @@ void main() {
 }
 END  # shader
 
-BUFFER kComputeBuffer TYPE vec2<int32> SIZE 524288 FILL 0
+BUFFER kComputeBuffer DATA_TYPE vec2<int32> SIZE 524288 FILL 0
```
dneto0: This is unrelated to the REPEAT functionality.

And I think ELEM_TYPE is better than DATA_TYPE.  Because actually the DATA_TYPE is _array of_ ....

src/amberscript/parser.cc
```diff
@@ -96,6 +90,7 @@ Result Parser::Parse(const std::string& data) {
     if (!r.IsSuccess())
       return Result(make_error(r.Error()));
   }
+  script_->SetCommands(std::move(command_list_));
```
dneto0: Maybe it's a subtle point, but this actually does ensure that command_list is empty afterward.
The std::vector move constructor guarantees it.

src/amberscript/parser.cc
```diff
@@ -1309,5 +1326,50 @@ Result Parser::ParseDeviceFeature() {
   return ValidateEndOfStatement("DEVICE_FEATURE command");
 }
 
+Result Parser::ParseRepeat() {
```
dneto0: This whole thing works without a stack because REPEAT is not itself repeatable.  Works for me.

src/amberscript/parser.h
```diff
@@ -67,6 +67,9 @@ class Parser : public amber::Parser {
   Result ParseExpect();
   Result ParseCopy();
   Result ParseDeviceFeature();
+  Result ParseRepeat();
+  bool isRepeatable(const std::string& name) const;
```
dneto0: nit: Is there a reason this is lower-case-first?

src/vkscript/command_parser.cc
```diff
@@ -647,7 +647,13 @@ Result CommandParser::ProcessSSBO() {
       return Result("Invalid size value for ssbo command: " +
                     token->ToOriginalString());
 
-    cmd->SetSize(token->AsUint32());
+    uint32_t size = token->AsUint32();
+    cmd->SetSize(size);
+
+    // Resize the buffer so we'll correctly create the descriptor sets.
+    auto* buf = cmd->GetBuffer();
```
dneto0: Is this actually related?

src/vulkan/engine_vulkan.cc
```diff
@@ -211,6 +211,29 @@ Result EngineVulkan::CreatePipeline(amber::Pipeline* pipeline) {
     info.vk_pipeline->AsGraphics()->SetIndexBuffer(buf);
   }
 
+  for (const auto& buf_info : pipeline->GetBuffers()) {
```
dneto0: Is this actually related?

src/vulkan/pipeline.cc
```diff
@@ -244,6 +244,8 @@ Result Pipeline::AddPushConstant(const BufferCommand* command) {
 Result Pipeline::AddDescriptor(const BufferCommand* cmd) {
   if (cmd == nullptr)
     return Result("Pipeline::AddDescriptor BufferCommand is nullptr");
+  if (cmd->IsPushConstant())
```
dneto0: is this  related?

src/vulkan/pipeline.cc
```diff
@@ -296,12 +298,14 @@ Result Pipeline::AddDescriptor(const BufferCommand* cmd) {
         "and binding");
   }
 
-  auto* buf_desc = static_cast<BufferDescriptor*>(desc);
-  Result r =
-      buf_desc->AddToBuffer(cmd->GetDatumType().GetType(), cmd->GetOffset(),
-                            cmd->GetSize(), cmd->GetValues());
-  if (!r.IsSuccess())
-    return r;
+  if (!cmd->GetValues().empty()) {
```
dneto0: Is this related?

docs/amber_script.md
dj2: The field is already called DATA_TYPE, this just fixes the examples to match the parser.

src/amberscript/parser.cc
dj2: Yes, but I need to swap it back to get the previous command list entries back into the list.

src/amberscript/parser.h
dj2: fix

src/vkscript/command_parser.cc
dj2: I order to make the repeat.amber script work but buffers needed to work. To make buffers work, requires the ssbo's to be sized correctly. So ... kinda?

src/vulkan/engine_vulkan.cc
dj2: ditto

src/vulkan/pipeline.cc
dj2: ditto

src/vulkan/pipeline.cc
dj2: ditto

PR #444, opened 2019-04-04
Add EXPECT EQ_BUFFER command

### LINKED ISSUE #431: Add EXPECT EQ_BUFFER
We want to be able to compare two complete buffers. Add:

```
EXPECT buffer1 EQ_BUFFER buffer2
```

hevrard:
Add EXPECT EQ_BUFFER command.

Fix #431

### DIFF
--- a/src/amberscript/parser.cc
+++ b/src/amberscript/parser.cc
@@ -1080,13 +1080,18 @@ Result Parser::ParseExpect() {
       return Result("unknown buffer name for EXPECT EQ_BUFFER command");
 
     if (buffer->GetBufferType() != buffer_2->GetBufferType())
-      return Result("EXPECT EQ_BUFFER command cannot compare buffers of different type");
+      return Result(
+          "EXPECT EQ_BUFFER command cannot compare buffers of different type");
     if (buffer->GetSize() != buffer_2->GetSize())
-      return Result("EXPECT EQ_BUFFER command cannot compare buffers of different size");
+      return Result(
+          "EXPECT EQ_BUFFER command cannot compare buffers of different size");
     if (buffer->GetWidth() != buffer_2->GetWidth())
-      return Result("EXPECT EQ_BUFFER command cannot compare buffers of different width");
+      return Result(
+          "EXPECT EQ_BUFFER command cannot compare buffers of different width");
     if (buffer->GetHeight() != buffer_2->GetHeight())
-      return Result("EXPECT EQ_BUFFER command cannot compare buffers of different height");
+      return Result(
+          "EXPECT EQ_BUFFER command cannot compare buffers of different "
+          "height");
 
     auto cmd = MakeUnique<CompareBufferCommand>(buffer, buffer_2);
     script_->AddCommand(std::move(cmd));
--- a/src/amberscript/parser.cc
+++ b/src/amberscript/parser.cc
@@ -1082,9 +1082,10 @@ Result Parser::ParseExpect() {
       return Result("invalid buffer name in EXPECT EQ_BUFFER command");
 
     auto* buffer_2 = script_->GetBuffer(token->AsString());
-    if (!buffer_2)
+    if (!buffer_2) {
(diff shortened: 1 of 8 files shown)


hevrard:
Three remarks:
1. Should the ParseExpect function be split per EXPECT comparator type ("EQ_BUFFER", "IDX", future ones)?
2. Are buffer dimensions (width, height) relevant to compare at parsing time?
3. When buffer values are different, should we report the first different value item that differs? (right now it just says "value are different")

src/amberscript/parser.cc
```diff
@@ -1066,8 +1066,42 @@ Result Parser::ParseExpect() {
...
+    return Result("Invalid comparator in EXPECT command");
+
+  if (token->AsString() == "EQ_BUFFER") {
+    token = tokenizer_->NextToken();
+    if (!token->IsString())
+      return Result("invalid buffer name in EXPECT EQ_BUFFER command");
+
+    auto* buffer_2 = script_->GetBuffer(token->AsString());
+    if (!buffer_2)
+      return Result("unknown buffer name for EXPECT EQ_BUFFER command");
```
dj2: Add the buffer name?
```
("unknown buffer name for EXPECT EQ_BUFFER command: " + token->AsString())
```

src/amberscript/parser.cc
```diff
@@ -1066,8 +1066,42 @@ Result Parser::ParseExpect() {
...
+
+    if (buffer->GetBufferType() != buffer_2->GetBufferType())
```
dj2: And {}'s on an if where the body is multiple lines

src/buffer.cc
```diff
@@ -221,6 +221,21 @@ Result Buffer::CopyTo(Buffer* buffer) const {
...
+Result Buffer::IsEqual(Buffer* buffer) const {
+  if (buffer->buffer_type_ != buffer_type_)
+    return Result{"Buffers have a different type"};
+  if (buffer->size_ != size_)
+    return Result{"Buffers have a different size"};
+  if (buffer->width_ != width_)
+    return Result{"Buffers have a different width"};
+  if (buffer->height_ != height_)
+    return Result{"Buffers have a different height"};
+  if (buffer->values_ != values_)
```
dj2: We probably want to do this the hard way. We want to be able to spit out two things, the first index where the buffers differ and their values and the number of indexes in which the buffers differ so, something along the lines of:

```
Buffers have different values. 1234 values differed, first difference at 12 values 7 != 8
```

Something like that.

src/amberscript/parser.cc
```diff
@@ -1066,8 +1066,42 @@ Result Parser::ParseExpect() {
...
+      return Result(
+          "EXPECT EQ_BUFFER command cannot compare buffers of different type");
+    if (buffer->GetSize() != buffer_2->GetSize())
+      return Result(
+          "EXPECT EQ_BUFFER command cannot compare buffers of different size");
+    if (buffer->GetWidth() != buffer_2->GetWidth())
```
dj2: I think these are fine as we should always know the size, may as well catch early if possible.

src/amberscript/parser.cc
```diff
@@ -1066,8 +1066,42 @@ Result Parser::ParseExpect() {
     return Result("unknown buffer name for EXPECT command");
 
   token = tokenizer_->NextToken();
-  if (!token->IsString() || token->AsString() != "IDX")
-    return Result("missing IDX in EXPECT command");
+
+  if (!token->IsString())
+    return Result("Invalid comparator in EXPECT command");
+
+  if (token->AsString() == "EQ_BUFFER") {
```
dj2: We may refactor later to split apart if it gets to big, but I think this is fine for now.

src/amberscript/parser.cc
hevrard: Good point, I also added name of unknown buffer if the first buffer argument is unknown

PR #447, opened 2019-04-04
[amberscript] Fix BUFFER commands in examples

jrprice:
Replace TYPE with DATA_TYPE as per the specification.


dj2:
Thank you. I've been trying to keep them up to date, but with the format in flux some things slip through the cracks.

PR #448, opened 2019-04-05
Generate a Format from a DataBuffer.

dj2:
When sending a VertexBuffer to Vulkan the data format is passed as a
VkFormat string. When the buffer is provided in AmberScript it is not a
FORMAT buffer. This CL adds the necessary code to generate a Format from
a DatumType. This will allow creating the necessary format to send to
Vulkan.


src/datum_type.cc
```diff
@@ -52,4 +56,24 @@ uint32_t DatumType::SizeInBytes() const {
   return bytes;
 }
 
+std::unique_ptr<Format> DatumType::AsFormat() const {
+  uint32_t bits_per_element = ElementSizeInBytes() * 8;
+  static const char* prefixes = "RGBA";
```
dneto0: We're trusting that row_count_ is at most 4.  Since the data is already validated at parse time, this is ok.
I would normally prefer a bounds assert.

PR #450, opened 2019-04-05
 [vulkan] Allow setting vertex buffer with a data buffer 

dj2:
This CL updates the vulkan engine to allow using a data buffer when
setting the vertex buffer. The needed format string will be
auto-generated from the data type set in the data buffer.


src/buffer.cc
```diff
@@ -340,7 +340,7 @@ Result FormatBuffer::CopyData(const std::vector<Value>& data) {
       continue;
     }
 
-    const auto& components = format_->GetComponents();
+    const auto components = format_->GetComponents();
```
dneto0: Why not keep this a const ref?

dj2: Accidentally dropped. I'd changed GetComponents but things went sideways and I changed it back and missed this one ....

dj2: done

PR #451, opened 2019-04-05
[vulkan] copy fb data onto device

dj2:
This adds the code to copy the framebuffer from the host to the device before executing the draw commands. This allows using the same buffer over multiple pipelines.


dj2:
Opened too soon ...

PR #452, opened 2019-04-05
[vulkan] copy fb data onto device

dj2:
This CL copies the framebuffer data from the host to the device before each draw/clear command. This allows sharing a framebuffer over multiple pipelines.


dj2:
This CL allows it to work ... it doesn't work yet, the next CL in the chain has a test as it all works. We needed to add DERIVE_PIPELINE first. CL incoming once I fix the formatting errors here....

PR #454, opened 2019-04-08
Fix integration with Dawn

dneto0:
Building on #440, adjust to using 64bit offsets in the Dawn API. (That is new in Dawn.

### DIFF
--- a/samples/config_helper_dawn.cc
+++ b/samples/config_helper_dawn.cc
@@ -16,12 +16,17 @@
 #include <iostream>
 #include "samples/dawn_device_metal.h"
 namespace sample {
+
 ConfigHelperDawn::ConfigHelperDawn() = default;
 ConfigHelperDawn::~ConfigHelperDawn() = default;
-namespace {  // Callback which prints a message from a Dawn device operation.
+
+namespace {
+
+// Callback which prints a message from a Dawn device operation.
 void PrintDeviceError(const char* message, ::dawn::CallbackUserdata) {
   std::cout << "Device error: " << message << std::endl;
 }
+
 }  // namespace
 
 amber::Result ConfigHelperDawn::CreateConfig(
--- a/src/dawn/engine_dawn.cc
+++ b/src/dawn/engine_dawn.cc
@@ -118,7 +118,7 @@ Result MakeFramebufferBuffer(const ::dawn::Device& device,
 struct MapResult {
   Result result;
   const void* data = nullptr;
-  int dataLength = 0;
+  uint64_t dataLength = 0;
 };
 
 // Handles the update from an asynchronous buffer map request, updating the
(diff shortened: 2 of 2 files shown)


samples/config_helper_dawn.cc
```diff
@@ -23,8 +23,7 @@ ConfigHelperDawn::ConfigHelperDawn() = default;
 
 ConfigHelperDawn::~ConfigHelperDawn() = default;
 
-namespace {
-// Callback which prints a message from a Dawn device operation.
+namespace {  // Callback which prints a message from a Dawn device operation.
```
dj2: Move the comment down to above the method. I thought it was a namespace comment initially.

samples/config_helper_dawn.cc
```diff
@@ -43,12 +42,19 @@ amber::Result ConfigHelperDawn::CreateConfig(
   auto r = dawn::CreateMetalDevice(&dawn_instance_, &dawn_device_);
   if (!r.IsSuccess())
     return r;
-#else
-  return amber::Result("Can't make Dawn engine config");
+#else  // assuming VULKAN
+  dawn_instance_.DiscoverDefaultAdapters();
```
dj2: Did we determine if the metal code above could move into this for loop?

src/dawn/engine_dawn.cc
```diff
@@ -286,12 +291,48 @@ Result EngineDawn::DoClear(const ClearCommand*) {
...
+
+  queue_.Submit(1, &command_buffer);
+
+  auto* pipeline = command->GetPipeline();
+  const std::vector<amber::Pipeline::BufferInfo>& out_color_attachment =
+      pipeline->GetColorAttachments();
+
+  // And any further commands start afresh.
+  DestroyCommandBufferBuilder();
+  MapResult map = MapBuffer(*device_, fb_buffer);
```
dj2: We Unmap the fb_buffer on 312 and then re-map here? Can we just keep the map above around?

src/dawn/engine_dawn.cc
```diff
@@ -286,12 +291,48 @@ Result EngineDawn::DoClear(const ClearCommand*) {
...
+
+  for (size_t i = 0; i < 1; ++i) {
+    auto& img = map;
+    auto& info = out_color_attachment[i];
+    auto* values = info.buffer->ValuePtr();
+    values->resize(info.buffer->GetSizeInBytes());
+    // assert map.size == info->buffer->GetSizeInBytes()
```
dj2: instead of assert:

```
if (map.size != info->buffer->GetSizeInBytes())
  return Result("Dawn: framebuffer size does not match buffer size");
```

samples/config_helper_dawn.cc
sarahM0: It doesn't work right now and we decided to leave it as it is.

src/dawn/engine_dawn.cc
dneto0: Next PR will rework how mapping/unmapping occurs.  This buffer will be mapped only within the lifetime of this DoClear call, so it will be assumed to be unmapped at the beginning, then mapped temporarily, copied out of, then forcibly unmapped at the end again.

I'll make that change in the next PR.

src/dawn/engine_dawn.cc
dneto0: Fixed with a TODO, since life is more complicated than this.

PR #455, opened 2019-04-09
Dawn engine: refactor

dneto0:
- Eliminate lots of state from the Dawn engine
  - The Dawn Queue
    - Create and destroy within a single Amber command execution instead.
  - The Dawn CommandBufferEncoder:
    - Create and destroy within a single Amber command execution instead.
- Map an Amber Pipeline object to a Dawn-engine Pipeline object
- Use the framebuffer width and height from the Amber pipeline object,
  instead of hardcoding it.
- The framebuffer buffer is only mapped during the execution of an Amber
  command, then immediately unmapped afterward.

### DIFF
--- a/src/dawn/engine_dawn.cc
+++ b/src/dawn/engine_dawn.cc
@@ -19,6 +19,7 @@
 #include <cstdint>
 #include <cstring>
 #include <iostream>
+#include <string>
 #include <utility>
 #include <vector>
 
@@ -36,8 +37,6 @@ namespace {
 // The minimum multiple row pitch observed on Dawn on Metal.  Increase this
 // as needed for other Dawn backends.
 const uint32_t kMinimumImageRowPitch = 256;
-const auto kFramebufferFormat = ::dawn::TextureFormat::R8G8B8A8Unorm;
-const auto kAmberFramebufferFormatType = FormatType::kR8G8B8A8_UNORM;
 
 // Creates a device-side texture for the framebuffer, and returns it through
 // |result_ptr|.  Assumes the device exists and is valid.  Assumes result_ptr
@@ -87,8 +86,12 @@ Result MakeFramebufferBuffer(const ::dawn::Device& device,
   assert(size_ptr);
 
   // TODO(dneto): Handle other formats.
-  if (format != ::dawn::TextureFormat::R8G8B8A8Unorm) {
-    return Result("Dawn::MakeFramebufferBuffer: Unhandled framebuffer format");
+  // Actually, even better would be to pass the ::amber::Format down here,
+  // and use its GetSizeInBytes() method to determine the texel size in bytes.
+  if (format != ::dawn::TextureFormat::B8G8R8A8Unorm) {
+    return Result(
+        "Dawn::MakeFramebufferBuffer: Unhandled framebuffer format: " +
+        std::to_string(static_cast<uint32_t>(format)));
   }
   // Number of bytes for each texel in the default format.
   const uint32_t default_texel_bytes = 4;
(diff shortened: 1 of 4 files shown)


dneto0:
Hopefully fixed the clang-format check, which was formatting differences and also an include-what-you-use violation

src/dawn/engine_dawn.cc
```diff
@@ -259,86 +260,103 @@ Result EngineDawn::SetShader(ShaderType type,
...
+  // TODO(dneto): Likely, we can create the render objects during
+  // CreatePipeline.
+  Result result = CreateRenderObjectsIfNeeded(render_pipeline);
   if (!result.IsSuccess())
     return result;
 
-  // TODO(sarahM0): Refactor this code later.  This backend is in transition
-  // between the old design and the new design of how Amber operates.
+  // TODO(sarahM0): Adapt this to the case where there might be many colour
+  // attachments. Also whne there is a depth/stencil attachment.
```
dj2: nit: when

src/dawn/engine_dawn.cc
```diff
@@ -417,64 +437,52 @@ Result EngineDawn::DoBuffer(const BufferCommand*) {
...
-        "EngineDawn: Can't create command buffer builder: device is not "
-        "initialized");
-  }
-  command_buffer_builder_ = device_->CreateCommandEncoder();
-  if (command_buffer_builder_)
-    return {};
-  return Result("EngineDawn: Can't create command buffer builder");
+Result EngineDawn::CreateRenderObjectsIfNeeded(
+    RenderPipelineInfo* render_pipeline) {
+  return CreateFramebufferIfNeeded(render_pipeline);
```
dj2: Do we expect to do more here? Or else, why not just call CreateFramebufferIfNeeded directly to make it clearer whta's happening?

src/dawn/engine_dawn.cc
```diff
@@ -417,64 +437,52 @@ Result EngineDawn::DoBuffer(const BufferCommand*) {
...
 
-Result EngineDawn::CreateFramebufferIfNeeded() {
-  Result result;
+  // First make the Dawn color attachment textures that the render pipeline will
+  // write into.
   {
     ::dawn::Texture fb_texture;
-    result = MakeFramebufferTexture(*device_, kFramebufferFormat, &fb_texture);
+
+    result = MakeFramebufferTexture(*device_, kFramebufferFormat, width, height,
```
dj2: The framebuffer format should come from the pipeline, the format is specified in the script.

src/dawn/engine_dawn.h
```diff
@@ -61,31 +61,32 @@ class EngineDawn : public Engine {
   Result DoBuffer(const BufferCommand* cmd) override;
 
  private:
-  // Creates a command buffer builder if it doesn't already exist.
-  Result CreateCommandBufferBuilderIfNeeded();
-  // Destroys the current command buffer builder.
-  void DestroyCommandBufferBuilder();
+  // Returns the Dawn-specific render pipepline for the given command,
```
dj2: nit: pipeline

src/dawn/pipeline_info.h
```diff
@@ -62,12 +66,20 @@ struct RenderPipelineInfo {
 
 struct ComputePipelineInfo {
   ComputePipelineInfo() {}
-  explicit ComputePipelineInfo(::dawn::ShaderModule comp)
-      : compute_shader(comp) {}
+  explicit ComputePipelineInfo(::amber::Pipeline* the_pipeline,
```
dj2: explicit is only needed if there is only one parameter

src/dawn/engine_dawn.cc
dneto0: done in most recent commit.  This is the biggest change to what you've already seen.

src/dawn/pipeline_info.h
dneto0: removed.

dneto0:
Hey @dj2 and @sarahM0 this is ready for re-review.   Getting the framebuffer format from the Amber pipeline object was kind of a big change.

This is also rebased against latest master.

PR #458, opened 2019-04-10
Add Buffer::GetFormat

dj2:
This CL adds a GetFormat method directly to the Buffer class. The method
is overriden by the two sub-classes. The DataBuffer will cache the
generated format.


dj2:
For context, this chain ends with only a single Buffer class. Data and Format buffers are merged ....

PR #459, opened 2019-04-10
AmberScript and README updates

dj2:


docs/amber_script.md
```diff
@@ -118,8 +116,6 @@ BUFFER <name> DATA_TYPE <type> SIZE <size_in_items> <initializer>
 BUFFER <name> FORMAT <format_string>
 ```
 
-TODO(dsinclair): Does framebuffer need a format attached to it?
```
dj2: framebuffer doesn't exist anymore.

docs/amber_script.md
```diff
@@ -98,8 +98,6 @@ either image buffers or, what the target API would refer to as a buffer.
  * vec[2,3,4]\<type>
  * mat[2,3,4]x[2,3,4]\<type>    -- useful?
 
-TODO(dneto): Support half-precision floating point.
```
dj2: An upcoming CL allows this to be a image format as well. So this could be done with R16_SFLOAT as the format.

docs/amber_script.md
```diff
@@ -207,8 +203,6 @@ The following commands are all specified within the `PIPELINE` command.
  * uniform
  * storage
 
-TODO(dsinclair): Reserve `input` as a buffer type for input attachments.
```
dj2: Colour attachments aren't a buffer type, they're done with ATTACH now.

docs/amber_script.md
```diff
@@ -348,9 +342,7 @@ CLEAR <pipeline>
 
 #### Comparators
  * EQ
- * EQ\_FLOAT
```
dj2: I'm not sure these are useful, so removing for now.

docs/amber_script.md
dneto0: Input attachments are not colour attachments.  They are another kind of attachment with their own index space.  See InputAttachmentIndex decoration in SPIR-V.

But point taken, the syntax should be something like:

    BIND BUFFER  myInputAttachment AS input LOCATION 0

(reusing LOCATION here, but could just as well be INDEX.)

PR #464, opened 2019-04-10
Convert ProbeSSBO to use Format objects.

dj2:
This CL moves the ProbeSSBO command to use a format object instead of a
Datum type.

### DIFF
--- a/src/verifier.cc
+++ b/src/verifier.cc
@@ -25,6 +25,7 @@
 namespace amber {
 namespace {
 
+const uint8_t kBitsPerByte = 8;
 const double kEpsilon = 0.000001;
 const double kDefaultTexelTolerance = 0.002;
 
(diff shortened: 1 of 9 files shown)


src/verifier.cc
```diff
@@ -647,8 +647,10 @@ Result Verifier::ProbeSSBO(const ProbeSSBOCommand* command,
                                 "while expected data is not");
   }
 
-  const auto& datum_type = command->GetDatumType();
-  size_t bytes_per_elem = datum_type.ElementSizeInBytes();
+  auto* fmt = command->GetFormat();
+  // TODO(dsinclair): This assumes that all components are the same number
+  // of bits which may not alway shold.
```
hevrard: Typo: ... alway shold -> always hold

src/verifier.cc
```diff
@@ -647,8 +647,10 @@ Result Verifier::ProbeSSBO(const ProbeSSBOCommand* command,
...
+  size_t bytes_per_elem = fmt->GetComponents()[0].num_bits / 8;
```
hevrard: Should the '8' number litteral be replaced by a const?

src/verifier.cc
```diff
@@ -647,8 +647,10 @@ Result Verifier::ProbeSSBO(const ProbeSSBOCommand* command,
                                 "while expected data is not");
   }
 
-  const auto& datum_type = command->GetDatumType();
-  size_t bytes_per_elem = datum_type.ElementSizeInBytes();
+  auto* fmt = command->GetFormat();
```
hevrard: For the sake of clarity, maybe rename this variable to "format" ?

But I can see "fmt" used in other places... Also, beyond this PR, both "cmd" and "command" are used for variable names. Should we be strict on using explicit variable names?

src/verifier.cc
dj2: done

src/verifier.cc
dj2: I've been going with fmt mostly just because it's shorter to type. I also typically do cmd for command, although sometimes it got typed out fully. I'm not usually too strict on variable naming, I think this shouldn't be to confusing.

src/verifier.cc
dj2: done

PR #466, opened 2019-04-11
Merge Data and Format buffers.

dj2:
This Cl removes the two buffer subclasses and uses the buffer class
directly everywhere. This removes the restriction on what types of
things can be done with a given buffer.

### DIFF
--- a/src/buffer.h
+++ b/src/buffer.h
@@ -28,9 +28,6 @@
 
 namespace amber {
 
-class DataBuffer;
-class FormatBuffer;
-
 /// Types of buffers which can be created.
 enum class BufferType : int8_t {
   /// Unknown buffer type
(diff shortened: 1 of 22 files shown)


src/buffer.h
```diff
@@ -156,67 +150,15 @@ class Buffer {
...
-  uint32_t GetTexelStride() { return format_->SizeInBytes(); }
-
-  // When copying the image to the host buffer, we specify a row length of 0
-  // which results in tight packing of rows.  So the row stride is the product
-  // of the texel stride and the number of texels in a row.
-  uint32_t GetRowStride() { return GetTexelStride() * GetWidth(); }
-
- private:
-  Result CopyData(const std::vector<Value>& data);
+  std::vector<uint8_t> values_;
```
dneto0: values_ is a deceptive name because we have a Value class which is a different thing.
Should this be bytes_ ?

src/format.cc
```diff
@@ -22,7 +24,17 @@ Format::Format(const Format&) = default;
 
 Format::~Format() = default;
 
-uint32_t Format::SizeInBytes() const {
+std::unique_ptr<Format> Format::Clone() const {
+  auto ret = MakeUnique<Format>();
```
dneto0: why not use the copy constructor?

`*ret = *this;`

Or even better:
`return MakeUnique<Format>(*this); ` 

Why do we need a clone method at all?

PR #470, opened 2019-04-15
Move API back to C++11.

dj2:
This CL adds C++11 back into the Amber API. This will cause breakage
when rolling Amber due to the changed enum names and the change of the
config object to a unique_ptr.


dj2:
ping

dj2:
Going to close this. I'll look into reviving it later.

PR #479, opened 2019-04-17
Enable syntax highlighting in amberscript doc

dj2:
Pretend that our syntax is the same as the Groovy language. This gives some, slightly weird, syntax highlighting in the code blocks.


dj2:
What do you think? You can use the 'View File' button to see what the highlighting looks like.

PR #482, opened 2019-04-18
Fix format sizes due to layout.

dj2:
The STD140 and 430 layouts were not handled properly by the format code.
This CL extends the code to add the correct format padding for the
layouts and updates the buffer code to attempt to read the expected
number of values and handle the padding as needed.

### DIFF
--- a/src/format_test.cc
+++ b/src/format_test.cc
@@ -55,8 +55,10 @@ TEST_F(FormatTest, SizeInBytesVector) {
   auto fmt = fp.Parse("R32G32B32_SFLOAT");
   ASSERT_TRUE(fmt != nullptr);
 
-  EXPECT_EQ(12U, fmt->SizeInBytes());
-  EXPECT_EQ(12U, fmt->SizeInBytesPerRow());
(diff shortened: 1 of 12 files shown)


dj2:
I _think_ this is right, but please make sure my values used in the FormatStdTest make sense.

src/format_test.cc
```diff
@@ -77,4 +93,57 @@ TEST_F(FormatTest, RowCount) {
...
 }
 
+struct StdData {
+  const char* name;
+  const char* fmt;
+  uint32_t column_count;
+  bool is_std140;
+  uint32_t values_per_element;
+  uint32_t size_in_bytes_per_row;
+  uint32_t size_in_bytes;
```
sarahM0: I think "size_in_bytes == values_per_element * sizeOf(float)"
And "size_in_bytes_per_row = size_in_bytes / column_count".
In that case could we calculate them instead of writing the values?

dj2: My preference for tests is to be explicit and wordy. So, this looks terrible but is very explicit on what it's expecting.

PR #484, opened 2019-04-20
Move buffer filling code into Buffer class.

dj2:
This CL moves the code to fill a buffer with a vector of Value objects
into the Buffer class.


tests/cases/multiple_ubo_update_with_graphics_pipeline.vkscript
```diff
@@ -92,10 +92,12 @@ uniform ubo 1:0 float 0 0.1  0 0 0 \
                         0.8  0 0 0 \
                         0.9  0 0 0 \
                         0.10 0 0 0 \
-                        0.11
+                        0.11 0 0 0
```
dj2: Extra zeros are just cosmetic, make the test read better, but would have been inserted anyway

tests/cases/multiple_ubo_update_with_graphics_pipeline.vkscript
```diff
@@ -115,27 +118,12 @@ draw arrays TRIANGLE_LIST 0 6
...
-                        0.3  0 0 0 \
-                        0.4  0 0 0 \
-                        0.5  0 0 0 \
-                        0.6  0 0 0 \
-                        0.7  0 0 0 \
-                        0.8  0 0 0 \
-                        0.9  0 0 0 \
-                        0.10 0 0 0 \
-                        0.11
+                       0.3 0.2 0.1  0.4
```
dj2: Third row of zeros wasn't used, the data is 2 vec4's in the shader

tests/cases/multiple_ubo_update_with_graphics_pipeline.vkscript
```diff
@@ -115,27 +118,12 @@ draw arrays TRIANGLE_LIST 0 6
 relative probe rect rgba (0.0, 0.0, 1.0, 1.0) (1.0, 0.75, 0.625, 0.5)
 
 uniform ubo 0:0 vec4 0 0.5 0.5 0.25 1.0 \
-                       0.3 0.2 0.1  0.4 \
-                       0   0   0    0
-
-# SPIRV generated by glslang uses std140 layout as a default for UBO
-uniform ubo 1:0 float 0 0.1  0 0 0 \
```
dj2: This is setting the same values as before, so skip it.

tests/cases/multiple_ubo_update_with_graphics_pipeline.vkscript
```diff
@@ -162,8 +150,8 @@ probe ssbo float 2:3 0  ~= 0.1 0.2  0.3  0.4 \
                            0.5 0.6  0.7  0.8 \
                            0.9 0.10 0.11
 
+uniform ubo 0:0 vec4  0  0.5 0.5 0.35 0.6 \
```
dj2: This is the one change. The previous code updated in the middle of an element to half way through the second element. I changed Amber to dis-allow this type of operation, it now always requires a full item to be updated. This makes things a lot simpler internally (and this is a really confusing thing to do in a test)

PR #493, opened 2019-04-24
[amberscript] Add instance and device extensions

### LINKED ISSUE #346: Allow setting instance and device extensions
Need to extend the amber script grammar to allow parsing of extensions for both device and instance level.

dj2:
This CL adds support for parsing instance and device extensions in
amberscript test files.

Fixes #346

### DIFF
--- a/src/amberscript/parser_extension_test.cc
+++ b/src/amberscript/parser_extension_test.cc
@@ -1,0 +1,139 @@
+// Copyright 2019 The Amber Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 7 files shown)


src/amberscript/parser_extension_test.cc
```diff
@@ -0,0 +1,139 @@
...
+
+  Parser parser;
+  Result r = parser.Parse(in);
+  ASSERT_FALSE(r.IsSuccess());
+
+  EXPECT_EQ("1: INSTANCE_EXTENSION missing name", r.Error());
+}
+
+TEST_F(AmberScriptParserTest, ExtensionInstanceInvalidName) {
+  std::string in = "INSTANCE_EXTENSION 1234";
```
dneto0: I guess it's ok to prevent numbers as names.   That's been the pattern and I don't expect it to change.
But we can keep an eye on this assumption.

PR #494, opened 2019-04-24
[amberscript] Add INDEXED parsing

### LINKED ISSUE #473: Parse INDEXED flag to DRAW_ARRAYS
Add parsing for INDEXED DRAW_ARRAY commands.

dj2:
This CL adds parsing of the INDEXED parameter to DRAW_ARRAYS. This
amends the spec slightly to move the INDEXED after the topology. This
better matches with other DRAW_ARRAY commands.

Fixes #473

### DIFF
--- a/src/amberscript/parser_run_test.cc
+++ b/src/amberscript/parser_run_test.cc
@@ -619,11 +619,17 @@ BUFFER vtex_buf DATA_TYPE vec3<float> DATA
 4 5 6
 7 8 9
 END
+BUFFER idx_buf DATA_TYPE vec3<float> DATA
+9 8 7
+6 5 4
+3 2 1
+END
 
 PIPELINE graphics my_pipeline
   ATTACH my_shader
   ATTACH my_fragment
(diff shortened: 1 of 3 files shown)


src/amberscript/parser_run_test.cc
```diff
@@ -608,6 +608,46 @@ RUN my_pipeline DRAW_ARRAY AS TRIANGLE_LIST)";
...
+7 8 9
+END
+
+PIPELINE graphics my_pipeline
+  ATTACH my_shader
+  ATTACH my_fragment
+  VERTEX_DATA vtex_buf LOCATION 0
+END
+
+RUN my_pipeline DRAW_ARRAY AS TRIANGLE_LIST INDEXED)";
```
dneto0: The pipeline doesn't actually have an INDEX_DATA.  It seems we could validate that and return an error if the pipeline doesn't actually have an index buffer attached to it.

dj2: Good idea. Done.

PR #495, opened 2019-04-29
[vkscript] Store push constant data in a buffer.

dj2:
Currently the push constant data is stored in a vector of Value objects.
It is also possible to store push constant data in a buffer. This CL
converts the VkScript backend to use a buffer instead of Values and
removes the option of using Values.


src/vkscript/command_parser_test.cc
```diff
@@ -3340,11 +3340,12 @@ TEST_F(CommandParserTest, Uniform) {
   EXPECT_EQ(1U, fmt->ColumnCount());
   EXPECT_EQ(3U, fmt->RowCount());
 
-  const auto& values = cmd->GetValues();
-  std::vector<float> results = {2.1f, 3.2f, 4.3f};
-  ASSERT_EQ(results.size(), values.size());
+  const auto* buf = cmd->GetBuffer();
+  const auto* values = buf->GetValues<float>();
+  std::vector<float> results = {2.1f, 3.2f, 4.3f, 0.f};
+  ASSERT_EQ(results.size(), buf->ValueCount());
```
dneto0: FYI this is where EXPECT_THAT(expression, Eq(expected)) will help compare containers.

PR #496, opened 2019-05-06
Exclude spirv-tools targets from all.

s-perron:
When asking the build environment to build all, everything in
spirv-tools is being build.  There are some options that turn off a few
things, but that is harder to keep track of.  This change will simply
exclude everything in spirv-tools from all, and the only targets that will
be built are those that amber depends on.


s-perron:
This reduces the number to targets built from 437 to 349.

Issue #16, opened 2018-11-14
Dawn: Implement DoDrawRect


dj2:
Add support to Dawn engine for the DoDrawRect command.

PR #498, opened 2019-05-06
Add support for DoDrawRect in dawn backend

sarahM0:
- Implements #16 
- Add support for drawRect in Dawn backend
- Update kokoro build for Dawn to use a fixed old version while Dawn creators are refactoring their work
- Eliminated the need for module_for_typ_ state form EngineDawn by merging two functions
- Added a temporary Helper struct to EngineDawn
- struct Helper is used to create and initialize two dawn objects (RenderPassDescriptor and RenderPipelineDescriptor)


src/dawn/engine_dawn.cc
```diff
@@ -151,11 +152,20 @@ void HandleBufferMapCallback(DawnBufferMapAsyncStatus status,
   }
 }
 
+uint32_t Align(uint32_t value, size_t alignment) {
+  assert(alignment <= UINT32_MAX);
```
dj2: s/UINT32_MAX/std::numeric_limits<uint32_t>::max()/

src/dawn/engine_dawn.cc
```diff
@@ -151,11 +152,20 @@ void HandleBufferMapCallback(DawnBufferMapAsyncStatus status,
...
+}
+}  // namespace
+
 // Maps the given buffer.  Assumes the buffer has usage bit
-// dawn::BufferUsageBit::MapRead set.  Returns a MapResult structure, with the
-// status saved in the .result member and the host pointer to the mapped data
-// in the |.data| member. Mapping a buffer can fail if the context is lost, for
-// example. In the failure case, the .data member of the result will be null.
+// ::dawn::BufferUsageBit::MapRead set.  Returns a MapResult structure, with
+// the status saved in the .result member and the host pointer to the mapped
```
dj2: nit: Remove the . before member names and just have |result|

src/dawn/engine_dawn.cc
```diff
@@ -179,8 +189,253 @@ MapResult MapBuffer(const ::dawn::Device& device, const ::dawn::Buffer& buf) {
...
+
+  ::dawn::Extent3D copySize = {width, height, 1};
+
+  auto encoder = device.CreateCommandEncoder();
+  encoder.CopyTextureToBuffer(&textureCopyView, &bufferCopyView, &copySize);
+
+  auto commands = encoder.Finish();
+  auto queue = device.CreateQueue();
+  queue.Submit(1, &commands);
+
```
dj2: Maybe wrap lines 201-226 inside a block to make it clear all those things go out of scope at that point?

src/dawn/engine_dawn.cc
```diff
@@ -179,8 +189,253 @@ MapResult MapBuffer(const ::dawn::Device& device, const ::dawn::Buffer& buf) {
...
+  ::dawn::Buffer buffer = device.CreateBuffer(&descriptor);
+  buffer.SetSubData(0, size, reinterpret_cast<const uint8_t*>(data));
+  return buffer;
+}
+
+::dawn::SamplerDescriptor GetDefaultSamplerDescriptor() {
+  ::dawn::SamplerDescriptor desc;
+
+  float kLodMin = 0.0;
```
dj2: Move these up to the anonymous namespace at the top of the file if they are constants. (and add a const at the beginning.)

src/dawn/engine_dawn.cc
```diff
@@ -179,8 +189,253 @@ MapResult MapBuffer(const ::dawn::Device& device, const ::dawn::Buffer& buf) {
...
+  const auto fb_buffer_size = dawn_row_pitch * height + width * pixelSize;
+
+  ::dawn::BufferDescriptor descriptor;
+  descriptor.size = fb_buffer_size;
+  descriptor.usage =
+      ::dawn::BufferUsageBit::MapRead | ::dawn::BufferUsageBit::TransferDst;
+  ::dawn::Buffer fb_buffer = device.CreateBuffer(&descriptor);
+
+  ::dawn::TextureCopyView textureCopyView;
+  textureCopyView.texture = render_pipeline.fb_texture;
```
dj2: Can you call CreateTextureCopyView?

src/dawn/engine_dawn.cc
```diff
@@ -179,8 +189,253 @@ MapResult MapBuffer(const ::dawn::Device& device, const ::dawn::Buffer& buf) {
...
+  textureCopyView.level = 0;
+  textureCopyView.slice = 0;
+  textureCopyView.origin = {0, 0, 0};
+
+  ::dawn::BufferCopyView bufferCopyView;
```
dj2: Call CreateBufferCopyView?

src/dawn/engine_dawn.cc
```diff
@@ -179,8 +189,253 @@ MapResult MapBuffer(const ::dawn::Device& device, const ::dawn::Buffer& buf) {
...
+  }
+  return device.CreatePipelineLayout(&descriptor);
+}
+
+::dawn::TextureView CreateDefaultDepthStencilView(
+    const ::dawn::Device& device,
+    const ::dawn::TextureFormat depth_stencil_format) {
+  ::dawn::TextureDescriptor descriptor;
+  descriptor.dimension = ::dawn::TextureDimension::e2D;
+  descriptor.size.width = 250;
```
dj2: Follupwup: This width and height should come from the framebuffer size in the pipeline.

Maybe add a TODO for now?

src/dawn/engine_dawn.cc
```diff
@@ -179,8 +189,253 @@ MapResult MapBuffer(const ::dawn::Device& device, const ::dawn::Buffer& buf) {
...
+      return shaderc_glsl_vertex_shader;
+  }
+}
+
+::dawn::ShaderModule CreateShaderModule(const ::dawn::Device& device,
+                                        ::dawn::ShaderStage stage,
+                                        const char* source) {
+  shaderc_shader_kind kind = ShadercShaderKind(stage);
+
+  shaderc::Compiler compiler;
```
dj2: This should not need to call shaderc, the shader module is already compiled by the time we create the framework. The data is in the shader info.

src/dawn/engine_dawn.cc
```diff
@@ -179,8 +189,253 @@ MapResult MapBuffer(const ::dawn::Device& device, const ::dawn::Buffer& buf) {
...
+  auto result =
+      compiler.CompileGlslToSpv(source, strlen(source), kind, "myshader?");
+  if (result.GetCompilationStatus() != shaderc_compilation_status_success) {
+    std::cerr << result.GetErrorMessage();
+    return {};
+  }
+#ifdef DUMP_SPIRV_ASSEMBLY
```
dj2: I'd remove these (although this one might be useful to add as a general param and hook into ShaderCompiler.) Maybe file a bug to add?

src/dawn/engine_dawn.cc
```diff
@@ -249,15 +502,33 @@ Result EngineDawn::Initialize(EngineConfig* config,
...
+    const std::vector<uint32_t>& code = shader_info.GetData();
+    descriptor.code = code.data();
+    descriptor.codeSize = uint32_t(code.size());
+
+    auto shader = device_->CreateShaderModule(&descriptor);
+    if (!shader) {
+      return Result("Dawn::SetShader: failed to create shader");
+    }
+    if (module_for_type.count(type)) {
+      Result("Dawn::SetShader: module for type already exists");
```
dj2: missing a return?

src/dawn/engine_dawn.cc
```diff
@@ -471,67 +653,247 @@ Result EngineDawn::DoClear(const ClearCommand* command) {
...
-  }
-
-  result = CreateCommandBufferBuilderIfNeeded();
+  auto* amber_format =
+      render_pipeline.pipeline->GetColorAttachments()[0].buffer->GetFormat();
+  if (!amber_format)
+    return Result("Color attachment 0 has no format!");
+  ::dawn::TextureFormat fb_format{};
+  GetDawnTextureFormat(*amber_format, &fb_format);
   if (!result.IsSuccess())
```
dj2: Result hasn't been used yet, do we want to remove this? If so, move the Reuslt on 659 down to 676 where it's used.

src/dawn/engine_dawn.cc
```diff
@@ -471,67 +653,247 @@ Result EngineDawn::DoClear(const ClearCommand* command) {
...
-        "any graphics pipeline commands?");
-  }
-
-  result = CreateCommandBufferBuilderIfNeeded();
+  auto* amber_format =
+      render_pipeline.pipeline->GetColorAttachments()[0].buffer->GetFormat();
+  if (!amber_format)
+    return Result("Color attachment 0 has no format!");
+  ::dawn::TextureFormat fb_format{};
+  GetDawnTextureFormat(*amber_format, &fb_format);
```
dj2: Or, should this be doing result = here?

src/dawn/engine_dawn.cc
```diff
@@ -471,67 +653,247 @@ Result EngineDawn::DoClear(const ClearCommand* command) {
...
+    depth_stencil_format = ::dawn::TextureFormat::D32FloatS8Uint;
+  }
+  renderPipelineDescriptor.layout = MakeBasicPipelineLayout(device, nullptr);
+  renderPipelineDescriptor.primitiveTopology =
+      ::dawn::PrimitiveTopology::TriangleList;
+  renderPipelineDescriptor.sampleCount = 1;
+
+  // Set defaults for the vertex stage descriptor.
+  cVertexStage.module = render_pipeline.vertex_shader;
+  cVertexStage.entryPoint = "main";
```
dj2: This needs to come from the shader_info as the entry point is configurable. Maybe add a TODO and a followup bug?

src/dawn/engine_dawn.cc
```diff
@@ -471,67 +653,247 @@ Result EngineDawn::DoClear(const ClearCommand* command) {
...
+  renderPipelineDescriptor.vertexStage = std::move(&cVertexStage);
+
+  // Set defaults for the fragment stage desriptor.
+  cFragmentStage.module = render_pipeline.fragment_shader;
+  cFragmentStage.entryPoint = "main";
```
dj2: ditto

src/dawn/engine_dawn.cc
```diff
@@ -471,67 +653,247 @@ Result EngineDawn::DoClear(const ClearCommand* command) {
...
+Result EngineDawn::Helper::CreateRenderPassDescriptor(
+    const RenderPipelineInfo& render_pipeline,
+    const ::dawn::Device& device) {
+  std::initializer_list<::dawn::TextureView> colorAttachmentInfo = {
+      render_pipeline.fb_texture.CreateDefaultView()};
+
+  for (uint32_t i = 0; i < kMaxColorAttachments; ++i) {
+    mColorAttachmentsInfo[i].loadOp = ::dawn::LoadOp::Clear;
+    mColorAttachmentsInfo[i].storeOp = ::dawn::StoreOp::Store;
+    mColorAttachmentsInfo[i].clearColor = {0.0f, 0.0f, 0.0f, 0.0f};
```
dj2: Add a TODO that this should be the result of the DoClearColor command?

src/dawn/engine_dawn.cc
```diff
@@ -471,67 +653,247 @@ Result EngineDawn::DoClear(const ClearCommand* command) {
...
+    cColorAttachmentsInfoPtr[i] = nullptr;
+  }
 
-    ::dawn::Extent3D extent = {kFramebufferWidth, kFramebufferHeight, depth};
+  cDepthStencilAttachmentInfo.clearDepth = 1.0f;
```
dj2: Add TODO about this being the result of the DoClearDepth command result?

src/dawn/engine_dawn.cc
```diff
@@ -471,67 +653,247 @@ Result EngineDawn::DoClear(const ClearCommand* command) {
...
+  cDepthStencilAttachmentInfo.clearStencil = 0;
```
dj2: Ditto DoClearStencil

src/dawn/engine_dawn.h
```diff
@@ -15,6 +15,8 @@
 #ifndef SRC_DAWN_ENGINE_DAWN_H_
 #define SRC_DAWN_ENGINE_DAWN_H_
 
+#include <shaderc/shaderc.hpp>
```
dj2: This shouldn't be needed. Shaderc should only be known to ShaderCompiler.{h|cpp}

src/dawn/engine_dawn.h
```diff
@@ -69,6 +75,42 @@ class EngineDawn : public Engine {
     return pipeline_map_[command->GetPipeline()].render_pipeline.get();
   }
 
+  // Helper stuff
+  struct Helper {
```
dj2: If this is only used internally, move this struct definition to the anonymous namespace of engine_dawn.cpp

src/dawn/engine_dawn.cc
sarahM0: yes you're right! I fixed them.

src/dawn/engine_dawn.cc
sarahM0: Good catch! I missed saving the outcome of GetDawnTextureFormat in _result_. Now it's being used.

src/dawn/engine_dawn.cc
```diff
@@ -37,6 +39,41 @@ namespace {
 // The minimum multiple row pitch observed on Dawn on Metal.  Increase this
 // as needed for other Dawn backends.
 const uint32_t kMinimumImageRowPitch = 256;
+const float kLodMin = 0.0;
+const float kLodMax = 1000.0;
+
+// Helper stuff
+struct Helper {
+  ::dawn::RenderPassColorAttachmentDescriptor*
+      cColorAttachmentsInfoPtr[kMaxColorAttachments];
```
dneto0: What is the "c" prefix on the name here, and "m" on mColorAttachmentsInfo?

We don't need those prefixes.

src/dawn/engine_dawn.cc
```diff
@@ -37,6 +39,41 @@ namespace {
 // The minimum multiple row pitch observed on Dawn on Metal.  Increase this
 // as needed for other Dawn backends.
 const uint32_t kMinimumImageRowPitch = 256;
+const float kLodMin = 0.0;
+const float kLodMax = 1000.0;
+
+// Helper stuff
+struct Helper {
```
dneto0: Seems like a better name would be DawnPipelineInfo ?

src/dawn/engine_dawn.cc
```diff
@@ -151,11 +188,19 @@ void HandleBufferMapCallback(DawnBufferMapAsyncStatus status,
   }
 }
 
+uint32_t Align(uint32_t value, size_t alignment) {
```
dneto0: Needs a method comment.

// Returns |value| but rounded up to a multiple of |alignment|.  |alignment| is assumed to be a power of 2.

src/dawn/engine_dawn.h
```diff
@@ -29,6 +29,10 @@
 namespace amber {
 namespace dawn {
 
+static constexpr uint32_t kMaxColorAttachments = 4u;
```
dneto0: constexpr doesn't work in VS2013.  I'm not sure if we care to support it or not.

If we need it, an alternative is

enum : uint32_t { kMaxColorAttachments = 4u, kMaxVertexInputs = 16u, kMaxVertexAttributes = 16u };

src/dawn/engine_dawn.cc
```diff
@@ -179,8 +224,173 @@ MapResult MapBuffer(const ::dawn::Device& device, const ::dawn::Buffer& buf) {
   return map_result;
 }
 
+::dawn::BufferCopyView CreateBufferCopyView(::dawn::Buffer buffer,
```
dneto0: Needs a method comment.

src/dawn/engine_dawn.cc
```diff
@@ -179,8 +224,173 @@ MapResult MapBuffer(const ::dawn::Device& device, const ::dawn::Buffer& buf) {
...
+  ::dawn::BufferCopyView bufferCopyView;
+  bufferCopyView.buffer = buffer;
+  bufferCopyView.offset = offset;
+  bufferCopyView.rowPitch = rowPitch;
+  bufferCopyView.imageHeight = imageHeight;
+
+  return bufferCopyView;
+}
+
+::dawn::TextureCopyView CreateTextureCopyView(::dawn::Texture texture,
```
dneto0: needs a method comment.

src/dawn/engine_dawn.cc
```diff
@@ -179,8 +224,173 @@ MapResult MapBuffer(const ::dawn::Device& device, const ::dawn::Buffer& buf) {
...
+  ::dawn::TextureCopyView textureCopyView;
+  textureCopyView.texture = texture;
+  textureCopyView.level = level;
+  textureCopyView.slice = slice;
+  textureCopyView.origin = origin;
+
+  return textureCopyView;
+}
+
+MapResult MapTextureToHostBuffer(const RenderPipelineInfo& render_pipeline,
```
dneto0: Needs a method comment

src/dawn/engine_dawn.cc
```diff
@@ -179,8 +224,173 @@ MapResult MapBuffer(const ::dawn::Device& device, const ::dawn::Buffer& buf) {
...
+                  row_stride);
+    }
+  }
+  // Always unmap the buffer at the end of the engine's command.
+  fb_buffer.Unmap();
+
+  return map;
+}
+
+::dawn::Buffer CreateBufferFromData(const ::dawn::Device& device,
```
dneto0: Needs method comment

src/dawn/engine_dawn.cc
```diff
@@ -249,15 +457,33 @@ Result EngineDawn::Initialize(EngineConfig* config,
...
+    if (module_for_type.count(type)) {
+      return Result("Dawn::SetShader: module for type already exists");
+    }
+    module_for_type[type] = shader;
   }
 
   switch (pipeline->GetType()) {
     case PipelineType::kCompute: {
-      auto module = module_for_type_[kShaderTypeCompute];
+      auto module = module_for_type[kShaderTypeCompute];
```
dneto0: Hm. Now that I read this again I'd recommend making this "auto&"  so we avoid unnecessary copies.

src/dawn/engine_dawn.cc
```diff
@@ -267,15 +493,17 @@ Result EngineDawn::CreatePipeline(::amber::Pipeline* pipeline) {
 
     case PipelineType::kGraphics: {
       // TODO(dneto): Handle other shader types as well.  They are optional.
-      auto vs = module_for_type_[kShaderTypeVertex];
-      auto fs = module_for_type_[kShaderTypeFragment];
+      auto vs = module_for_type[kShaderTypeVertex];
```
dneto0: Same here, make this auto&

src/dawn/engine_dawn.cc
```diff
@@ -471,67 +608,253 @@ Result EngineDawn::DoClear(const ClearCommand* command) {
...
-    texture_copy_view.level = level;
-    texture_copy_view.slice = slice;
-    texture_copy_view.origin = {x, y, z};
+  renderPipelineDescriptor.layout = MakeBasicPipelineLayout(device, nullptr);
+  renderPipelineDescriptor.primitiveTopology =
+      ::dawn::PrimitiveTopology::TriangleList;
+  renderPipelineDescriptor.sampleCount = 1;
+
+  // Set defaults for the vertex stage descriptor.
+  snprintf(entryPoint,
```
dneto0: This seems like an awful lot of work.    Also, where is it writing the data to?  No memory has been allocated to it.
I would change entryPoint to a std::string, and then:

  entryPoint = ... GetEntryPoint()
 ...
  cVertexStage.entryPoint = entryPoint.c_str();

src/dawn/engine_dawn.cc
```diff
@@ -471,67 +608,253 @@ Result EngineDawn::DoClear(const ClearCommand* command) {
...
+           sizeof(render_pipeline.pipeline->GetShaders()[0].GetEntryPoint()),
+           "%s",
+           render_pipeline.pipeline->GetShaders()[0].GetEntryPoint().c_str());
+  cVertexStage.module = render_pipeline.vertex_shader;
+  cVertexStage.entryPoint = entryPoint;
+  renderPipelineDescriptor.vertexStage = &cVertexStage;
+
+  // Set defaults for the fragment stage descriptor.
+  cFragmentStage.module = render_pipeline.fragment_shader;
+  cFragmentStage.entryPoint = entryPoint;
```
dneto0: The fragment shader entry point name can be different.  You need to look it up again.  And I think do that by fragment shader type?

src/dawn/engine_dawn.cc
dneto0: Actually, why is it always the [0] entry of the GetShaders result?   Shouldn't we look it up by shader type?

src/dawn/engine_dawn.cc
```diff
@@ -471,67 +608,253 @@ Result EngineDawn::DoClear(const ClearCommand* command) {
...
+  renderPipelineDescriptor.colorStates = &cColorStates[0];
+
+  /// Set defaults for the depth stencil state descriptors.
+  stencilFace.compare = ::dawn::CompareFunction::Always;
+  stencilFace.failOp = ::dawn::StencilOperation::Keep;
+  stencilFace.depthFailOp = ::dawn::StencilOperation::Keep;
+  stencilFace.passOp = ::dawn::StencilOperation::Keep;
+  cDepthStencilState.format = fb_format;
+  cDepthStencilState.depthWriteEnabled = false;
+  cDepthStencilState.depthCompare = ::dawn::CompareFunction::Always;
```
dneto0: I'm not certain this is right, but ... hey as long as the tests pass.  :-)

src/dawn/engine_dawn.h
sarahM0: No it's not needed I replaced it with const and moved them to the anonymes space in engine_dawn.cc.

src/dawn/engine_dawn.cc
sarahM0: Neither am I :) I've done what Dawn examples do.

src/dawn/engine_dawn.cc
sarahM0: How about DawnPipelineHelper? it's not hold any info, basically it's there so a bunch of variables do not go out of scope while we need them.

src/dawn/engine_dawn.cc
```diff
@@ -36,7 +38,48 @@ namespace {
...
+static const uint32_t kMinimumImageRowPitch = 256;
+static const float kLodMin = 0.0;
+static const float kLodMax = 1000.0;
+static const uint32_t kMaxColorAttachments = 4u;
+static const uint32_t kMaxVertexInputs = 16u;
+static const uint32_t kMaxVertexAttributes = 16u;
+
+// This structure is a container for a few variables that are created during
+// CreateRenderPipelineDescriptor and CreateRenderPassDescriptor and we want to
+// make sure they don't go out of scope before we are done with them
```
dneto0: The comment about the lifetime of the values is insightful.  Lifecycle info is an important aspect that's often overlooked, and the source of bugs.

src/dawn/engine_dawn.h
```diff
@@ -84,9 +84,8 @@ class EngineDawn : public Engine {
   // pipelines.
   std::unordered_map<amber::Pipeline*, ::amber::dawn::Pipeline> pipeline_map_;
 
-  // TODO(dneto): Remove this. Shaders are attached to the pipeline.
   std::unordered_map<ShaderType, ::dawn::ShaderModule, CastHash<ShaderType>>
```
dneto0: This variable should be deleted now.  You've absorbed it into one of the methods.

src/dawn/engine_dawn.cc
```diff
@@ -36,7 +38,48 @@ namespace {
...
+  ::dawn::PipelineStageDescriptor vertexStage;
+  ::dawn::StencilStateFaceDescriptor stencilFace;
+  ::dawn::VertexInputDescriptor vertexInput;
+  std::array<::dawn::VertexInputDescriptor, kMaxVertexInputs> tempInputs;
+  std::array<::dawn::VertexAttributeDescriptor, kMaxVertexAttributes>
+      tempAttributes;
+  ::dawn::VertexAttributeDescriptor vertexAttribute;
+  ::dawn::BlendDescriptor blend;
+  ::dawn::ColorStateDescriptor colorStateDescriptor;
+  std::string vertextEntryPoint;
```
dneto0: Spelling:   "vertexEntryPoint"

PR #500, opened 2019-05-08
Update VkScript buffer document.

### LINKED ISSUE #497: Buffer data forced to use consistent types
The type declaration for each buffer must be consistent across all subdata.  This is a newly enforced rule.  It's not documented.

If ultimately all the scalars use the same basic type then we can fake it by using lowest-common denominator type.  For example, if we want to mix uint with uvec4, then use uint where instead of uvec4.  But we still can't mix float and integer data in a single buffer.

Example:
```
[compute shader]
#version 450

layout(set = 0, binding = 0, std140) uniform UBO {
        uint a[2];
        uvec4 b;
        uvec3 c;
        uvec2 d;
} Ubo;

layout(set = 0, binding = 1, std430) buffer BUF {
        uint a[2];
        uvec4 b;
        uvec3 c;
        uvec2 d;
} Buf;

void main() {
        Buf.a[0] = Ubo.a[0];
        Buf.a[1] = Ubo.a[1];
        Buf.b = Ubo.b;
        Buf.c = Ubo.c;
        Buf.d = Ubo.d;
}

[test]
# Array stride in UBO is a multiple of 16
uniform ubo 0:0 uint 0 1
uniform ubo 0:0 uint 16 2
# Really uvec4, but can fake it if we say "uint" here
uniform ubo 0:0 uvec4 32 3 4 5 6
# Really uvec3, but can fake it if we say "uint" here
uniform ubo 0:0 uvec3 48 7 8 9
# Really uvec2, but can fake it if we say "uint" here
uniform ubo 0:0 uvec2 64 10 11
ssbo 0:1 72

compute 1 1 1

probe ssbo uint 0:1 0 == 1
probe ssbo uint 0:1 4 == 2
probe ssbo uint 0:1 16 == 3 4 5 6
probe ssbo uint 0:1 32 == 7 8 9
probe ssbo uint 0:1 48 == 10 11
```

With this, Amber complains:
ubo_to_ssbo.vkscript: 31: probe ssbo format does not match buffer format

dj2:
This CL updates the VkScript document to make note that buffers must be
given consistent types.

Fixes #497

### DIFF
--- a/docs/vk_script.md
+++ b/docs/vk_script.md
@@ -213,6 +213,9 @@ Sets the push constants at `offset`. The `type` is from the *Data Types*
 section below. The `values` must be a non-zero multiple of the requested
 `type`.
 
+When setting push constant data each call to `uniform` must use the same
+`type` or the script will be rejected as invalid.
+
 
 ### Unifom UBO
  * `uniform ubo _binding_ _type_ _offset_ _values_+`
@@ -220,6 +223,9 @@ section below. The `values` must be a non-zero multiple of the requested
 Sets the values in the uniform buffer at `binding` and `offset`. The `type`
 is from the *Data Types* section below. The `values` must be a non-zero
 multiple of the requested `type`.
+
+When setting data into a single `binding`, each call to `uniform ubo` must
+use the same `type` or the script will be rejected as invalid.
 
 
 ### SSBO size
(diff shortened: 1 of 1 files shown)


dj2:
ping

PR #511, opened 2019-05-15
Add HLSL support through DXC

### LINKED ISSUE #34: AmberScript: Add support for DXC
We currently link in shaderc to compile GLSL shaders. We should add the option to build in DXC as well to compile HLSL shaders.

dj2:
This Cl adds DXC to the Amber third_party directory and enables using
HLSL as a shader language in AmberScript.

Fixes #34


dj2:
Hm, cleaned out my third_party/dxc and this is failing to link again.....will take a look.

dj2:
This requires a patch to DXC in order to build:

```
diff --git a/tools/clang/tools/dxcompiler/CMakeLists.txt b/tools/clang/tools/dxcompiler/CMakeLists.txt
index 5887191c..c9d3d82f 100644
--- a/tools/clang/tools/dxcompiler/CMakeLists.txt
+++ b/tools/clang/tools/dxcompiler/CMakeLists.txt
@@ -113,7 +113,7 @@ set(GENERATED_HEADERS
   ClangStmtNodes
   )
 
-add_clang_library(dxcompiler SHARED ${SOURCES})
+add_clang_library(dxcompiler STATIC ${SOURCES})
 if (WIN32)
   # No DxcEtw on non-Windows platforms.
   add_dependencies(dxcompiler DxcEtw)
```

We'll probably want to make that change contingent on BUILD_SHARED_LIBS being off or LLVM_BUILD_STATIC being on

src/dxc_helper.cc
```diff
@@ -0,0 +1,146 @@
...
+    return Result("DXCCreateInstance for DXCCompiler failed");
+  }
+
+  IDxcOperationResult* result;
+  std::wstring src_filename =
+      L"amber." + std::wstring(profile.begin(), profile.end());
+  if (compiler->Compile(source,               /* source text */
+                        src_filename.c_str(), /* original file source */
+                        std::wstring(entry.begin(), entry.end())
+                            .c_str(), /* entry point name */
```
dneto0: I had to stop for a while, worrying about the lifetime of the temporary.  But it dies at the end of the full expression, so that's late enough.  Phew.

src/dxc_helper.cc
```diff
@@ -0,0 +1,146 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "src/dxc_helper.h"
+
+#include <algorithm>
+#include <sstream>
+
+#include "third_party/dxc/include/dxc/Support/Global.h"
```
dneto0: This seems over-descriptive.  In future we should allow DXC's source to be in some configurable location.  But that's for the future.

README.md
```diff
@@ -113,6 +113,11 @@ The available flags which can be defined are:
 cmake -DAMBER_SKIP_TESTS=True -DAMBER_SKIP_SPIRV_TOOLS=True -GNinja ../..
 ```
 
+#### DXC
+
+DXC can be enabled in Amber by adding the `-DAMBER_USE_DXC=true` flag when
+running cmake.
+
```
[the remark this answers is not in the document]
dneto0: Although we would also need updates to the vk_script and amberscript docs.  So I'll file an issue for that.

dneto0:
> We'll probably want to make that change contingent on BUILD_SHARED_LIBS being off or LLVM_BUILD_STATIC being on

Right, we'll need that change before we can land this.

src/dxc_helper.cc
dj2: Include paths should be from the root src/ directory. Or are you thinking we should allow DXC to live outside amber?

src/dxc_helper.cc
```diff
@@ -0,0 +1,146 @@
...
+}  // namespace
+
+Result Compile(const std::string& src,
+               const std::string& entry,
+               const std::string& profile,
+               std::vector<uint32_t>* generated_binary) {
+  DxcInitThreadMalloc();
+
+  if (hlsl::options::initHlslOptTable()) {
+    DxcCleanupThreadMalloc();
```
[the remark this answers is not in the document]
dj2: Ah, yea, we could do some kind of lifetime scope thing if we want.

README.md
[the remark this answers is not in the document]
dj2: I have no plans on updating vkscript to support hlsl. This will be AmberScript only.

src/dxc_helper.cc
dj2: Heh, I didn't think about that at the time.

src/dxc_helper.cc
dneto0: I mean DXC should be able to live outside of Amber.  Like, if I had one giant build tree that included many tools for the Vulkan shader stack.
So, I'd have 

#include "dxc/Support/Global.h" 

so that I can always find with a -I option on the compiler, where that is controlled from a CMake string expression.

dj2:
It looks like on windows we always want to build DXCompiler as SHARED. Building as STATIC doesn't seem to link for me....

dj2:
Note, this no longer requires any changes into DXC to work. So, it should be good to land once bots + reviewers are happy.

CMakeLists.txt
```diff
@@ -112,6 +118,20 @@ add_definitions(-DAMBER_ENGINE_VULKAN=$<BOOL:${Vulkan_FOUND}>)
...
 add_definitions(-DAMBER_ENABLE_SPIRV_TOOLS=$<BOOL:${AMBER_ENABLE_SPIRV_TOOLS}>)
 add_definitions(-DAMBER_ENABLE_SHADERC=$<BOOL:${AMBER_ENABLE_SHADERC}>)
+add_definitions(-DAMBER_ENABLE_DXC=$<BOOL:${AMBER_ENABLE_DXC}>)
+
+set(CMAKE_DEBUG_POSTFIX "")
+
+# This has to be done very early so the link path will get set correctly for all
+# the various libraries and binaries.
+if (${AMBER_ENABLE_DXC})
+  link_directories("${CMAKE_BINARY_DIR}/third_party/dxc/lib")
```
dneto0: For later:    We can use `<projectname>_BINARY_DIR` to specify this dir, without assuming its source was under third_party.

It's kind of weird, but DXC sets the project name to "LLVM", so then it would still look weird!
See https://github.com/microsoft/DirectXShaderCompiler/blob/master/CMakeLists.txt#L28

src/CMakeLists.txt
```diff
@@ -51,11 +51,31 @@ if (${Dawn_FOUND})
...
+endif()
+
 add_library(libamber ${AMBER_SOURCES})
 amber_default_compile_options(libamber)
 target_include_directories(libamber PRIVATE "${CMAKE_BINARY_DIR}")
 set_target_properties(libamber PROPERTIES OUTPUT_NAME "amber")
 
+if (${AMBER_ENABLE_DXC})
+  target_include_directories(libamber PRIVATE
+    "${PROJECT_SOURCE_DIR}/third_party/dxc/include"
```
dneto0: Similarly, this could be `${LLVM_SOURCE_DIR}/include` and `${LLVM_BINARY_DIR}/include`

PR #512, opened 2019-05-15
Add support for building SwiftShader

dj2:
This CL adds swiftshader into Amber and allows amber to compile with swiftshader if AMBER_ENABLE_SWIFTSHADER is provided. If swiftshader is built, you can set your `VK_ICD_FILENAMES` to point to the generated icd (out/sw/Linux/vk_swiftshader_icd.json for instance) and swiftshader will be used to run the tests.

### DIFF
--- a/src/vulkan/find_vulkan.cmake
+++ b/src/vulkan/find_vulkan.cmake
@@ -19,17 +19,19 @@ set(Vulkan_FOUND FALSE)
 set(VULKAN_CTS_HEADER FALSE)
 set(VULKAN_LIB "")
 
-if (${AMBER_USE_LOCAL_VULKAN})
-  set(Vulkan_FOUND TRUE)
-  set(VulkanHeaders_INCLUDE_DIR
-    ${PROJECT_SOURCE_DIR}/third_party/vulkan-headers/include
-    CACHE PATH "vk headers dir" FORCE)
-  set(VulkanRegistry_DIR
-    ${PROJECT_SOURCE_DIR}/third_party/vulkan-headers/registry
-    CACHE PATH "vk_registry_dir" FORCE)
-  include_directories(BEFORE "${VulkanHeaders_INCLUDE_DIR}")
-  set(VULKAN_LIB vulkan)
-  message(STATUS "Amber: using local vulkan")
+if (NOT ${Vulkan_FOUND})
+  if (${AMBER_USE_LOCAL_VULKAN})
+    set(Vulkan_FOUND TRUE)
+    set(VulkanHeaders_INCLUDE_DIR
+      ${PROJECT_SOURCE_DIR}/third_party/vulkan-headers/include
+      CACHE PATH "vk headers dir" FORCE)
+    set(VulkanRegistry_DIR
+      ${PROJECT_SOURCE_DIR}/third_party/vulkan-headers/registry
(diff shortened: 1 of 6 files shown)


third_party/CMakeLists.txt
```diff
@@ -146,6 +146,63 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
   set(VULKAN_LOADER_BUILD_FIXES
     -Wno-unknown-warning-option
   )
+
+  set(SWIFTSHADER_BUILD_FIXES
```
dneto0: oh my

src/vulkan/find_vulkan.cmake
```diff
@@ -19,17 +19,19 @@ set(Vulkan_FOUND FALSE)
...
-  set(VulkanHeaders_INCLUDE_DIR
-    ${PROJECT_SOURCE_DIR}/third_party/vulkan-headers/include
-    CACHE PATH "vk headers dir" FORCE)
-  set(VulkanRegistry_DIR
-    ${PROJECT_SOURCE_DIR}/third_party/vulkan-headers/registry
-    CACHE PATH "vk_registry_dir" FORCE)
-  include_directories(BEFORE "${VulkanHeaders_INCLUDE_DIR}")
-  set(VULKAN_LIB vulkan)
-  message(STATUS "Amber: using local vulkan")
+if (NOT ${Vulkan_FOUND})
```
dneto0: It was a little mysterious to see this be conditioned this way.  But it follows the pattern in the rest of the file and is innocuous otherwise.  So ok with me.

third_party/CMakeLists.txt
```diff
@@ -318,3 +385,17 @@ if (${AMBER_USE_LOCAL_VULKAN})
...
 
+if (${AMBER_ENABLE_SWIFTSHADER})
+  set(CXX_BACK ${CMAKE_CXX_FLAGS})
+  SET(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} "${SWIFTSHADER_BUILD_FIXES}")
+  STRING(REGEX REPLACE ";" " " CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+
+  set(BUILD_EGL FALSE CACHE BOOL FALSE)
+  set(BUILD_GLESv2 FALSE CACHE BOOL FALSE)
+  set(BUILD_GLES_CM FALSE CACHE BOOL FALSE)
+  set(BUILD_VULKAN TRUE CACHE BOOL TRUE)
```
dneto0: Strange we don't need a docstring ?  Or, the docstrings here are FALSE and TRUE.... :-)

Should these be...
set(BUILD_EGL FALSE CACHE BOOL "Should SwiftShader build EGL?")
...

and the one that turns Vulkan on should also have FORCE:

set(BUILD_VULKAN FALSE CACHE BOOL "Should SwiftShader build Vulkan?" FORCE)

dj2: Doing them as "" would match the rest of the file. We don't really expect these to be set, we just have to set them in to CACHE so cmake doesn't lose them.

third_party/CMakeLists.txt
alan-baker: Also you need some more options for GNU.

Issue #501, opened 2019-05-09
Crash when probing without setting data to a buffer


cmarcelo:
After d26ee22 "Unify buffer sizes. (#465)", using probe in a buffer without previously setting it will cause crash apparently due to format not being set. Seems to me that is a bug since would be fine for a buffer to be used as output and only probed at the end.

- If that's the case then probe should also be able to "set the format" of a buffer.

- If that's not the case: would be nice if this behavior was documented and, since amber can identify those cases, an proper error should be printed.

Snippet of the stack trace:

```
#0  amber::Format::SizeInBytes (this=0x0)
#1  0x0000555556b91ced in amber::vulkan::BufferDescriptor::MoveResourceToBufferOutput (this=0x555568758720) at /usr/include/c++/9/bits/unique_ptr.h:357
#2  0x0000555556b8e732 in amber::vulkan::Pipeline::ReadbackDescriptorsToHostDataQueue (this=this@entry=0x55556571b730) at /usr/include/c++/9/bits/unique_ptr.h:357
#3  0x0000555556b9293a in amber::vulkan::ComputePipeline::Compute (this=0x55556571b730, x=<optimized out>, y=1, z=1)
#4  0x0000555556b892f4 in amber::vulkan::EngineVulkan::DoCompute (this=this@entry=0x555568757ac0, command=0x5555687269e0)
#5  0x0000555556b5188a in amber::Executor::ExecuteCommand (this=this@entry=0x7fffffffb9cf, engine=engine@entry=0x555568757ac0, cmd=0x5555687269e0)
#6  0x0000555556b51ac9 in amber::Executor::Execute (this=this@entry=0x7fffffffb9cf, engine=0x555568757ac0, script=<optimized out>, shader_map=..., executionType=amber::ExecutionType::kExecute)
    at /usr/include/c++/9/bits/unique_ptr.h:357
#7  0x0000555556b42342 in amber::Amber::ExecuteWithShaderData (this=<optimized out>, recipe=<optimized out>, opts=0x7fffffffbc40, shader_data=...) at /usr/include/c++/9/bits/unique_ptr.h:357
```

cmarcelo:
Example snippet that leads to a crash. Uncommenting the subdatas there's no crash.

```
[test]
ssbo 0:0 2048
ssbo 0:1 2176
uniform ubo 0:2 int 0 32

# ssbo 0:0 subdata int 0  0
# ssbo 0:1 subdata int 0  0

compute entrypoint main
compute 32 1 1

[...]

PR #514, opened 2019-05-16
[vkscript] Set default format into buffers.

### LINKED ISSUE #513: Support scratch buffers: no subdata, no probe
Followup to #501 

Currently Amber blindly copies all buffers back and forth between CPU and GPU, and for that it requires a format.  To get a format, the script needs to include either a subdata or probe.
This requirement is confusing and pollutes user scripts.

We should support scratch buffers that have a size but no probe or subdata.

dj2:
When the SSBO size command is used there is currently no format set into
a buffer. This CL sets a default format of `R8_SINT`. A flag is also set
that the format is defaulted and allows it to be overridden.

This permits the usage of scratch buffers, but allows the buffer to be
transferred to/from the device as we can set the buffer size to the
number of provided elements.

Fixes #513
Issue #501

### DIFF
--- a/docs/vk_script.md
+++ b/docs/vk_script.md
@@ -232,7 +232,7 @@ use the same `type` or the script will be rejected as invalid.
  * `ssbo _binding_ _size_`
 
 Sets the number of elements in the SSBO at `binding` to `size`. The buffer will
-be created with a default format of `int`. This default format will be
+be created with a default format of `char`. This default format will be
 overridden by a call to `ssbo subdata` or `probe ssbo`.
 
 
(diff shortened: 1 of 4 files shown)


docs/vk_script.md
```diff
@@ -231,9 +231,9 @@ use the same `type` or the script will be rejected as invalid.
 ### SSBO size
  * `ssbo _binding_ _size_`
 
-Sets the number of elements in the SSBO at `binding` to `size`. Note, Amber
-requires a buffer to have a format. So, a `subdata` or `probe ssbo` call must
-be made with this buffer in order to set the buffer format.
+Sets the number of elements in the SSBO at `binding` to `size`. The buffer will
+be created with a default format of `int`. This default format will be
```
dneto0: It's `char` since it's R8_SINT

dj2: done

Issue #515, opened 2019-05-16
Amber only ever uses the first tolerance value.


dj2:
Currently if you set tolerances into amberscript or vkscript the verifier only uses the first value. The code in verifier.cc needs to be updated to pull tolerance values from each of the possible values.

PR #517, opened 2019-05-16
[amberscript] Add TOLERANCE parsing.

### LINKED ISSUE #418: Parse EXPECT TOLERANCE
The amberscript parser needs to handle the expect tolerance command.

dj2:
This CL adds the TOLERANCE flag in to the EXPECT parsing. The TOLERANCE
requires an EQ buffer comparison (to match how VkScript works).

Fixes #418

### DIFF
--- a/docs/amber_script.md
+++ b/docs/amber_script.md
@@ -387,7 +387,8 @@ EXPECT {buffer_name} IDX _x_ {comparator} _value_+
 # Checks that |buffer_name| at |x| has values within |tolerance| of |value|
 # The |tolerance| can be specified as 1-4 float values separated by spaces.
 # The tolerances may be given as a percentage by placing a '%' symbol after
-# the value.
+# the value. If less tolerance values are provided then are needed for a given
+# data component the default tolerance will be applied.
 EXPECT {buffer_name} IDX _x_ TOLERANCE _tolerance_{1,4} EQ _value_+
 
 # Checks that |buffer_name| at |x|, |y| for |width|x|height| pixels has the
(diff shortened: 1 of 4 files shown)


docs/amber_script.md
```diff
@@ -385,10 +385,10 @@ CLEAR {pipeline}
 EXPECT {buffer_name} IDX _x_ {comparator} _value_+
 
 # Checks that |buffer_name| at |x| has values within |tolerance| of |value|
-# when compared with the given |comparator|. The |tolerance| can be specified
-# as 1-4 float values separated by spaces.
-EXPECT {buffer_name} IDX _x_ TOLERANCE \
-    _tolerance_{1,4} {comparator} _value_+
+# The |tolerance| can be specified as 1-4 float values separated by spaces.
```
dneto0: Missing the last half of the sentence that used to be on line 388

dneto0: Also, how do we apply tolerances if the number of tolerance values differ from the data type.

E.g. I give 2 tolerance scalars, but the data is vec4.   Does it repeat, or fill the remaining tolerance components with 0?  I would expect the latter

dj2: The last half isn't true anymore. I implemented this as VkScript at the moment where EQ is the only comparitor that takes tolerances. So, you don't really specify the comparitor anymore as much as say EQ all the time.

We can do either, it isn't specified for VkScript and, at the moment it's moot as the original implementation of the verify only ever uses the first value (I filed #515 to fix that). I updated the text to say we'd use the default tolerance (instead of 0) which is  0.000001.

PR #521, opened 2019-05-23
Simplify build settings

dj2:
This CL targets the build warnings settings to just the amber source files. This allows us to remove all the overrides needed for building third_party software.


dj2:
@alan-baker

alan-baker:
My version of clang on Mac required turning off warnings as errors for swiftshader. There's a tautological compare somewhere. Otherwise this worked.

PR #525, opened 2019-05-24
Shader specialization

### LINKED ISSUE #510: Support shader specialization in Amberscript
Proposed syntax would to add a new subcommand to the ATTACH command for shaders. An example pipeline would look like:
```
PIPELINE compute kCompute
  ATTACH kComputeShader [...] [SPECIALIZE <spec id> AS <type> <value>]+
  ...
END
```

The specialization mapping would be stored in ShaderInfo. Storage details TBD. The Vulkan engine would setup VkPipelineShaderStageInfo with VkSpecializationInfo that uses the mapping provided in the attach command. Dawn would do something equivalent.

I plan to implement this.

alan-baker:
Fixes #510

* Parser support for SPECIALIZE subcommand of ATTACH
 * tests
* Documentation
* Support Vulkan pipeline creation using shader specialization
 * test for specialization


docs/amber_script.md
```diff
@@ -197,6 +197,11 @@ The following commands are all specified within the `PIPELINE` command.
   # Attach a 'multi' shader to the pipeline of |shader_type| and use the entry
   # point with |name|. The provided shader _must_ be a 'multi' shader.
   ATTACH {name_of_multi_shader} TYPE {shader_type} ENTRY_POINT {name}
+
+  # Attach specialized shader. Specialization can be specified multiple times.
+  # Specialization values must be a 32-bit type.
+  ATTACH SPECIALIZE 1 AS uint32 4
+  ATTACH SPECIALIZE 1 AS uint32 4 SPECIALIZE 4 AS float 1.0
```
dj2: This seems like it's missing the {name of shader} and possibly TYPE lines?

src/amberscript/parser.cc
```diff
@@ -480,15 +480,88 @@ Result Parser::ParsePipelineAttach(Pipeline* pipeline) {
...
+}
+
+Result Parser::ParseShaderSpecialization(Pipeline* pipeline) {
+  auto token = tokenizer_->NextToken();
+  if (!token->IsInteger())
+    return Result("specialization ID must be an integer");
+
+  auto spec_id = token->AsUint32();
+  if (spec_id == 0)
+    return Result("specialization ID must be greater than 0");
```
dj2: Is there a max spec Id bounds we should check as well?

src/amberscript/parser.cc
```diff
@@ -480,15 +480,88 @@ Result Parser::ParsePipelineAttach(Pipeline* pipeline) {
...
+  switch (type.GetType()) {
+    case DataType::kUint32:
+      value = token->AsUint32();
+      break;
+    case DataType::kInt32: {
+      union {
+        uint32_t u;
+        int32_t i;
+      } u;
+      u.i = token->AsInt32();
```
dj2: Why do the union dance and not just token->AsUint32()?

src/amberscript/parser.cc
```diff
@@ -480,15 +480,88 @@ Result Parser::ParsePipelineAttach(Pipeline* pipeline) {
...
+      value = u.u;
+      break;
+    }
+    case DataType::kFloat: {
+      union {
+        uint32_t u;
+        float f;
+      } u;
+      u.f = token->AsFloat();
```
dj2: You probably want to do token->ConvertToDouble() here before getting AsFloat otherwise a 1 will come back as 0 as the double value is zero and the token is an integer.

Also, I dont' think you need the union dance.

src/amberscript/parser_attach_test.cc
```diff
@@ -307,5 +307,213 @@ END)";
...
+TEST_F(AmberScriptParserTest, PipelineSpecializationMultipleSpecializations) {
+  std::string in = R"(
+SHADER compute my_shader GLSL
+#shaders
+END
+PIPELINE compute my_pipeline
+  ATTACH my_shader TYPE compute ENTRY_POINT my_ep \
+      SPECIALIZE 1 AS uint32 4 \
+      SPECIALIZE 2 AS uint32 5 \
+      SPECIALIZE 5 AS uint32 1
```
dj2: I like this, it reads well and keeps the data grouped together nicely.

src/amberscript/parser_attach_test.cc
```diff
@@ -307,5 +307,213 @@ END)";
...
+  ASSERT_EQ(1U, pipelines.size());
+
+  const auto* pipeline = pipelines[0].get();
+  const auto& shaders = pipeline->GetShaders();
+  ASSERT_EQ(1U, shaders.size());
+
+  EXPECT_EQ(1, shaders[0].GetSpecialization().size());
+  EXPECT_EQ(4, shaders[0].GetSpecialization().at(1));
+}
+
```
dj2: Can you add one more test where we setup a pipeline with shader specialization and then we `DERIVE pipeline2 FROM pipeline` and make sure pipeline2 shaders get the specializations?

src/vulkan/engine_vulkan.cc
```diff
@@ -78,12 +78,13 @@ EngineVulkan::~EngineVulkan() {
...
-    for (auto mod_it = info.shaders.begin(); mod_it != info.shaders.end();
-         ++mod_it) {
+    for (auto mod_it = info.shader_info.begin();
+         mod_it != info.shader_info.end(); ++mod_it) {
       auto vk_device = device_->GetVkDevice();
-      if (vk_device != VK_NULL_HANDLE && mod_it->second != VK_NULL_HANDLE)
-        device_->GetPtrs()->vkDestroyShaderModule(vk_device, mod_it->second,
-                                                  nullptr);
+      if (vk_device != VK_NULL_HANDLE &&
+          mod_it->second.shader != VK_NULL_HANDLE)
```
dj2: Add braces now that this is multi-line

src/vulkan/engine_vulkan.cc
```diff
@@ -257,7 +258,35 @@ Result EngineVulkan::SetShader(amber::Pipeline* pipeline,
     return Result("Vulkan::Calling vkCreateShaderModule Fail");
   }
 
-  info.shaders[type] = shader;
+  info.shader_info[type].shader = shader;
+
+  for (auto& shader_info : pipeline->GetShaders()) {
+    if (shader_info.GetShaderType() == type) {
```
dj2:
```
if (shader_info.GetShaderType() != type)
  continue;
```

src/vulkan/engine_vulkan.cc
```diff
@@ -257,7 +258,35 @@ Result EngineVulkan::SetShader(amber::Pipeline* pipeline,
...
+      const auto& shader_spec_info = shader_info.GetSpecialization();
+      if (!shader_spec_info.empty()) {
```
dj2:
```
if (shader_spec_info.empty())
   continue;
```

tests/cases/shader_specialization.amber
```diff
@@ -0,0 +1,109 @@
+#!amber
+
```
dj2: Add a copyright

src/amberscript/parser.cc
alan-baker: I don't believe so.

src/amberscript/parser.cc
alan-baker: Ok, it wasn't clear those were always valid, I'll change the code.

src/amberscript/parser_attach_test.cc
alan-baker: Done

tests/cases/shader_specialization.amber
alan-baker: Done

src/vulkan/engine_vulkan.cc
alan-baker: Done

src/vulkan/engine_vulkan.cc
alan-baker: Done

src/vulkan/engine_vulkan.cc
alan-baker: Done

src/amberscript/parser.cc
alan-baker: In fact zero is valid, so I'll remove that error.

PR #526, opened 2019-05-27
Add AMBER_ENABLE_SHARED_CRT option

dneto0:
On Windows, using this will cause Amber to build with /MD or MDd.
Normally Amber builds with /MT or MTd.


dneto0:
With AMBER_ENABLE_SHARED_CRT, this fails to link the gtest exectuables.

PR #527, opened 2019-05-27
/MD and /MT are truly project-wide settings

dneto0:


dneto0:
I've tested this with a Windows Release build, with and without AMBER_ENABLE_SHARED_CRT

PR #529, opened 2019-05-27
Avoid Vulkan-loader error for Wayland include dir

dneto0:


CMakeLists.txt
```diff
@@ -44,7 +44,6 @@ option(AMBER_SKIP_SAMPLES
 option(AMBER_USE_LOCAL_VULKAN "Build with vulkan in third_party" OFF)
 option(AMBER_ENABLE_SWIFTSHADER
   "Build using SwiftShader" ${AMBER_ENABLE_SWIFTSHADER})
-option(AMBER_USE_LOCAL_VULKAN "Build with vulkan in third_party" OFF)
```
dneto0: This is redundant with line 44

dneto0:
This avoids an error at CMake config time when building Swiftshader, for me anyway.
Otherwise I error out at https://github.com/KhronosGroup/Vulkan-Loader/blob/master/CMakeLists.txt#L108

dneto0:
With this, I can see a Swiftshsader device with "amber -d V" and the correct VK_ICD_FILENAMES setting.

third_party/CMakeLists.txt
```diff
@@ -51,6 +51,16 @@ if (${AMBER_USE_LOCAL_VULKAN})
   add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/vulkan-headers)
 
   set(BUILD_TESTS FALSE)
+
+  # The vulkan-loader CMake file assumes that directory exists if
+  # Wayland support is to be built.
+  if(NOT EXISTS ${WAYLAND_CLIENT_INCLUDE_DIR})
+    message(STATUS "Amber: Disabling Wayland support in Vulkan-Loader")
+    set(BUILD_WSI_WAYLAND_SUPPORT OFF CACHE BOOL "" FORCE)
```
dj2: Should we just turn it off unconditionally like we do x11?

CMakeLists.txt
dj2: Yea, I think it was a bad merge at one point. I've also removed it in my DXC CL, so we're going to conflict.

third_party/CMakeLists.txt
dneto0: I was trying to be minimally invasive.  
I unconditionally turned off X11 support because of a missing include header that I didn't want to track down.

Since this is a local build of SwiftShader (at least for now), and we want it to be headless, I could see turning everything off

dneto0: I was trying to be minimally invasive.  
I unconditionally turned off X11 support because of a missing include header that I didn't want to track down.

Since this is a local build of SwiftShader (at least for now), and we want it to be headless, I could see turning everything off

dj2:
We can do it later if we want. Rolling this in for now so we can move forward.

PR #531, opened 2019-05-29
[WIP] Adding CLSPV support

### LINKED ISSUE #429: Compile and build CLSPV in Amber
Add CLSPV to the third_party in amber and hookup the build. This should be an optional component that is off by default.

dj2:
This CL adds the necessary code to build CLSPV as part of the Amber
build.

Fixes #429


dj2:
This is marked as WIP because it requires a patch to LLVM in order to build. I see no way of getting this to work otherwise. It seems that even with testing disabled in LLVM the LLVMTestingSupport library is being created. But, if testing is disabled then gtest isn't available and configure fails.

```
diff --git a/lib/CMakeLists.txt b/lib/CMakeLists.txt
index c458927e3cd..34f78cb4398 100644
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -27,5 +27,7 @@ add_subdirectory(Passes)
 add_subdirectory(TextAPI)
 add_subdirectory(ToolDrivers)
 add_subdirectory(XRay)
-add_subdirectory(Testing)
+if (${LLVM_INCLUDE_TESTING})
+  add_subdirectory(Testing)
+endif()
 add_subdirectory(WindowsManifest)
```

dj2:
@alan-baker  fyi

dj2:
Oh, this also needs a CLSPV patch, which I'll push up.

dj2:
https://github.com/google/clspv/pull/374

alan-baker:
FYI https://reviews.llvm.org/D63097 is for the LLVM patch.

alan-baker:
Subsumed by #574

PR #545, opened 2019-06-10
Attach buffers and textures at create pipeline state

### LINKED ISSUE #16: Dawn: Implement DoDrawRect

### LINKED ISSUE #23: Dawn: Add vertex, index, ubo, and ssbo buffers when creating the pipeline
Implement SetBuffer in the dawn engine.

### LINKED ISSUE #469: Dawn: CreateFramebufferIfNeeded should be done once during CreatePipeline
Right now we call CreateFramebufferIfNeeded.  It will create texture and buffer objects *each* time. That leaks objects like mad, which is bad.

Really this should be done exactly once, when we create the pipeline, and we can rename the method to CreateColorAndDepthStencilAttachments. 

This resolves a TODO(dneto) in CreateFramebufferIfNeeded.

### LINKED ISSUE #491: Dawn doesn't support push constants
Dawn doesn't have support for push constants and there are no plans on adding them. We need to detect the use of push constants in a script and fail the execution during `CreatePipeline`.

Currently the pipeline does not know there are push constants so we should add a 'HasPushConstants' flag to the pipeline which gets set in the appropriate places of the vkscript and amberscript parsers.

sarahM0:
fixes #469 
fixes #491 
fixes #16 
fixes #23


src/dawn/engine_dawn.h
```diff
@@ -73,12 +73,18 @@ class EngineDawn : public Engine {
...
-
-  // TODO(dneto): Remove this. Shaders are attached to the pipeline.
-  Result SetShader(ShaderType type, const std::vector<uint32_t>& data);
+  Result AttachBuffersAndTextures(RenderPipelineInfo* render_pipeline);
 
   ::dawn::Device* device_ = nullptr;  // Borrowed from the engine config.
+  ::dawn::Buffer fb_buffer_;
+  bool fb_is_created_ = false;
+
+  ::dawn::TextureView texture_view_;
```
dneto0: I assume texture_view_ is the view into fb_texture_ ?  If so then I'd put the fb_texture_ member first, then texture_view_ second and add a comment saying that it's a view into fb_texture_

tests/cases/draw_rect_multiple_color_attachemnt.amber
```diff
@@ -0,0 +1,40 @@
...
+
+PIPELINE graphics my_pipeline
+  ATTACH vert_shader
+  ATTACH frag_shader
+  FRAMEBUFFER_SIZE 800 600
+  BIND BUFFER framebuffer AS color LOCATION 0
+  BIND BUFFER framebuffer1 AS color LOCATION 1
+END
+
+RUN my_pipeline DRAW_RECT POS 0 0 SIZE 250 250
```
dneto0: I would increase this to SIZE 800 600 since you set the framebuffer size to that.

src/dawn/pipeline_info.h
```diff
@@ -50,19 +50,13 @@ struct RenderPipelineInfo {
...
-  uint32_t fb_num_rows = 0;
-  /// The number of data bytes in the framebuffer host-side buffer.
-  uint32_t fb_size = 0;
 
-  // TODO(dneto): Record index data
-  // TODO(dneto): Record buffer data
+  ::dawn::Buffer vertex_buffer;
+  ::dawn::Buffer index_buffer;
+
+  bool hasBinding = false;
```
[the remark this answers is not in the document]
dneto0: I think this bool state variable can be replaced by checking bind_group against false?

src/dawn/engine_dawn.cc
```diff
@@ -659,7 +688,11 @@ Result DawnPipelineHelper::CreateRenderPipelineDescriptor(
     depth_stencil_format = ::dawn::TextureFormat::D32FloatS8Uint;
   }
 
-  renderPipelineDescriptor.layout = MakeBasicPipelineLayout(device, nullptr);
+  if (render_pipeline.hasBinding)
```
dneto0: I think you can replace this "if" by passing down a reference to render_pipeline.bind_group_layout down to the  helper, and then the helper can do different things based on whether the layout compares equal to false or not.

src/dawn/engine_dawn.cc
```diff
@@ -874,8 +906,12 @@ Result EngineDawn::DoDrawRect(const DrawRectCommand* command) {
   ::dawn::RenderPassEncoder pass =
       encoder.BeginRenderPass(renderPassDescriptor);
   pass.SetPipeline(pipeline);
-  pass.SetVertexBuffers(0, 1, &vertexBuffer, vertexBufferOffsets);
-  pass.SetIndexBuffer(indexBuffer, 0);
+  if (render_pipeline->hasBinding) {
```
dneto0: Recommend  
   
     if (render_pipeline->bind_group) { ... }

src/dawn/engine_dawn.cc
```diff
@@ -317,7 +281,7 @@ MapResult MapTextureToHostBuffer(const RenderPipelineInfo& render_pipeline,
     }
   }
   // Always unmap the buffer at the end of the engine's command.
-  fb_buffer.Unmap();
+  render_pipeline.fb_buffer.Unmap();
```
dneto0: Also need to copy the buffer data for buffers that are not colour-attachments back into the Amber buffer objects.
Need a TODO here at least.

src/dawn/engine_dawn.cc
```diff
@@ -111,8 +111,8 @@ Result MakeTexture(const ::dawn::Device& device,
   return Result("Dawn: Failed to allocate a framebuffer texture");
 }
 
-// Creates a host-side buffer for the framebuffer, and returns it through
-// |result_ptr|. The buffer will be used as a transfer destination and
+// Creates a host-side buffer  of |size| for the framebuffer, and returns it
```
dneto0: .. of |size| _bytes_ ...

src/dawn/engine_dawn.cc
dneto0: ok, we can leave this for a future cleanup,.

src/dawn/engine_dawn.cc
```diff
@@ -109,54 +111,24 @@ Result MakeTexture(const ::dawn::Device& device,
...
```
dneto0: Still needs the units:

// ... buffer of |size| _bytes_ for the framebuffer...

PR #552, opened 2019-06-17
[WIP] Update code to catch up with Dawn updates

sarahM0:
WIP: please do not review
fixes: #550

### DIFF
--- a/src/dawn/engine_dawn.cc
+++ b/src/dawn/engine_dawn.cc
@@ -45,6 +45,53 @@ static const uint32_t kMaxColorAttachments = 4u;
 static const uint32_t kMaxVertexInputs = 16u;
 static const uint32_t kMaxVertexAttributes = 16u;
 
+// A DS for creating and setting the defaults for VertexInputDescriptor
+// Copied from Dawn utils source code.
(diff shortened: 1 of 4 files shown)


src/dawn/engine_dawn.cc
```diff
@@ -712,29 +757,29 @@ Result DawnPipelineHelper::CreateRenderPipelineDescriptor(
...
 
   // Set defaults for the input state descriptors.
-  tempInputState.indexFormat = ::dawn::IndexFormat::Uint32;
-  // Fill the default values for vertexInput.
-  vertexInput.inputSlot = 0;
-  vertexInput.stride = 0;
-  vertexInput.stepMode = ::dawn::InputStepMode::Vertex;
-  for (uint32_t i = 0; i < kMaxVertexInputs; ++i) {
-    tempInputs[i] = vertexInput;
+  vertexInputDescriptor = new ComboVertexInputDescriptor();
```
dneto0: It seems unusual to have an allocation of this object, and then hand ownership over to the Dawn API.  I guess this leaks.  I would expect this to be allocated on the stack and you pass in the address to the .vertexInput member of renderPipelineDiscriptor ??

Issue #553, opened 2019-06-18
Validator might falsely fail requiring fragment shader


archimedus:
Enforcement of validation introduced in https://github.com/google/amber/pull/502 breaks certain tests with message

> graphics pipeline requires a vertex shader

This validation is not always correct, due to **Vulkan permits graphics pipelines without fragment shader** (please see 1.1.106, Valid Usage section of `VkGraphicsPipelineCreateInfo`):

> The stage member of one element of pStages must be either VK_SHADER_STAGE_VERTEX_BIT or VK_SHADER_STAGE_MESH_BIT_NV.

Method `vkscript::Parser::SkipValidationForTest()` cannot be used to alleviate the issue due to it is inaccessible from outside i.e. from class `amber::Amber`.

Please consider:
 * remove the validation check
 * make some alternative to `SkipValidationForTest()` available outside to give control over validation (for example extend `Amber::Parse()` with `bool validation=true` parameter)

As more long term solution it might be useful to create some amber file directives to control compilation/validation. If you find this idea useful please create separate issue, as I want this to be fixed quickly.

Thanks.

dj2:
Is the fix for this just to not require a fragment shader in the validation? Does the card provide a default shader? If I have a pass through vertex shader and no fragment shader what happens?

archimedus:
Output into SSBO is done. Example is here:

https://github.com/KhronosGroup/VK-GL-CTS/blob/master/external/vulkancts/data/vulkan/amber/pipeline/vertex_only/position_to_ssbo.amber

dj2:
Ah, cool. So, it sounds like we just have to validate that the vertex shader is set and remove the check for the fragment shader?

Thanks a lot for pointing out the example. If you make the PR to fix the issue can you add the test to the tests/cases folder so we don't break it in the future?

archimedus:
> Ah, cool. So, it sounds like we just have to validate that the vertex shader is set and remove the check for the fragment shader?

Fragment shader is not obligatory for graphics pipeline.

> Thanks a lot for pointing out the example. If you make the PR to fix the issue can you add the test to the tests/cases folder so we don't break it in the future?

 I supposed it would be better to exclude one test from the validation, but as far as I understand it's not an option. I abandoned my fix and pending for your fix. I suppose you will provide it faster. Thanks.

PR #554, opened 2019-06-18
Disable validation after parse

archimedus:
The CL makes it possible to disable validation for certain tests.
Required to resolve https://github.com/google/amber/issues/553


dj2:
I don't think this is the right way to fix this issue. Just skipping validation makes the script potentially invalid. We aren't able to check buffers or anything else we need to verify is present.

It sounds like the correct solution is to fix the validation to work properly. If we're failing valid scripts that's the part that needs to be corrected.

src/amber.cc
```diff
@@ -96,6 +98,9 @@ amber::Result Amber::Parse(const std::string& input, amber::Recipe* recipe) {
   else
     parser = MakeUnique<vkscript::Parser>();
 
+  if (!validate)
+    parser->SkipValidationForTest();
```
dj2: This is only to be used in testing and shouldn't appear anywhere else.

Issue #21, opened 2018-11-14
Dawn: Implement DoBuffer


dj2:
Implement DoBuffer in the Dawn Engine.

sarahM0:
Do we know the max size of an ssbo while creating the pipelines? (before executing the ssbo command)

PR #562, opened 2019-06-24
Set the element count so we'll know the max size of a ssbo

sarahM0:
This change sets the max element count of an ssbo. The is needed since Dawn backend needs to know the buffer size while creating it and passing to the pipeline layout. 
issue: #21

### DIFF
--- a/src/vkscript/command_parser.cc
+++ b/src/vkscript/command_parser.cc
@@ -615,7 +615,21 @@ Result CommandParser::ProcessSSBO() {
     if (!r.IsSuccess())
       return r;
 
-    buf->SetDataWithOffset(values, cmd->GetOffset());
+    // Multiply by the input needed because the value count will use the needed
+    // input as the multiplier
+    uint32_t value_count =
+        ((cmd->GetOffset() / buf->GetFormat()->SizeInBytes()) *
+         buf->GetFormat()->InputNeededPerElement()) +
+        static_cast<uint32_t>(values.size());
+    // The buffer should only be resized to become bigger. This means that if a
+    // command was run to set the buffer size we'll honour that size until a
+    // request happens to make the buffer bigger.
+    if (value_count > buf->ValueCount())
+      buf->SetValueCount(value_count);
+
+    // Even if the value count doesn't change, the buffer is still resized
+    // because this maybe the first time data is set into the buffer.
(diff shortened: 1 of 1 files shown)


src/vkscript/command_parser.cc
```diff
@@ -615,6 +615,10 @@ Result CommandParser::ProcessSSBO() {
     if (!r.IsSuccess())
       return r;
 
+    // Resize the buffer so we'll know the max size
+    if (buf->ElementCount() < values.size())
+      buf->SetElementCount(static_cast<uint32_t>(values.size()));
```
dj2: I don't think this is quite correct. You need to take the offset into account as well as the values size. If we set 2 values at offset 50, we need to make sure the buffer is 50+(values.size()) bytes, which also doesn't necessarily match up to element counts ....

sarahM0: Yes that's right. How about calling SetDataWithOffset?

src/vkscript/command_parser.cc
```diff
@@ -615,9 +615,7 @@ Result CommandParser::ProcessSSBO() {
     if (!r.IsSuccess())
       return r;
 
-    // Resize the buffer so we'll know the max size
-    if (buf->ElementCount() < values.size())
-      buf->SetElementCount(static_cast<uint32_t>(values.size()));
+    buf->SetDataWithOffset(values, cmd->GetOffset());
```
dj2: I don't think this is going to do what we want either. The trick with VkScript ssbo data is that it's set late. So, this is going to set the data during parse, but we don't want it set until we see the command. So, if you do:

```
ssbo data 1 2 3
draw
ssbo data 1 4 5
draw
```

with this change both draws would see `4 5` in the data, when only the second should see it.

Do you have an example issue where this arises? Is the problem that you need the full buffer size at all times, even outside the resize commands?

sarahM0: While creating the the pipeline layout and bindings, which are created before executing commands, pointers to already created buffers are needed (not like Vulkan that needs just the handles). And to create a Dawn buffer, size has to be known. That means I need to know the max size of a buffer while creating the pipeline.

I thought at each ssbo command Vulkan backend resets the data. How about copying the first few lines of that function to just set the size and not the data?

dj2: I think that makes sense.

Issue #558, opened 2019-06-19
Improve portability of samples in tests/cases: use supported vertex formats


dneto0:
E.g. draw_triangle_list.amber uses 3-channel 8bits-per-channel UNORM :

```
BUFFER vert_color DATA_TYPE R8G8B8_UNORM DATA
255   0   0
255   0   0
255   0   0
...
```

But, for example, Vulkan does not require R8G8B8_UNORM to be supported as a vertex format.
See the checkmarks for VK_FORMAT_FEATURE_VERTEX_BUFFER_BIT in the table "Mandatory format support: 1-3 byte-sized channels".  VK_FORMAT_R8G8B8_UNORM has no checkmark.

So change that sample to a 4-channel format with same channel width and also UNORM.
The next table "Mandatory format support: 4 byte-sized channels" shows that VK_FORMAT_R8G8B8A8_UNORM is required to be supported.

Review the other samples and update them to use only required vertext formats.  This will improve portability of the test case.s

dj2:
@sarahM0 how much of this is left? I know a bunch got changed with the doBuffer support?

sarahM0:
I haven't check compute_.* tests, yet. And maybe a couple more are overlooked.

sarahM0:
I just finished checking compute samples. Looks like they're mostly fine. I still need to check comput_push_const_.*.

These are not, but I think that's the point so I'm gonna leave them be:
  "draw_triangle_list_in_r16g16b16a16_snorm_color_frame.vkscript",
  "draw_triangle_list_in_r16g16b16a16_uint_color_frame.vkscript",
  "draw_triangle_list_in_r32g32b32a32_sfloat_color_frame.vkscript",
  "draw_triangle_list_in_r8g8b8a8_snorm_color_frame.vkscript",
  "draw_triangle_list_in_r8g8b8a8_srgb_color_frame.vkscript",

dj2:
Awesome, thank you for reviewing all of them.

dj2:
Were you able to review the `compute_push_const_*` tests to determine if they needed changes? Sounds like this bug can be closed out and we can make specific bugs in the future if we find failing tests?

PR #564, opened 2019-06-25
[Dawn] Implement doBuffer in Dawn backend

sarahM0:
Fixs #21, except issue #573 
Addresses issue #558 
Organizes suppressed tests list for Dawn backend (run_tests.py)
Removes the unnecessary resize in process ssbo command


src/dawn/engine_dawn.cc
```diff
@@ -932,8 +933,29 @@ Result EngineDawn::DoPatchParameterVertices(
   return Result("Dawn:DoPatch not implemented");
 }
 
-Result EngineDawn::DoBuffer(const BufferCommand*) {
-  return Result("Dawn:DoBuffer not implemented");
+Result EngineDawn::DoBuffer(const BufferCommand* command) {
+  Result result;
+
+  // TODO(SarahM0): Make this work for compute pipeline
```
jaebaek: If this does not support compute pipeline yet, just return error message for compute pipeline here?

```
if (command->GetPipelineType == Compute) // fix this correctly
return Result("Compute pipeline is not supported by Dawn backend");
```

src/dawn/engine_dawn.cc
```diff
@@ -932,8 +933,29 @@ Result EngineDawn::DoPatchParameterVertices(
...
+  RenderPipelineInfo* render_pipeline = GetRenderPipeline(command);
+  if (!render_pipeline)
+    return Result("DoBuffer invoked on invalid or missing render pipeline");
+  if (!command->IsSSBO() && !command->IsUniform())
+    return Result("DoBuffer not supported buffer type");
```
jaebaek: "DoBuffer only supports SSBO and uniform buffers"

src/dawn/engine_dawn.cc
```diff
@@ -932,8 +933,29 @@ Result EngineDawn::DoPatchParameterVertices(
...
+  ::dawn::Buffer* dawn_buffer = buffer_map_[command->GetBinding()];
+  if (dawn_buffer) {
+    Buffer* amber_buffer = command->GetBuffer();
+    if (amber_buffer) {
+      amber_buffer->SetDataWithOffset(command->GetValues(),
+                                      command->GetOffset());
+
+      dawn_buffer->SetSubData(command->GetOffset(),
+                              amber_buffer->GetSizeInBytes(),
+                              amber_buffer->ValuePtr()->data());
```
jaebaek: I am not versed on `dawn::Buffer::SetSubData(offset, size_in_byte, data_ptr)`, but does it mean

"Copy data_ptr[0 : size_in_byte] to dawn::Buffer::buffer[offset : offset + size_in_byte]"?

src/dawn/engine_dawn.h
```diff
@@ -89,6 +89,9 @@ class EngineDawn : public Engine {
   // Mapping from the generic engine's Pipeline object to our own Dawn-specific
   // pipelines.
   std::unordered_map<amber::Pipeline*, ::amber::dawn::Pipeline> pipeline_map_;
+
+  std::unordered_map<uint32_t, ::dawn::Buffer*> buffer_map_;
```
jaebaek: In Dawn, is there no concept of descriptor set number?
I only saw "binding" number in the above lines of code.

src/dawn/engine_dawn.cc
sarahM0: Yes, that's right.

src/dawn/engine_dawn.h
sarahM0: Yes you're right! It's fixed in the new commits.

src/vkscript/command_parser.cc
```diff
@@ -633,10 +633,6 @@ Result CommandParser::ProcessSSBO() {
     if (value_count > buf->ValueCount())
       buf->SetValueCount(value_count);
 
-    // Even if the value count doesn't change, the buffer is still resized
-    // because this maybe the first time data is set into the buffer.
-    buf->ResizeTo(buf->GetSizeInBytes());
```
dj2: This is required, please don't remove this.

sarahM0:
The problem is for the buffers with pack of zero each time we process ssbo,
it resizes the buffer. As a result even if the valueCount is the same the
size gets multiplied by ElementCount().
However the buffer is resized correctly whenever we call
setDatsEithOffset().

dj2:
> The problem is for the buffers with pack of zero each time we process ssbo, it resizes the buffer. As a result even if the valueCount is the same the size gets multiplied by ElementCount(). However the buffer is resized correctly whenever we call setDatsEithOffset().

I'm not sure what you mean by a pack of zero? The reason we have to do this is because we may have an ssbo size, but never set any data into the buffer. In the case of an `ssbo size` we have to use the number as the byte size as we don't know the size of the element, so we assume each element is 1 byte.

So, you do:

```
ssbo size foo 100
draw ...
probe foo 100 2 2
```

The draw shader, uses the ssbo as output, so it has to have the size set, but we never set any data. So, we don't really know what's in the buffer, and we have to use the 100 as the byte size.

src/vkscript/command_parser.cc
sarahM0: The problem is for the buffers with pack of zero each time we process ssbo,
it resizes the buffer. As a result even if the valueCount is the same the
size gets multiplied by ElementCount().
However the buffer is resized correctly whenever we call
setDataWithOffset().

sarahM0: Dan: 
'm not sure what you mean by a pack of zero? The reason we have to do this is because we may have an ssbo size, but never set any data into the buffer. In the case of an ssbo size we have to use the number as the byte size as we don't know the size of the element, so we assume each element is 1 byte.

So, you do:

ssbo size foo 100
draw ...
probe foo 100 2 2
The draw shader, uses the ssbo as output, so it has to have the size set, but we never set any data. So, we don't really know what's in the buffer, and we have to use the 100 as the byte size.

sarahM0: The outcome of not removing this line is the buffer size getting bigger at each processSSBO() call. For example when I run something like after the first call buffer size is 256 and the next time is 4096.

ssbo 0 subdata vec4 0 1.0 0.0 0.0 1.0
ssbo 0 subdata vec4 0 0.0 1.0 0.0 1.0

I'm looking in the code to why is this happening (yesterday I found it: an if condition base on format's pack size but I cannot find it right now, I don't know why :( ).
-----------------------
I don't think this affects "ssbo size" since it is only executed when the token is "ssbo subdata". Also I added this line and a couple lines before it to set the max ssbo size, it wasn't in the code previously. 
----------------------
I keep looking to see why each call to resize is making the buffer bigger although the elementCount is not changed.

dj2: In the above example, it may call the resize, but it should do nothing as the size didn't change with the `subdata` call? If it's resizing something is wrong.

if you had:

```
ssbo size 0:0 400
ssbo 0:0 subdata vec4 0 0 1 0 0
```

the ssbo should still be `400` afterwards as it's the larger size, resize should only go up.

src/dawn/engine_dawn.cc
```diff
@@ -1271,35 +1305,62 @@ Result EngineDawn::AttachBuffersAndTextures(
...
-    render_pipeline->bind_group_layout =
-        MakeBindGroupLayout(*device_, bindings);
-    render_pipeline->bind_group = MakeBindGroup(
-        *device_, render_pipeline->bind_group_layout, bindingInitalizerHelper);
+  // TODO(sarahM0): fix issue: Add support for doBuffer with sparse descriptor
+  // sets #573
+  if (render_pipeline->used_descriptor_set.size() != 0 &&
+      render_pipeline->used_descriptor_set.size() != max_descriptor_set + 1) {
+    return Result(
+        "AttachBuffersAndTextures: Sparse descriptor_set is not supported");
```
dneto0: good!

src/vkscript/command_parser.cc
dj2: Ohh, @sarahM0 from your example, is the complete example something like:

```
ssbo size 256
ssbo subdata vec4 0 1 2 3 4
```

? If so, I'm guessing the bug is that the `ssbo size` set the element count to `256`. We then changed from a format of `byte` to `vec4` which has a much larger memory size. We originally had `256 * byte` size and changed it to `256 * 4 * 4`?

If so, I think you'll need to track in the buffer if we had just a `size` command, and don't take that as the element count if we're re-calculating the size after a format change. Or .... something along those lines.

src/dawn/pipeline_info.h
```diff
@@ -53,9 +64,16 @@ struct RenderPipelineInfo {
   ::dawn::Buffer fb_buffer;
   std::vector<::dawn::Buffer> vertex_buffers;
   ::dawn::Buffer index_buffer;
+  /// storage and uniform buffers
+  std::vector<::dawn::Buffer> buffers_;
```
dneto0: Nit:  *class* member names end with an underscore, but *struct* member names do not.

src/vkscript/command_parser.cc
sarahM0: Let's say we have a simple script like this:
---------------------------------------------------------------------------------
ssbo 0 subdata vec4 0 1.0 0.0 0.0 1.0
draw rect -1 -1 1 1
relative probe rect rgba (0.0, 0.0, 0.5, 0.5) (1.0,   0,   0, 1.0)

ssbo 0 subdata vec4 0 0.0 1.0 0.0 1.0
draw arrays TRIANGLE_LIST 0 6
relative probe rect rgb (0.0, 0.0, 1.0, 1.0) (0, 1.0, 
---------------------------------------------------------------------------------
It has to do with the ValueCount() function in buffer.h:
  /// Returns the number of values in the buffer.
  uint32_t ValueCount() const {
    if (!format_)
      return 0;
    // Packed formats are single values.
    if (format_->GetPackSize() > 0)
      return element_count_;
    return element_count_ * format_->ValuesPerElement();
  }

When pack_size_in_bytes_ of the buffer format object is 0 the return value is multiplied by format_->ValuesPerElement().
pack_size_in_bytes_ == 0 for most formats, those the do not have PACK in them, such as UNORM and UFLOAT.

Let me know what you think. I'll revert my changes in commad_parser.cc and open another PR for this.

dj2: RIght, but in that script, in both cases element count should be 1 and valuesPerElement should be 4. So, the size should be correct?

This is only a problem with using ssbo size before subdata, yea?

dj2: (and, yes, please file a bug for this so we can figure out what's going wrong and the combination of things that cause it)

sarahM0: You're right the first time:  element count == 1 and valuesPerElement == 4
Then we do:
    if (value_count > buf->ValueCount())
      buf->SetValueCount(value_count);
as a result: element count == 4. Then we do:
    buf->ResizeTo(buf->GetSizeInBytes());
Hence the size increases at each call.

sarahM0: Issue #575

PR #569, opened 2019-06-27
Add a dawn backend flag to the script

### LINKED ISSUE #567: Add a file or script, listing all the tests that pass in Dawn backend

sarahM0:
fixes #567


tests/run_tests.py
```diff
@@ -88,6 +92,8 @@ def RunTest(self, tc):
     cmd = [self.options.test_prog_path, '-q']
     if tc.IsParseOnly():
       cmd += ['-p']
+    if tc.IsUseDawn():
+      cmd += ['-e'] +  ['dawn']
```
dneto0: This is fine.  But it could have been   

     cmd += ['-e', 'dawn']

tests/run_tests.py
```diff
@@ -52,12 +52,89 @@
     # DXC not currently building on bot
     "draw_triangle_list_hlsl.amber"
    ]
-}
+ }
+
+SUPPRESSIONS_DAWN = [
+  "compute_accumulated_ubo_definition.amber",
```
dj2: Can you add a comment as to why things are suppressed? Grouping them is fine and a single #not implemented is also fine. Just, if there are long lived suppressions knowing why is handy.

tests/run_tests.py
```diff
@@ -126,6 +213,10 @@ def SummarizeResults(self):
       for failure in self.failures:
         print failure
 
+      print '\nSummary of Successes:'
```
dj2: I'm not sure if we want to add this, printing the successes is going to hide the failures.

sarahM0: I'll remove it. I added it before I add the suppression list to see what is passing and now there is no need for it.

PR #570, opened 2019-06-27
Error out when DrawRect command is invoked on a pipeline with multiple vertex buffers attached

### LINKED ISSUE #568: Error out when DrawRect command is invoked on a pipeline with multiple vertex buffers attached
The reason is we don't know where to bind the new vertex/index buffer which are going to be created according to the drawRect command. We could say e.g. location 1 is always reserved for drawRect command. However that imposes limitation on the shader code and it's generally too much since we can alway replace a drawRect command with a drawArray.
- We can issue the error in the executor (no need for change in the backends even if we add a new one)

sarahM0:
fix #568


dneto0:
Even if the pipeline only has one vertex buffer, you can still fail with draw_rect because the vertex format has to match as well.  But that can be for another PR.

src/amberscript/parser.cc
```diff
@@ -1005,6 +1005,11 @@ Result Parser::ParseRun() {
     if (!pipeline->IsGraphics())
       return Result("RUN command requires graphics pipeline");
 
+    if (pipeline->GetVertexBuffers().size() > 1)
```
dj2: nit {}'s

src/amberscript/parser.cc
```diff
@@ -1005,6 +1005,11 @@ Result Parser::ParseRun() {
...
+      return Result(
+          "RUN DRAW_RECT is not supported in a pipeline with more than one "
+          "vertex buffers attached");
```
dj2: nit: 'vertex buffer attached'

src/vkscript/command_parser.cc
```diff
@@ -254,6 +254,11 @@ Result CommandParser::ProcessDrawRect() {
   auto cmd = MakeUnique<DrawRectCommand>(pipeline_, pipeline_data_);
   cmd->SetLine(tokenizer_->GetCurrentLine());
 
+  if (pipeline_->GetVertexBuffers().size() > 1)
```
dj2: ditto

src/vkscript/command_parser.cc
```diff
@@ -254,6 +254,11 @@ Result CommandParser::ProcessDrawRect() {
...
+    return Result(
+        "draw rect command is not supported in a pipeline with more than one "
+        "vertex buffers attached");
```
dj2: ditto

src/amberscript/parser.cc
```diff
@@ -1005,6 +1005,11 @@ Result Parser::ParseRun() {
     if (!pipeline->IsGraphics())
       return Result("RUN command requires graphics pipeline");
 
+    if (pipeline->GetVertexBuffers().size() > 1)
+      return Result(
```
dj2: Instead of doing this twice, can the code move into the Validate method in Pipeline? It's called at the end of both VkScript and AmberScript parsing.

sarahM0: That's a good idea, however the way I see it, I need to pass in a pointer to commands to the validate function as an argument since pipeline object does not have access to script, right?

dj2: Ah, crap. Good point, you don't know a drawRect has happened inside pipeline validation. The way you have it currently is good.

src/amberscript/parser.cc
sarahM0: Um.. what do you mean?

dj2: The if() has a multi-line body, add {}'s around the body. Sorry, I realize I never explained myself, heh.

PR #571, opened 2019-07-04
Use simple include paths

jaebaek:
This change allows DXC to be picked up from any location, which may
not be in third_party/dxc/include/. A different build system or
environment may place it in a different directory. If amber is part
of another project that already has a copy of DXC, we would not
want two of them.

### DIFF
--- a/src/dxc_helper.cc
+++ b/src/dxc_helper.cc
@@ -27,12 +27,12 @@
 
 // clang-format off
 // The order here matters, so don't reformat.
-#include "third_party/dxc/include/dxc/Support/WinAdapter.h"
-#include "third_party/dxc/include/dxc/Support/WinIncludes.h"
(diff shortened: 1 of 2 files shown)


dj2:
Is this change needed for something specific? The third_party prefix makes it easy to see where they come from.

s-perron:
@dj2 I think this change allows DXC to be picked up from any location, which may not be in `third_party/dxc/include/`.  A different build system or environment may place it in a different directory.  If amber is part of another project that already has a copy of DXC, we would not want two of them.

dj2:
That makes sense, can the description please be updated to that effect? It's hard to tell from just the title as to _why_ this is being done.

jaebaek:
Updated based what @s-perron mentioned.

PR #577, opened 2019-07-15
Set entry point names during pipeline creation

alan-baker:
Previously pipeline names were only set if there was an entry point command executed. This doesn't match up with how AmberScript specifies entry point names.

Now, if the entry point name was specified in the attach command, it will get set at pipeline creation.

### DIFF
--- a/src/vulkan/engine_vulkan.cc
+++ b/src/vulkan/engine_vulkan.cc
@@ -198,7 +198,7 @@ Result EngineVulkan::CreatePipeline(amber::Pipeline* pipeline) {
     r = ToVkShaderStage(shader_info.GetShaderType(), &stage);
     if (!r.IsSuccess())
       return r;
-    const auto& name = shader_info.GetEntryPointName();
+    const auto& name = shader_info.GetEntryPoint();
     if (!name.empty()) {
       info.vk_pipeline->SetEntryPointName(stage, name);
     }
--- a/src/vulkan/engine_vulkan.cc
+++ b/src/vulkan/engine_vulkan.cc
@@ -192,6 +192,18 @@ Result EngineVulkan::CreatePipeline(amber::Pipeline* pipeline) {
 
   info.vk_pipeline = std::move(vk_pipeline);
 
+  // Set the entry point names for the pipeline.
+  for (const auto& shader_info : pipeline->GetShaders()) {
+    VkShaderStageFlagBits stage = VK_SHADER_STAGE_FLAG_BITS_MAX_ENUM;
+    r = ToVkShaderStage(shader_info.GetShaderType(), &stage);
+    if (!r.IsSuccess())
+      return r;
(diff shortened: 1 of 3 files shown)


src/vulkan/engine_vulkan.cc
```diff
@@ -192,6 +192,18 @@ Result EngineVulkan::CreatePipeline(amber::Pipeline* pipeline) {
...
   info.vk_pipeline = std::move(vk_pipeline);
 
+  // Set the entry point names for the pipeline.
+  for (const auto& shader_info : pipeline->GetShaders()) {
+    VkShaderStageFlagBits stage = VK_SHADER_STAGE_FLAG_BITS_MAX_ENUM;
+    r = ToVkShaderStage(shader_info.GetShaderType(), &stage);
+    if (!r.IsSuccess())
+      return r;
+    const auto& name = shader_info.GetEntryPoint();
+    if (!name.empty()) {
```
jaebaek: If it is empty, it would be better to just return an error here?

```
if (name.empty())
  return Result("Vulkan: pipeline entry point is empty");
info.vk_pipeline->SetEntryPointName(stage, name);
```

alan-baker: It is valid to not set an entry name if you plan to use the default name.

tests/cases/non_default_entry_point.amber
```diff
@@ -0,0 +1,75 @@
...
+         %20 = OpAccessChain %_ptr_StorageBuffer_uint %16 %uint_0 %uint_0
+         %21 = OpAccessChain %_ptr_StorageBuffer_uint %17 %uint_0 %uint_0
+         %22 = OpLoad %uint %20
+               OpStore %21 %22
+               OpReturn
+               OpFunctionEnd
+
+END
+
+BUFFER in_buf DATA_TYPE uint32 DATA
```
dj2: For future note, this could have been a single line:

```groovy
BUFFER in_buf DATA_TYPE uint32 DATA 9 END
```

tests/cases/non_default_entry_point.amber
```diff
@@ -0,0 +1,75 @@
...
+
+PIPELINE compute my_pipeline
+  ATTACH my_shader ENTRY_POINT foo
+  BIND BUFFER in_buf AS storage DESCRIPTOR_SET 0 BINDING 0
+  BIND BUFFER out_buf AS storage DESCRIPTOR_SET 0 BINDING 1
+END
+
+RUN my_pipeline 1 1 1
+
+EXPECT out_buf EQ_BUFFER in_buf
```
dj2: Awesome, nice to see EQ_BUFFER being useful for tests. @hevrard

hevrard: Glad to see this too! :+1:

PR #578, opened 2019-07-15
Support parsing OpenCL C kernels

### LINKED ISSUE #428: Add a SET command for OPENCL-C
Along with binding buffers to kernels we need to be able to set scalar arguments:

```
  out[z] = in[z];
END
BUFFER in_buf TYPE int4 DATA
7 4 2 3
END

PIPELINE my_pipeline
  ATTACH my_kernel
  SET OPENCL-C DATA KERNEL foo ARG z 2
END
RUN my_pipeline
```

This would set the z argument to 2 when the pipeline is executed. If there are multiple arguments they may need to get combined into a single POD argument based on what the clspv computer requests.

alan-baker:
Fixes #428

* New shader format: OpenCLC
  * Add parser tests
* Build fixes for clspv
  * add include director
  * add link libraries
  * add macro definition
* Add clspv helper for compiling

Note: this depends on #577 to be able to set entry points correctly because "main" is an invalid kernel name.


src/clspv_helper.cc
```diff
@@ -0,0 +1,34 @@
...
+
+namespace amber {
+namespace clspvhelper {
+
+Result Compile(const std::string& src_str,
+               std::vector<uint32_t>* generated_binary) {
+  // TODO(alan-baker): Parse the descriptor map.
+  std::vector<clspv::version0::DescriptorMapEntry> entries;
+  if (clspv::CompileFromSourceString(src_str, "", "", generated_binary,
+                                     &entries))
```
dj2: nit: add {}'s, if condition is multi-line

Issue #559, opened 2019-06-20
Dawn: handle Int16 vertex index format


dneto0:

dneto0:
Actually, WebGPU MVP only has 32-bit data types. So this is not needed.

PR #579, opened 2019-07-15
[Dawn] Update code to catch up with dawn changes

sarahM0:
fixes #552, #559


src/dawn/engine_dawn.cc
```diff
@@ -77,6 +77,10 @@ struct ComboVertexInputDescriptor {
     // cBuffers[0] has two attributes, then cBuffers[1].attributes should
     // point to &cAttributes[2]. Likewise, if cBuffers[1] has 3 attributes,
     // then cBuffers[2].attributes should point to &cAttributes[5].
+
+    // In amber-dawn, it is the vertex input descriptor is always created
```
dj2: nit: In amber-dawn, the vertex …

Issue #524, opened 2019-05-23
Add ability to convert a Recipe back to a text representation


dj2:
Currently you can pass in compiled shaders into a recipe to use during rendering. We'd like the ability to modify buffers to use during rendering. When those are done, the script being run isn't necessarily the script that was start with. We'd like the ability to take a given recipe and turn it back into an Amberscript file.

For shaders, we'd attempt to write out the spirv-asm for the shader, in the case where disassembly fails we'd write the shader as HEX data.

dj2:
@afd

dneto0:
+1 to the request to write out SPIR-V disassembly.

dj2:
Thinking about this, the implementation is going to be a bit tricky. We can have different compile options per pipeline which means, we should have different spriv disassembly results per pipeline. So, a single shader may need to be output `n` times, once per pipeline.

This means we're going to have to rename the shaders as we output them to accommodate.

dj2:
This also has ramifications for buffers. We don't store the starting state for buffers, so if you dump the script after the execution you'll get the _ending_ buffer not the _starting_ buffer as you probably want.

Either the user must dump the script before running (which is wasteful in the case where there is no issues and we don't care about having the text version) or we start storing the before state of buffers. We'd probably want to store if it's a `FILL` or other type and what data used used to create it. For `DATA` buffers we should store the initial data.

PR #582, opened 2019-07-16
Specify target version for GLSL compilation

### LINKED ISSUE #549: Set glslang environment based on amber command line arguments
GLSL shaders using subgroup operations require Vulkan 1.1/SPIR-V 1.3, but amber doesn't pass the Vulkan version through shaderc to glslang. So a shader that uses subgroup operations generates an error like "'subgroup op' : requires SPIR-V 1.3" even if the command line includes "-v 1.1".

The issue appears to be that ShaderCompiler::CompileGlsl doesn't set the target environment in its local options variable. Adding "options.SetTargetEnvironment(shaderc_target_env_vulkan, shaderc_env_version_vulkan_1_1);" as a hack fixes this. So we need code to translate the amber command line arguments to the appropriate shaderc options so glslang's client/version get set correctly.

dneto0:
Fixes #549

### DIFF
--- a/include/amber/amber.h
+++ b/include/amber/amber.h
@@ -90,14 +90,14 @@ struct Options {
   /// Holds engine specific configuration. Ownership stays with the caller.
   EngineConfig* config;
   /// The SPIR-V environment to target.
-  /// E.g. spv1.0, spv1.3, vulkan1.0, vulkan1.1spv1.4.
+  /// E.g. "spv1.0", "spv1.3", "vulkan1.0", "vulkan1.1spv1.4".
   /// If a Vulkan environment, uses the highest version of SPIR-V required
   /// to be supported by that Vulkan environment.  For SPIR-V 1.4 in
-  /// Vulkan, use vulkan1.1spv1.4.
+  /// Vulkan, use "vulkan1.1spv1.4".
   /// If a SPIR-V environment is specified, assume the lowest version
   /// of Vulkan that requires support for that version of SPIR-V.
   /// Shader compilers may limit the list of supported environments.
-  /// The default is an empty string, and is the same as spv1.0.
+  /// If empty, a default of "spv1.0" is used.
   std::string spv_env;
   /// Lists the buffers to extract at the end of the execution
   std::vector<BufferInfo> extractions;
--- a/include/amber/amber.h
+++ b/include/amber/amber.h
@@ -90,6 +90,14 @@ struct Options {
   /// Holds engine specific configuration. Ownership stays with the caller.
   EngineConfig* config;
   /// The SPIR-V environment to target.
+  /// E.g. spv1.0, spv1.3, vulkan1.0, vulkan1.1spv1.4.
+  /// If a Vulkan environment, uses the highest version of SPIR-V required
+  /// to be supported by that Vulkan environment.  For SPIR-V 1.4 in
+  /// Vulkan, use vulkan1.1spv1.4.
+  /// If a SPIR-V environment is specified, assume the lowest version
+  /// of Vulkan that requires support for that version of SPIR-V.
+  /// Shader compilers may limit the list of supported environments.
+  /// The default is an empty string, and is the same as spv1.0.
   std::string spv_env;
   /// Lists the buffers to extract at the end of the execution
   std::vector<BufferInfo> extractions;
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -170,6 +170,9 @@ if (${AMBER_ENABLE_TESTS})
   target_include_directories(amber_unittests PRIVATE
       ${gmock_SOURCE_DIR}/include)
   target_link_libraries(amber_unittests libamber gmock_main)
+  if (${AMBER_ENABLE_SHADERC})
+    target_link_libraries(amber_unittests shaderc SPIRV)
+  endif()
   amber_default_compile_options(amber_unittests)
   add_test(NAME amber_unittests COMMAND amber_unittests)
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -170,9 +170,6 @@ if (${AMBER_ENABLE_TESTS})
   target_include_directories(amber_unittests PRIVATE
       ${gmock_SOURCE_DIR}/include)
   target_link_libraries(amber_unittests libamber gmock_main)
-  if (${AMBER_ENABLE_SHADERC})
-    target_link_libraries(amber_unittests shaderc SPIRV)
-  endif()
   amber_default_compile_options(amber_unittests)
   add_test(NAME amber_unittests COMMAND amber_unittests)
 
--- a/src/shader_compiler.cc
+++ b/src/shader_compiler.cc
@@ -314,7 +314,7 @@ Result ParseSpvEnv(const std::string& spv_env,
     values = {kVulkan, kVulkan_1_1, kSpv_1_3};
   } else if (spv_env == "spv1.4") {
     values = {kVulkan, kVulkan_1_1, kSpv_1_4};
-  } else if (spv_env == "Vulkan1.0") {
+  } else if (spv_env == "vulkan1.0") {
     values = {kVulkan, kVulkan_1_0, kSpv_1_0};
   } else if (spv_env == "vulkan1.1") {
     // Vulkan 1.1 requires support for SPIR-V 1.3.
--- a/src/shader_compiler.cc
+++ b/src/shader_compiler.cc
@@ -172,6 +172,17 @@ Result ShaderCompiler::CompileGlsl(const Shader* shader,
   shaderc::Compiler compiler;
   shaderc::CompileOptions options;
 
+  uint32_t env = 0u;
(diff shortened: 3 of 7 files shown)


dneto0:
This doesn't have end-to-end testing, unfortunately.  It would be great to have https://github.com/google/amber/issues/524 in place, then we could match against the disassembled shader.

dneto0:
I made this dependent on the DEPS update in https://github.com/google/amber/pull/583.  Only pull this one in after the other has landed.

dneto0:
Rebasing.

include/amber/amber.h
```diff
@@ -90,6 +90,14 @@ struct Options {
...
   EngineConfig* config;
   /// The SPIR-V environment to target.
+  /// E.g. spv1.0, spv1.3, vulkan1.0, vulkan1.1spv1.4.
+  /// If a Vulkan environment, uses the highest version of SPIR-V required
+  /// to be supported by that Vulkan environment.  For SPIR-V 1.4 in
+  /// Vulkan, use vulkan1.1spv1.4.
+  /// If a SPIR-V environment is specified, assume the lowest version
+  /// of Vulkan that requires support for that version of SPIR-V.
+  /// Shader compilers may limit the list of supported environments.
+  /// The default is an empty string, and is the same as spv1.0.
```
dj2: Just, "The default is spv1.0" ?

src/CMakeLists.txt
```diff
@@ -170,6 +170,9 @@ if (${AMBER_ENABLE_TESTS})
   target_include_directories(amber_unittests PRIVATE
       ${gmock_SOURCE_DIR}/include)
   target_link_libraries(amber_unittests libamber gmock_main)
+  if (${AMBER_ENABLE_SHADERC})
+    target_link_libraries(amber_unittests shaderc SPIRV)
```
dj2: I wonder how this worked before ....

src/shader_compiler.cc
```diff
@@ -241,4 +252,68 @@ Result ShaderCompiler::CompileOpenCLC(const Shader*,
...
 #endif  // AMBER_ENABLE_CLSPV
 
+Result ParseSpvEnv(const std::string& spv_env,
+                   uint32_t* target_env,
+                   uint32_t* target_env_version,
+                   uint32_t* spirv_version) {
+  if (!target_env || !target_env_version || !spirv_version)
+    return Result("ParseSpvEnv: null pointer parameter");
+
+  const uint32_t vulkan = 0;
```
dj2: Move all these to the anonymous namespace and make constants.

src/shader_compiler.cc
```diff
@@ -241,4 +252,68 @@ Result ShaderCompiler::CompileOpenCLC(const Shader*,
...
+
+#if AMBER_ENABLE_SHADERC
+  static_assert(vulkan == shaderc_target_env_vulkan, "enum value mismatch");
+  static_assert(vulkan_1_0 == shaderc_env_version_vulkan_1_0,
+                "enum value mismatch");
+  static_assert(vulkan_1_1 == shaderc_env_version_vulkan_1_1,
+                "enum value mismatch");
+  static_assert(spv_1_0 == shaderc_spirv_version_1_0, "enum value mismatch");
+  static_assert(spv_1_1 == shaderc_spirv_version_1_1, "enum value mismatch");
+  static_assert(spv_1_2 == shaderc_spirv_version_1_2, "enum value mismatch");
```
dj2: If the above move up, this can be done up in the anonymous namespace as well.

src/shader_compiler.cc
```diff
@@ -241,4 +252,68 @@ Result ShaderCompiler::CompileOpenCLC(const Shader*,
...
+  static_assert(spv_1_3 == shaderc_spirv_version_1_3, "enum value mismatch");
+  static_assert(spv_1_4 == shaderc_spirv_version_1_4, "enum value mismatch");
```
dj2: Should this say: "enum spv1.4 value mismatch"? Same for others?

src/shader_compiler.cc
```diff
@@ -241,4 +252,68 @@ Result ShaderCompiler::CompileOpenCLC(const Shader*,
...
+    values = {vulkan, vulkan_1_1, spv_1_4};
+  } else if (spv_env == "vulkan1.0") {
+    values = {vulkan, vulkan_1_0, spv_1_0};
+  } else if (spv_env == "vulkan1.1") {
+    // Vulkan 1.1 requires support for SPIR-V 1.3.
+    values = {vulkan, vulkan_1_1, spv_1_3};
+  } else if (spv_env == "vulkan1.1spv1.4") {
+    values = {vulkan, vulkan_1_1, spv_1_4};
+  } else {
+    return Result(std::string("Unrecognized environment") + spv_env);
```
dj2: Space after "environment" ? Won't this end up being "Unrecognized environmentspv1.4" ?

src/shader_compiler.h
```diff
@@ -50,6 +50,15 @@ class ShaderCompiler {
   std::string spv_env_;
 };
 
+// Parses the SPIR-V environment string, and returns three values
+// corresponding to Shaderc's values for shaderc_target_env,
```
dj2: The Shaderc bit isn't really relevant.

Parses the SPIR-V environment string, and returns the corresponding
|target_env|, |target_env_version| and |spirv_version|. Returns a
failure value if the |spv_env| string is invalid.

src/shader_compiler_test.cc
```diff
@@ -216,4 +220,68 @@ TEST_F(ShaderCompilerTest, ReturnsCachedShader) {
...
+
+using ParseSpvEnvTest = ::testing::TestWithParam<ParseSpvEnvCase>;
+
+TEST_P(ParseSpvEnvTest, Samples) {
+  uint32_t target_env = 42u;
+  uint32_t env_version = 43u;
+  uint32_t spirv_version = 44u;
+  auto r = amber::ParseSpvEnv(GetParam().env_str, &target_env, &env_version,
+                              &spirv_version);
+  if (GetParam().ok) {
```
dj2:
```
if (!GetParam().ok) {
  EXPECT_FALSE(r.IsSuccess());
  return;
}
```
?

src/shader_compiler_test.cc
```diff
@@ -216,4 +220,68 @@ TEST_F(ShaderCompilerTest, ReturnsCachedShader) {
...
+const uint32_t vulkan_1_1 = ((uint32_t(1) << 22) | (1 << 12));
+const uint32_t spv_1_0 = uint32_t(0x10000);
+const uint32_t spv_1_1 = uint32_t(0x10100);
+const uint32_t spv_1_2 = uint32_t(0x10200);
+const uint32_t spv_1_3 = uint32_t(0x10300);
+const uint32_t spv_1_4 = uint32_t(0x10400);
+
+INSTANTIATE_TEST_SUITE_P(ParseSpvEnvFailures,
+                         ParseSpvEnvTest,
+                         ::testing::ValuesIn(std::vector<ParseSpvEnvCase>{
```
dj2: Since the tests are separate, would it be easier to just make to methods, one for the false and one for the true and remove the 2nd' param?

include/amber/amber.h
dneto0: A default-constructed Options object will have an empty string here.
How about
" If empty, a default of "spv1.0" is used."

dj2: sgtm

src/CMakeLists.txt
dneto0: Hm.. I don't actually need this.
I thought I did so I could access the shaderc/env.h header.  But I don't!  Will remove

src/shader_compiler_test.cc
dneto0: I like mine better.  No reason to have an early return; early return muddies the cases IMHO.

src/shader_compiler_test.cc
dneto0: Then I'd need to have two different test classes. I don't think that's worth the trouble.

src/CMakeLists.txt
dneto0: Removed

src/shader_compiler.cc
dneto0: Done

src/shader_compiler.cc
dneto0: done

src/shader_compiler.cc
dneto0: done

src/shader_compiler.cc
dneto0: done

src/shader_compiler.h
dneto0: done

src/shader_compiler_test.cc
dj2: k

src/y
```diff
@@ -0,0 +1,4 @@
+Usage
```
dj2: ...

dneto0: ugh

PR #584, opened 2019-07-19
Bind opencl buffers

### LINKED ISSUE #426: Add BIND BUFFER modifier to allow binding OPENCL-C arguments
Something along the lines of:

```
  out[z] = in[z];
END
BUFFER in_buf TYPE int4 DATA
7 4 2 3
END

PIPELINE my_pipeline
  ATTACH my_kernel
END
RUN my_pipeline
EXPECT out_buf IDX 0 EQ 7 4 2 3
```

This must also correctly detect ubo vs ssbo. The descriptorSet/Binding should be determined by the data provided by the CLSPV compiler.

alan-baker:
Fixes #426.

I went with a slightly different format than in the issue to better match the current grammar.
```
BIND BUFFER {buffer_name} AS {buffer_type} KERNEL ARG _name_
BIND BUFFER {buffer_name} AS {buffer_type} KERNEL ARGNO _number_
```
As the commands suggest, setting the buffer type is not currently handled, but could be supported in the future.

The descriptor map is (partially) parsed into the pipeline's shader info. This is used to update descriptor set and binding information before creating the pipelines. A different format than Clspv's data structure is used to avoid extra coupling between the two.


dj2:
For the syntax, I think I'd prefer to see `ARG_NAME` and `ARG_NUMBER` be explicit. It's going to be easy to forget that `ARG` means name implicity. `ARG_NO` makes me think there is no argument.

alan-baker:
> For the syntax, I think I'd prefer to see `ARG_NAME` and `ARG_NUMBER` be explicit. It's going to be easy to forget that `ARG` means name implicity. `ARG_NO` makes me think there is no argument.

Done

dneto0:
I'm confused.  Doesn't the descriptor map determine the buffer type?  Seems overdetermined to have to declare it in the pipeline as well? Or is it that compiler line arguments for the specific pipeline will determine the buffer type and Amber has to check consistency?

src/amberscript/parser_bind_test.cc
```diff
@@ -1149,5 +1149,147 @@ END)";
...
+
+PIPELINE compute my_pipeline
+  ATTACH my_shader
+  BIND BUFFER my_buf AS storage ARG_NAME arg
+END)";
+
+  Parser parser;
+  Result r = parser.Parse(in);
+  ASSERT_FALSE(r.IsSuccess());
+  EXPECT_EQ("9: missing DESCRIPTOR_SET for BIND command", r.Error());
```
dneto0: Yah. this is a bad message for this case.

src/amberscript/parser_bind_test.cc
```diff
@@ -1149,5 +1149,147 @@ END)";
...
+
+PIPELINE compute my_pipeline
+  ATTACH my_shader
+  BIND BUFFER my_buf AS storage KERNEL ARG_NUMBER in
+END)";
+
+  Parser parser;
+  Result r = parser.Parse(in);
+  ASSERT_FALSE(r.IsSuccess());
+  EXPECT_EQ("9: expected argument identifier number", r.Error());
```
dneto0: I'd drop the "identifier" here.

alan-baker:
> I'm confused. Doesn't the descriptor map determine the buffer type? Seems overdetermined to have to declare it in the pipeline as well? Or is it that compiler line arguments for the specific pipeline will determine the buffer type and Amber has to check consistency?

Right now you have to get it right. Since compiler options are not exposed yet, everything will be a storage buffer. As I mentioned in the PR description I was planning to handle buffer type in a subsequent PR.

alan-baker:
I went ahead and made parsing the storage class optional for opencl kernel args.

docs/amber_script.md
```diff
@@ -254,6 +254,16 @@ attachment content, depth/stencil content, uniform buffers, etc.
...
   BIND SAMPLER {sampler_name} DESCRIPTOR_SET _id_ BINDING _id_
+
+  # Bind OpenCL argument buffer by name. Specifying the buffer type is optional.
+  # Amber will set the type as appropriate for the argument buffer. All uses
+  # of the buffer must be across all pipelines.
+  BIND BUFFER {buffer_name} [AS {buffer_type}] KERNEL ARG_NAME _name_
+
+  # Bind OpenCL argument buffer by argument ordinal. Specifying the buffer type
+  # is optional. Amber will set the type as appropriate for the argument
+  # buffer. All uses of the buffer must be across all pipelines.
```
dneto0: Should say if it's 0-based.

dneto0: "all uses of the buffer must be ... ?consistent? ... across all pipelines"?

Do you mean: the inferred buffer type must be consistent across all pipelines?

(Because the descriptor set and binding can differ, it's important to say what aspect is consistent.)

PR #585, opened 2019-07-19
[Dawn] Implement DoCompute support in dawn backend and fix sparse descriptor set support

### LINKED ISSUE #21: Dawn: Implement DoBuffer

### LINKED ISSUE #573: Dawn: Add support for doBuffer with sparse descriptor sets
layout (set = 0, binding = 0) uniform vertexUniformBuffer0 {
            mat2 transform0;
        };
        layout (set = 2, binding = 0) uniform vertexUniformBuffer2 {
            mat2 transform2;
        };

It depends, in the SPIR-V code it should work at the moment, but we'll need the pipeline layout to contain an empty bind group layout at position 1, and we'll need an empty bindgroup on this empty bindgrouplayout set before we can draw. This is a limitation of Dawn today but will eventually disappear and won't be present in WebGPU.

### LINKED ISSUE #575: Resizing ssbo buffers to the same value increases their size
This issue came out of MR #564 - for more info read the comments on it.
This issue is causing three tests to fail in dawn backend (theses have multiple ssbo calls for the same buffer as a result that buffer's size gets multiplied till it gets so big it overflows and dawn backed asserts for creating a buffer of size 0):
./cases/draw_array_after_draw_rect.vkscript
./cases/draw_rect_after_draw_array.vkscript
./cases/draw_rect_and_draw_array_mixed.vkscript

sarahM0:
fixes #21 
fixes #573 
fixes #575


sarahM0:
My change in command_parser.cc is causing two unit tests to break. I temporary fixed them by relaxing assertions and I'm looking for the real issue.

src/dawn/engine_dawn.cc
```diff
@@ -345,6 +343,49 @@ MapResult MapTextureToHostBuffer(const RenderPipelineInfo& render_pipeline,
...
 
+// Maps device buffers back to the host buffers
+Result MapDeviceBufferToHostBuffer(const ComputePipelineInfo& compute_pipeline,
+                                   const ::dawn::Device& device) {
+  for (uint32_t i = 0; i < compute_pipeline.pipeline->GetBuffers().size();
+       i++) {
+    auto& device_buffer = compute_pipeline.buffers[i];
+    auto& host_buffer = compute_pipeline.pipeline->GetBuffers()[i];
+
+    // Create a copy of device buffer to it in a map read operation.
```
dneto0: Missing word here?  "... to _use_ it in a map read operation"

src/dawn/engine_dawn.cc
```diff
@@ -345,6 +343,49 @@ MapResult MapTextureToHostBuffer(const RenderPipelineInfo& render_pipeline,
...
+                               destination_offset, copy_size);
+    auto commands = encoder.Finish();
+    auto queue = device.CreateQueue();
+    queue.Submit(1, &commands);
+
+    MapResult mapped_device_buffer = MapBuffer(device, copy_device_buffer);
+    auto* values = host_buffer.buffer->ValuePtr();
+    values->resize(host_buffer.buffer->GetSizeInBytes());
+    // TODO(sarahM0): At the moment all compute tests pass. Investigate
+    // if ever alignment is an issue and if it's necessary to copy line by line
```
dneto0: - won't need to copy line by line because buffers are a big linear sequence of bytes. There's no enforced stride because there is no intrinsic "line".
- memcpy should work with byte-aligned inputs or outputs.

src/dawn/engine_dawn.cc
```diff
@@ -345,6 +343,49 @@ MapResult MapTextureToHostBuffer(const RenderPipelineInfo& render_pipeline,
...
+    auto encoder = device.CreateCommandEncoder();
+    encoder.CopyBufferToBuffer(device_buffer, source_offset, copy_device_buffer,
+                               destination_offset, copy_size);
+    auto commands = encoder.Finish();
+    auto queue = device.CreateQueue();
+    queue.Submit(1, &commands);
+
+    MapResult mapped_device_buffer = MapBuffer(device, copy_device_buffer);
+    auto* values = host_buffer.buffer->ValuePtr();
+    values->resize(host_buffer.buffer->GetSizeInBytes());
```
dneto0: I haven't read the rest of the code, but this is right only if the underlying vector is vector-of-8bit-ints. Otherwise this is wasteful.  This all depends on what the buffer format was assumed to be.

src/dawn/engine_dawn.cc
```diff
@@ -1166,8 +1212,42 @@ Result EngineDawn::DoDrawArrays(const DrawArraysCommand* command) {
...
+  if (!compute_pipeline)
+    return Result("DoComput: invoked on invalid or missing compute pipeline");
+
+  ::dawn::ComputePipelineDescriptor computePipelineDescriptor;
+  computePipelineDescriptor.layout = MakeBasicPipelineLayout(
+      device_->Get(), compute_pipeline->bind_group_layouts);
+
+  ::dawn::PipelineStageDescriptor pipelineStageDescriptor;
+  pipelineStageDescriptor.module = compute_pipeline->compute_shader;
+  pipelineStageDescriptor.entryPoint = "main";
```
dneto0: Vulkan allows an empty string here, which then defaults to "main".
I think this is ok. No request to change.

src/dawn/engine_dawn.cc
```diff
@@ -1387,5 +1480,95 @@ Result EngineDawn::AttachBuffersAndTextures(
...
+        return Result("AttachBuffers: unknown buffer type: " +
+                      std::to_string(static_cast<uint32_t>(
+                          buf_info.buffer->GetBufferType())));
+        break;
+      }
+    }
+
+    if (buf_info.descriptor_set > kMaxDawnBindGroup - 1) {
+      return Result(
+          "AttachBuffers: Dawn has maximum of 4 bindGroups "
```
dneto0: The 4 is magic here, and could go out of date.

Do this:

`std::string("AttachBuffers: Dawn has a maximum of ") + std::to_string(kMaxDawnBindGroup) + " bind groups `....

src/dawn/engine_dawn.cc
```diff
@@ -1387,5 +1480,95 @@ Result EngineDawn::AttachBuffersAndTextures(
...
+    layout_info.type = bindingType;
+    layouts_info[buf_info.descriptor_set].push_back(layout_info);
+
+    BindingInitializationHelper tempBinding = BindingInitializationHelper(
+        buf_info.binding, compute_pipeline->buffers.back(), 0,
+        buf_info.buffer->GetSizeInBytes());
+    bindingInitalizerHelper[buf_info.descriptor_set].push_back(tempBinding);
+  }
+
+  // TODO(sarahM0): fix issue: Add support for doBuffer with sparse descriptor
```
dneto0: Good comment!

src/vkscript/command_parser.cc
```diff
@@ -789,6 +789,21 @@ Result CommandParser::ProcessUniform() {
...
+                         static_cast<uint32_t>(values.size());
+  // The buffer should only be resized to become bigger. This means that if a
+  // command was run to set the buffer size we'll honour that size until a
+  // request happens to make the buffer bigger.
+  if (value_count > buf->ValueCount())
+    buf->SetValueCount(value_count);
+
+  // Even if the value count doesn't change, the buffer is still resized
+  // because this maybe the first time data is set into the buffer.
+  buf->ResizeTo(buf->GetSizeInBytes());
```
dneto0: Something is fishy.  ResizeTo says to resize the buffer to hold |element_count| elements.  How do we know the element size is 1 byte?

src/dawn/engine_dawn.cc
sarahM0: ok makes sense

src/dawn/engine_dawn.cc
sarahM0: Dawn isn't as forgiving. If I do not specify the entry point (remove this line) it errors out:
Device error: Entry point must be "main"

src/dawn/engine_dawn.cc
sarahM0: I think it always is:
since:
auto* values = host_buffer.buffer->ValuePtr();
and:
  std::vector<uint8_t>* ValuePtr() { return &bytes_; }

src/vkscript/command_parser.cc
sarahM0: I added ResizeToBytes() function to buffer class. That should resolve confusion.

src/buffer.h
```diff
@@ -150,6 +150,11 @@ class Buffer {
   /// initial count. This requires the format to have been set.
   void ResizeTo(uint32_t element_count);
 
+  /// Resizes the buffer to hold |element_count| elements. This requires the
+  /// format to have been set. This is separate from ResizeTo(). The given
+  /// argument here is |size_in_bytes| bytes vs |element_count| elements
+  void ResizeToBytes(uint32_t size_in_bytes);
```
dj2: Should this be called SetSizeInBytes and be put up above with GetSizeInBytes? Not really sure, I named `ResizeTo` poorly .....

src/dawn/engine_dawn.cc
```diff
@@ -1518,9 +1516,8 @@ Result EngineDawn::AttachBuffers(ComputePipelineInfo* compute_pipeline) {
     }
 
     if (buf_info.descriptor_set > kMaxDawnBindGroup - 1) {
-      return Result(
-          "AttachBuffers: Dawn has maximum of 4 bindGroups "
-          "(descriptor sets)");
+      return Result("AttachBuffers: Dawn has a maximum of " +
+                    std::to_string(kMaxDawnBindGroup) + "(descriptor sets)");
```
dj2: nit: space before the ( in "(descriptor sets" otherwise you'll get `4(descriptor sets` instead of `4 (descriptor sets)` as the error.

src/buffer.h
```diff
@@ -150,6 +150,11 @@ class Buffer {
   /// initial count. This requires the format to have been set.
   void ResizeTo(uint32_t element_count);
```
dj2: Maybe we should file a bug to rename this `SetElementCount` or `SetSizeToCountElements` or something ? ...

src/buffer.h
```diff
@@ -150,6 +150,11 @@ class Buffer {
...
 
+  /// Resizes the buffer to hold |element_count| elements. This requires the
```
dj2: Update comment to match as this doesn't work on element_count.

src/dawn/engine_dawn.cc
sarahM0: good eye!

src/buffer.h
sarahM0: I updated the comment, let know if you'd like to add anything specific

src/buffer.h
sarahM0: There is a SetElementCount already:
  /// Sets the number of elements in the buffer.
  void SetElementCount(uint32_t count) { element_count_ = count; }
I filed issue #597

src/vkscript/command_parser_test.cc
```diff
@@ -3339,7 +3339,7 @@ TEST_F(CommandParserTest, Uniform) {
   const auto* buf = cmd->GetBuffer();
   const auto* values = buf->GetValues<float>();
   std::vector<float> results = {2.1f, 3.2f, 4.3f, 0.f};
-  ASSERT_EQ(results.size(), buf->ValueCount());
+  EXPECT_TRUE(results.size() < buf->ValueCount());
```
dj2: This shouldn't be needed? Something has gone wrong if we have more then 4 values.

src/vkscript/command_parser_test.cc
```diff
@@ -3381,7 +3381,7 @@ TEST_F(CommandParserTest, UniformWithContinuation) {
   const auto* buf = cmd->GetBuffer();
   const auto* values = buf->GetValues<float>();
   std::vector<float> results = {2.1f, 3.2f, 4.3f, 0.f, 5.4f, 6.7f, 8.9f, 0.f};
-  ASSERT_EQ(results.size(), buf->ValueCount());
+  EXPECT_TRUE(results.size() < buf->ValueCount());
```
dj2: ditto

src/vkscript/command_parser_test.cc
sarahM0: buf->ValueCount()
= element_count_ * format_->ValuesPerElement()
= element_count_ * ValuesPerRow() * column_count_ && (is_std140_ && column_count_ > 1)
= element_count_ * 4 * column_count_
= element_count_ * 4 * 1
= element_count_ * 4
As a result:
with my change in command_parser.cc: element_count_ == 3 -> buf->ValueCount() == 12
without it: element_count_ == 1 -> buf->ValueCount() == 4

I'm a little lost! which element_count is correct for a "uniform vec3 32 2.1 3.2 4.3"?

dj2: So, the distinction between element and value. A vec3 is 1 element, but it's 4 values (4, not 3 due to the layout rules padding the last item). So, a `vec3 32 2.1 3.2 4.3` is a single element with 4 values.

The `element count * format->ValuesPerElement()` should give the correct number of values.

I'd actually expect that to return `12` instead of `9`. The offset of 32 bytes of float's means we have 2 elements, or 8 values in the offset, plus 4 values that are being appended.

sarahM0: I'm finding these functions a little confusing. Let's go back to the initial problem: Dawn need the  size to create a buffer. 
Do you have any suggestion about how to set it when there a ubo command in a vkscript?

dj2: So, it sounds like this is going to get tricky with VkScript and Dawn. In VkScript, the buffers can grow as the program executes. With AmberScript the buffers are static in size. There is no way to adjust the size of a buffer in Dawn? Can we, as the buffers change size delete the old ones and create new ones?

sarahM0: Dawn CreateBuffer API needs the size as an argument. Later on we can do SetSubData however it shouldn't over for the initial given size.
There is not DeleteBuffer and since we bind buffers to the pipeline we cannot just abandon the previous one and create a new one each time.

dj2: Is this a current Dawn limitation they're going to fix, or are buffers just immutable in size in Dawn and that's the way it will always be?

I guess currently we'll have to make sure the buffer is always resized to the largest size needed. That being said, for this test, the size should be known to be exact, we shouldn't need to make the comparison weaker.

dj2: Had a thought on the train this morning. When we're parsing the VkScript file should we maintain 2 sizes in the Buffer class. The `SizeInBytes` would be the initial size to make the buffer and a `MaxSizeInBytes` would be the total size needed to hold the buffer over all `ssbo size` and `ssbo subdata` size calls?

Then, the Vulkan side stays the same and uses the size and resizes as needed and the Dawn side can use the max size to set the buffer?

The buffer would initially be sized to whatever `SizeInBytes` comes out to be.

sarahM0: taking a look

src/vkscript/command_parser.cc
```diff
@@ -627,15 +627,15 @@ Result CommandParser::ProcessSSBO() {
...
-    buf->SetSizeInBytes(buf->GetSizeInBytes());
+    uint32_t element_count = value_count;
+    if (buf->GetFormat()->GetPackSize() == 0) {
+      // This divides by the needed input values, not the values per element.
+      // The assumption being the values coming in are read from the input,
+      // where components are specified. The needed values maybe less then the
+      // values per element.
+      element_count = value_count / buf->GetFormat()->InputNeededPerElement();
+    }
+    buf->SetMaxSizeInBytes(element_count * buf->GetFormat()->SizeInBytes());
```
dneto0: This new implementation allows the max-size-in-bytes to shrink.
I think that can happen if you have an ssbo subdata command with a big offset, then a later one with a small offset. That's because value_count computed a few lines above takes into account the offset of *this* command but not the prior size.

src/buffer.h
```diff
@@ -157,6 +157,15 @@ class Buffer {
   /// |element_count| elements
   void SetSizeInBytes(uint32_t size_in_bytes);
 
+  /// max_size_in_bytes_ is the total size in bytes needed to hold the buffer
```
dneto0: This is a _definition_ of the max_size_in_bytes_ field, which should go just before the member declaration at line 200.  

Separately, there's a fragile disconnect between setting max_size_in_bytes_, element_count_ and resizing the bytes_ vector.   No single one of them is the source of truth.
You'll need to write comments at their member declarations to define exactly what they mean, and therefore determine the single source of truth.  To be clear, this is a continuation of a previously existing problem with this class definition.

src/vkscript/command_parser.cc
```diff
@@ -789,20 +789,18 @@ Result CommandParser::ProcessUniform() {
...
-  // The buffer should only be resized to become bigger. This means that if a
-  // command was run to set the buffer size we'll honour that size until a
-  // request happens to make the buffer bigger.
-  if (value_count > buf->ValueCount())
-    buf->SetValueCount(value_count);
-
-  // Even if the value count doesn't change, the buffer is still resized
-  // because this maybe the first time data is set into the buffer.
-  buf->SetSizeInBytes(buf->GetSizeInBytes());
+  uint32_t element_count = value_count;
```
dneto0: This is a repeat of the code sequence at the SSBO routine.  That's a code smell that there's a common functionality should be refactored.  By the look of it, it seems like all this should be a method on the buffer itself.
In fact, after some digging, there is a similar calculation in Buffer::SetDataWithOffset.
(Hmmm.  But that seems to not accomodate the Pack case... ?)

Suggest defining
Buffer::ResizeForDataWithOffset(const std::vector<Value>& data, uint32_t offset);
This mirrors Buffer::SetDataWithOffset.

src/vkscript/command_parser_test.cc
```diff
@@ -3339,7 +3339,7 @@ TEST_F(CommandParserTest, Uniform) {
   const auto* buf = cmd->GetBuffer();
   const auto* values = buf->GetValues<float>();
   std::vector<float> results = {2.1f, 3.2f, 4.3f, 0.f};
-  EXPECT_TRUE(results.size() < buf->ValueCount());
+  EXPECT_EQ(results.size(), buf->ValueCount());
```
dneto0: I'm glad this is sorted out.

src/vkscript/command_parser.cc
sarahM0: You're right I forgot about this condition:
    if(buf->GetMaxSizeInBytes() > element_count * buf->GetFormat()->SizeInBytes())
      buf->SetMaxSizeInBytes(element_count * buf->GetFormat()->SizeInBytes());
Does this fix the issue you've mentioned?

src/buffer.h
sarahM0: I can predict this can get complicated so I filed issue #599

src/vkscript/command_parser.cc
sarahM0: It's not really resizing though. It is setting a fake size to be used in Dawn buffer creation.

src/vkscript/command_parser_test.cc
sarahM0: fixed

tests/cases/multiple_ssbo_with_sparse_descriptor_set_in_compute_pipeline_less_than_4.vkscript
```diff
@@ -0,0 +1,116 @@
+# Copyright 2018 The Amber Authors.
```
dj2: nit: 2019

src/buffer.h
```diff
@@ -150,10 +150,29 @@ class Buffer {
   /// initial count. This requires the format to have been set.
   void ResizeTo(uint32_t element_count);
 
+  /// Resizes the buffer to hold |size_in_bytes| over format_->SizeInBytes()
+  /// number of elements while resizing the buffer to |size_in_bytes| bytes.
+  /// This requires the format to have been set. This is separate from
+  /// ResizeTo() since the given argument here is |size_in_bytes| bytes vs
```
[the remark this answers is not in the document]
dneto0: meh, I'm ok with what is written here.   It's better to err with too much info than with too little. It's *good* that the parameter name has the units in it.

src/buffer.h
```diff
@@ -150,10 +150,29 @@ class Buffer {
...
+  /// |element_count| elements
+  void SetSizeInBytes(uint32_t size_in_bytes);
+
+  /// Sets the max_size_in_bytes_ to |max_size_in_bytes| bytes
+  void SetMaxSizeInBytes(uint32_t max_size_in_bytes);
+  /// Returns max_size_in_bytes_ if it is not zero. Otherwise it means this
```
dneto0: Good comment.

src/dawn/pipeline_info.h
```diff
@@ -82,6 +82,17 @@ struct ComputePipelineInfo {
...
   ::amber::Pipeline* pipeline = nullptr;
   ::dawn::ShaderModule compute_shader;
+
+  // storage and uniform buffers
+  std::vector<::dawn::Buffer> buffers;
+
+  std::vector<::dawn::BindGroup> bind_groups;
+  std::vector<::dawn::BindGroupLayout> bind_group_layouts;
+
+  // Mapping from the <descriptor_set, binding> to dawn buffer index in buffers
```
dneto0: Good and necessary comment.

src/buffer.h
```diff
@@ -150,10 +150,29 @@ class Buffer {
...
+  /// buffer is an amber buffer which has a fix size and returns
+  /// GetSizeInBytes()
+  uint32_t GetMaxSizeInBytes() const;
+
   /// Write |data| into the buffer |offset| bytes from the start. Write
   /// |size_in_bytes| of data.
   Result SetDataWithOffset(const std::vector<Value>& data, uint32_t offset);
 
+  /// At each ubo, ssbo size and ssbo subdata size calls, recalculates
```
dneto0: All together, these size-related changes make a *lot* more sense to me.  The new method name here is a big improvement.

src/dawn/engine_dawn.cc
```diff
@@ -292,6 +292,110 @@ ::dawn::TextureCopyView CreateTextureCopyView(::dawn::Texture texture,
...
+    // kMinimumImageRowPitch bytes, so it might have padding therefore memcpy
+    // is done row by row.
+    for (uint h = 0; h < height; h++) {
+      std::memcpy(values->data() + h * row_stride,
+                  static_cast<const uint8_t*>(mapped_device_texture.data) +
+                      h * dawn_row_pitch,
+                  row_stride);
+    }
+    // Always unmap the buffer at the end of the engine's command.
+    copy_buffer.Unmap();
```
dneto0: Note:  It's slightly fragile to have the unmap here when you have an early return earlier in the body of the loop.  Fortunately that's an error condition that will cause the entire script to fail, so it's still correct.
There are patterns for writing code to ensure that cleanups occur even in the presence of an early exit.  (e.g. a std::unique_ptr can be used to hold an allocated object that has to be released in any case when leaving the scope.) . Amber hasn't got such a mechanism in general and in this case it would only obscure the logic of the code.

This is just a FYI to keep an eye out for that kind of fragility.

No request to change anything here.

src/dawn/engine_dawn.cc
dneto0: I asked, and it turns out this is a temporary limitation in Dawn.  So I filed an issue to enhance this in the future. https://github.com/google/amber/issues/607

src/dawn/engine_dawn.h
```diff
@@ -68,15 +68,27 @@ class EngineDawn : public Engine {
...
+      const ::amber::PipelineCommand* command) {
+    return pipeline_map_[command->GetPipeline()].compute_pipeline.get();
+  }
   // Creates and attaches index, vertex, storage, uniform and depth-stencil
   // buffers. Sets up bindings. Also creates textures and texture views if not
-  // created yet.
+  // created yet. Used in the Graphics pipeline creation.
   Result AttachBuffersAndTextures(RenderPipelineInfo* render_pipeline);
+  // Creates and attaches index, vertex, storage, uniform and depth-stencil
+  // buffers.Used in the Compute pipeline creation.
```
dneto0: Nit: extra space after the period and before "Used..."

No need to block this MR

src/vkscript/command_parser.cc
```diff
@@ -621,21 +621,7 @@ Result CommandParser::ProcessSSBO() {
...
-    // The buffer should only be resized to become bigger. This means that if a
-    // command was run to set the buffer size we'll honour that size until a
-    // request happens to make the buffer bigger.
-    if (value_count > buf->ValueCount())
-      buf->SetValueCount(value_count);
-
-    // Even if the value count doesn't change, the buffer is still resized
-    // because this maybe the first time data is set into the buffer.
-    buf->ResizeTo(buf->GetSizeInBytes());
+    buf->RecalculateMaxSizeInBytes(values, cmd->GetOffset());
```
dneto0: Much nicer

src/dawn/engine_dawn.cc
sarahM0: Thanks for explaining this!

src/buffer.h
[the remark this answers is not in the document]
dj2: I agree, the variable name is right, adding the _bytes fixed some prior confusion with things. Was referring to the second bytes in the sentence.

src/dawn/engine_dawn.cc
dj2: We have a bit of this in the vulkan engine code. I created wrapper objects for command buffer and but they aren't great.  Hard to make nice wrappers for this stuff, heh.

src/buffer.cc
```diff
@@ -223,6 +243,23 @@ void Buffer::ResizeTo(uint32_t element_count) {
...
+}
+
+void Buffer::SetMaxSizeInBytes(uint32_t max_size_in_bytes) {
+  max_size_in_bytes_ = max_size_in_bytes;
+}
+
+uint32_t Buffer::GetMaxSizeInBytes() const {
+  if (max_size_in_bytes_ != 0)
+    return max_size_in_bytes_;
+  else
```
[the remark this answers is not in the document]
sarahM0: We set max_size_in_bytes_ in processSSBO/UBO while parsing, which means it stays at 0 for amber buffers. That why I return GetSizeInBytes() if it was zero (the comment in the header tries to explain this).

Issue #427, opened 2019-03-28
Add RUN support for OPENCL-C shaders


dj2:
The run command can currently either draw or execute compute kernels. What changes need to be done to support opencl-c?

```
SHADER opencl-c my_kernel CLSPV
kernel void foo(global int4* in, global int4* out, int z) {
  out[z] = in[z];
END
BUFFER in_buf TYPE int4 DATA
7 4 2 3
END
BUFFER out_buf TYPE int4 SIZE 1

PIPELINE my_pipeline
  ATTACH my_kernel
  BIND BUFFER foo OPENCL-C KERNEL in_buf ARG in
  BIND BUFFER foo OPENCL-C KERNEL out_buf ARG out
END
RUN my_pipeline  # Just this? Or something else?
```

dj2:
@alan-baker Does RUN need to take extra params for clspv?

alan-baker:
Number of workgroups

alan-baker:
Number of workgroups is supported already. The only additional thing I can think might be appropriate here is changing the POD data args that were set via the SET KERNEL command.

dj2:
I think we have the parts of this in place, I'm going to close and we can open additional issues if needed.

PR #589, opened 2019-07-22
Opencl set for kernel args

### LINKED ISSUE #428: Add a SET command for OPENCL-C

alan-baker:
Fixes #428 

Add support for setting plain-old-data kernel args in pipeline commands. Format differs slightly from issue to better match other commands (like shader specialization):
```
PIPELINE compute my_pipeline
  ATTACH my_shader ENTRY_POINT foo
  SET KERNEL ARG_NAME _name_ AS {data_type} _val_
  SET KERNEL ARG_NUMBER _number_ AS {data_type} _val_
END
```

Before creation of the pipelines by the engine, Amber will generate and populate these special buffers. Each pipeline owns its specific PoD buffers.

Currently there is no support for updating the buffer values into between RUN commands. I would defer that need to #427.

### DIFF
--- a/src/amberscript/parser_pipeline_set_test.cc
+++ b/src/amberscript/parser_pipeline_set_test.cc
@@ -26,6 +26,7 @@ SHADER compute my_shader OPENCL-C
 #shader
 END
 PIPELINE compute my_pipeline
+  ATTACH my_shader
   SET ARG_NAME a AS uint32 0
 END
 )";
@@ -33,7 +34,7 @@ END
   Parser parser;
   auto r = parser.Parse(in);
   ASSERT_FALSE(r.IsSuccess());
-  EXPECT_EQ("6: missing KERNEL in SET command", r.Error());
+  EXPECT_EQ("7: missing KERNEL in SET command", r.Error());
 }
 
 TEST_F(AmberScriptParserTest, OpenCLSetMissingArgName) {
@@ -42,6 +43,7 @@ SHADER compute my_shader OPENCL-C
 #shader
 END
 PIPELINE compute my_pipeline
+  ATTACH my_shader
   SET KERNEL a AS uint32 0
 END
 )";
@@ -49,7 +51,7 @@ END
   Parser parser;
   auto r = parser.Parse(in);
   ASSERT_FALSE(r.IsSuccess());
-  EXPECT_EQ("6: expected ARG_NAME or ARG_NUMBER", r.Error());
+  EXPECT_EQ("7: expected ARG_NAME or ARG_NUMBER", r.Error());
 }
 
 TEST_F(AmberScriptParserTest, OpenCLSetMissingArgIdentifier) {
@@ -58,6 +60,7 @@ SHADER compute my_shader OPENCL-C
 #shader
 END
 PIPELINE compute my_pipeline
+  ATTACH my_shader
   SET KERNEL ARG_NAME AS uint32 0
 END
 )";
@@ -65,7 +68,7 @@ END
   Parser parser;
   auto r = parser.Parse(in);
   ASSERT_FALSE(r.IsSuccess());
-  EXPECT_EQ("6: missing AS in SET command", r.Error());
+  EXPECT_EQ("7: missing AS in SET command", r.Error());
 }
 
 TEST_F(AmberScriptParserTest, OpenCLSetMissingArgIdentifierNumber) {
@@ -74,6 +77,7 @@ SHADER compute my_shader OPENCL-C
 #shader
 END
 PIPELINE compute my_pipeline
+  ATTACH my_shader
   SET KERNEL ARG_NUMBER AS uint32 0
 END
 )";
@@ -81,7 +85,7 @@ END
   Parser parser;
   auto r = parser.Parse(in);
   ASSERT_FALSE(r.IsSuccess());
-  EXPECT_EQ("6: expected argument number", r.Error());
+  EXPECT_EQ("7: expected argument number", r.Error());
 }
 
 TEST_F(AmberScriptParserTest, OpenCLSetMissingAs) {
@@ -90,6 +94,7 @@ SHADER compute my_shader OPENCL-C
 #shader
 END
 PIPELINE compute my_pipeline
+  ATTACH my_shader
   SET KERNEL ARG_NAME a uint32 0
 END
 )";
@@ -97,7 +102,7 @@ END
   Parser parser;
   auto r = parser.Parse(in);
   ASSERT_FALSE(r.IsSuccess());
-  EXPECT_EQ("6: missing AS in SET command", r.Error());
+  EXPECT_EQ("7: missing AS in SET command", r.Error());
 }
 
 TEST_F(AmberScriptParserTest, OpenCLSetMissingDataType) {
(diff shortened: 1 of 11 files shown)


src/amberscript/parser_pipeline_set_test.cc
```diff
@@ -0,0 +1,216 @@
...
+
+#include "gtest/gtest.h"
+#include "src/amberscript/parser.h"
+
+namespace amber {
+namespace amberscript {
+
+using AmberScriptParserTest = testing::Test;
+
+TEST_F(AmberScriptParserTest, OpenCLSetMissingKernel) {
```
dj2: Should we test that SET is called on an OPENCL-C shader type?

src/executor.cc
```diff
@@ -59,10 +59,17 @@ Result Executor::Execute(Engine* engine,
     if (!r.IsSuccess())
       return r;
 
+    // OpenCL specific pipeline updates.
     for (auto& pipeline : script->GetPipelines()) {
       r = pipeline->UpdateOpenCLBufferBindings();
       if (!r.IsSuccess())
         return r;
+      r = pipeline->GenerateOpenCLPoDBuffers();
```
dj2: GenerateOpenCLPodBuffers? (PoD -> Pod?)

src/pipeline.cc
```diff
@@ -430,4 +430,122 @@ Result Pipeline::UpdateOpenCLBufferBindings() {
   return {};
 }
 
+Result Pipeline::GenerateOpenCLPoDBuffers() {
+  if (!IsCompute() || GetShaders().empty() ||
+      GetShaders()[0].GetShader()->GetFormat() != kShaderFormatOpenCLC)
```
dj2: nit: add {}s

src/pipeline.cc
```diff
@@ -430,4 +430,122 @@ Result Pipeline::UpdateOpenCLBufferBindings() {
...
+        Pipeline::ShaderInfo::DescriptorMapEntry::Kind::POD;
+    for (const auto& entry : iter->second) {
+      if (entry.kind != Pipeline::ShaderInfo::DescriptorMapEntry::Kind::POD &&
+          entry.kind !=
+              Pipeline::ShaderInfo::DescriptorMapEntry::Kind::POD_UBO) {
+        continue;
+      }
+
+      // Found the right entry.
+      if (entry.arg_name == arg_info.name ||
```
dj2: uses_name && arg_name == ?

src/pipeline.cc
```diff
@@ -430,4 +430,122 @@ Result Pipeline::UpdateOpenCLBufferBindings() {
...
+        std::make_pair(descriptor_set, binding));
+    Buffer* buffer = nullptr;
+    if (buf_iter == opencl_pod_buffer_map_.end() ||
+        buf_iter->first.first != descriptor_set ||
+        buf_iter->first.second != binding) {
+      // Ensure no buffer was previously bound for this descriptor set and
+      // binding pair.
+      for (const auto& buf_info : GetBuffers()) {
+        if (buf_info.descriptor_set == descriptor_set &&
+            buf_info.binding == binding)
```
dj2: nit: add {}'s

src/pipeline.h
```diff
@@ -226,6 +226,24 @@ class Pipeline {
   /// Generates a default depth attachment in D32_SFLOAT_S8_UINT format.
   std::unique_ptr<Buffer> GenerateDefaultDepthAttachmentBuffer() const;
 
+  /// Information on values set for OpenCL-C plain-old-data args.
+  struct ArgSetInfo {
+    std::string name = "";
```
dj2: I think "" is the default for std::string anyway?

src/pipeline.h
```diff
@@ -244,6 +262,10 @@ class Pipeline {
 
   uint32_t fb_width_ = 250;
   uint32_t fb_height_ = 250;
+
+  std::vector<ArgSetInfo> set_arg_values_;
+  std::vector<std::unique_ptr<Buffer>> opencl_pod_buffers_;
+  std::map<std::pair<uint32_t, uint32_t>, Buffer*> opencl_pod_buffer_map_;
```
dj2: Add a  comment what the values in the pair are?

src/amberscript/parser_pipeline_set_test.cc
alan-baker: The way the buffers are generated, the information in the SET is ignored if the shader is not an OPENCL-C shader. Would you prefer an error to that behaviour?

dj2: My concern is that you either set the shader type wrong, or are trying to use set for something it isn't designed for at this point. So, even though it will do nothing, I think it would be useful to generate an error to say something along the lines of "SET must be used with OPENCL shaders`

src/pipeline.h
alan-baker: Just wanted to be specific I guess.

dneto0: Yes, members with default constructors are default-constructed.

src/pipeline.cc
```diff
@@ -430,4 +430,122 @@ Result Pipeline::UpdateOpenCLBufferBindings() {
...
+
+      // Add a new buffer for this descriptor set and binding.
+      opencl_pod_buffers_.push_back(MakeUnique<Buffer>());
+      buffer = opencl_pod_buffers_.back().get();
+      buffer->SetBufferType(
+          kind == Pipeline::ShaderInfo::DescriptorMapEntry::Kind::POD
+              ? BufferType::kStorage
+              : BufferType::kUniform);
+      DatumType char_type;
+      char_type.SetType(DataType::kUint8);
```
dneto0: Is the point here that it doesn't matter what we set here, because it gets overwritten later with whatever format the data is?

src/amberscript/parser_pipeline_set_test.cc
alan-baker: Ok, I'll add an error.

src/pipeline_test.cc
```diff
@@ -480,6 +479,169 @@ TEST_F(PipelineTest, ClspvUpdateBindingTypeMismatch) {
   ASSERT_FALSE(r.IsSuccess());
   EXPECT_EQ("Buffer buf2 must be an uniform binding", r.Error());
 }
-#endif  // AMBER_ENABLE_CLSPV
+
+TEST_F(PipelineTest, OpenCLGeneratePoDBuffers) {
```
dneto0: "Pod" here to match

src/pipeline.cc
alan-baker: It is also the most convenient way to calculate the resizing versus the data in the descriptor map. Everything in the descriptor map is byte-based so this also simpler code (e.g the called to `ResizeTo`).

PR #592, opened 2019-07-23
Add support for specifying compile options

### LINKED ISSUE #433: Figure out compiler options for CLSPV
For SPIRV we have a SHADER_OPTIMIZATION section in the PIPELINE. We should investigate something similar to pass options to CLSPV.

alan-baker:
Fixes #433

* Adds COMPILE_OPTIONS command
  * currently only supported for OPENCL-C shaders
  * plumbed to Clspv
* Added tests

### DIFF
--- a/docs/amber_script.md
+++ b/docs/amber_script.md
@@ -214,8 +214,8 @@ The following commands are all specified within the `PIPELINE` command.
 ```
 
 ```groovy
-  # Set the compile options used to compile the given shader. Currently, only
-  # support for OPENCL-C shaders.
+  # Set the compile options used to compile the given shader. Options are parsed
+  # the same as on the command line. Currently, only supported for OPENCL-C shaders.
   COMPILE_OPTIONS {shader_name}
     {option}+
   END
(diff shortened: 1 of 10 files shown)


src/pipeline.cc
```diff
@@ -108,6 +108,23 @@ Result Pipeline::SetShaderOptimizations(const Shader* shader,
...
+
+  for (auto& info : shaders_) {
+    const auto* is = info.GetShader();
+    if (is == shader) {
+      info.SetCompileOptions(opts);
+      return {};
+    }
+  }
+
+  return Result("unknown shader specified for optimizations: " +
```
dj2: nit: s/optimizations/compile options/

docs/amber_script.md
```diff
@@ -213,6 +213,14 @@ The following commands are all specified within the `PIPELINE` command.
...
 ```
 
+```groovy
+  # Set the compile options used to compile the given shader. Currently, only
+  # support for OPENCL-C shaders.
+  COMPILE_OPTIONS {shader_name}
+    {option}+
+  END
+```
+
```
dj2: Can you add an example here of what the compile options for clspv would look like? How do we handle options with arguments?

alan-baker: I add an OpenCL example in the examples section. It includes an option with an argument. Also added a note that arguments are parsed the same as on the command line.

PR #594, opened 2019-07-23
[Dawn] Add multiple color-attachment support

### LINKED ISSUE #547: Dawn: Graphics support does not work for multiple color attachments
This test is failing: 
amber/tests/cases/draw_rect_mutliple_color_attachment.amber

sarahM0:
fixs #547 and simplifies DoClear by using helper functions


src/dawn/engine_dawn.cc
```diff
@@ -811,6 +797,7 @@ Result DawnPipelineHelper::CreateRenderPipelineDescriptor(
 
   renderPipelineDescriptor.primitiveTopology =
       ::dawn::PrimitiveTopology::TriangleList;
+  // TODO(sarahM0): Figure out what sampleCount is, it's been used a few times
```
dneto0: That's for multi-sample rendering:  multiple invocations of the fragment shader (e.g. 4 or 8 or more) are processed with slightly different positions within the final (square) pixel, and the colour results can be blended together. This is used for multi-sample anti-aliasing (sometime MSAA).
See https://www.khronos.org/registry/vulkan/specs/1.1/html/vkspec.html#primsrast-multisampling 
especially the multi-sample locations within a square pixel, mapped out in Table 23 in that edition of the Vulkan spec.

src/dawn/engine_dawn.cc
```diff
@@ -715,61 +726,36 @@ Result EngineDawn::DoClearDepth(const ClearDepthCommand* command) {
...
-    depth_stencil_attachment.stencilStoreOp = ::dawn::StoreOp::Store;
-    depth_stencil_descriptor = &depth_stencil_attachment;
-  }
 
-  // Attach the depth/stencil and colour attachments to the render pass.
-  ::dawn::RenderPassDescriptor rpd;
-  rpd.colorAttachmentCount = 1;
-  rpd.colorAttachments = color_attachment_descriptor;
-  rpd.depthStencilAttachment = depth_stencil_descriptor;
+  DawnPipelineHelper helper;
```
dneto0: This whole routine is a lot simpler now!  :-)

src/dawn/engine_dawn.cc
```diff
@@ -621,6 +568,70 @@ Result EngineDawn::Initialize(EngineConfig* config,
...
+                                &copySize);
+    auto commands = encoder.Finish();
+    auto queue = device.CreateQueue();
+    queue.Submit(1, &commands);
+
+    MapResult mapped_device_texture = MapBuffer(device, copy_buffer);
+    if (!mapped_device_texture.result.IsSuccess())
+      return mapped_device_texture.result;
+
+    if (i < render_pipeline.pipeline->GetColorAttachments().size()) {
```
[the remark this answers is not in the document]
sarahM0: Fair point!

sarahM0:
../amber/tests/cases/draw_rect_multiple_color_attachment.amber
is the test that passes now (I'd forgotten  to add run_tests.py).

tests/cases/draw_rect_multiple_color_attachment.amber
```diff
@@ -36,5 +36,5 @@ PIPELINE graphics my_pipeline
 END
 
 RUN my_pipeline DRAW_RECT POS 0 0 SIZE 800 600
-EXPECT framebuffer IDX 0 0 SIZE 800 600 EQ_RGBA 0 128 0 204
+EXPECT framebuffer IDX 0 0 SIZE 800 600 EQ_RGBA 0 127 0 204
```
dj2: Hm, I think this will fail on a vulkan host now (as it worked previously). Is this just a numerical difference between the dawn and vulkan engines?

sarahM0: I see your point however right now Vulkan fails with 128 and passes with 127, at least on my machine. Is there any reasons for this to behave differently on different machines?

dj2: Maybe? But, if both Vulkan and Dawn engines pass with this on your machine then I'm good with it. Maybe add a comment that we flipped the 128 to a 127 and if folks see this failing we need to find a better solution than changing the number? Just so we don't end up constantly flipping it back and forth, heh.

sarahM0: How about changing 0.5 to 0.499 so eliminate the confusion?

dj2: If that works, go for it.

dneto0: Yeah, I suspect something funny with quantization of the very-low-precision floating point used in rendering.  I could see it flipping back and forth. 

All the other AmberScript scripts in the tests/cases directory use 127, _except_ for the clear_color.amber script, but it's using an explicit 128 in that component of the clear color and it's using a UNORM so that's right in the middle of the range.

In short, I'm ok with this change. But let's keep an eye on this as an issue coming up.

PR #595, opened 2019-07-24
Add support for "NaN" as a double value

rg3igalia:
This PR adds support for "NaN", "nan" and other lowercase/uppercase combinations to be interpreted as a double value token and read using strtod in the tokenizer.

This is interesting for an upcoming VK-GL-CTS CL I'm preparing in which I test some SPIR-V instructions like OpFOrdEqual or OpFUnordEqual. To make the new tests interesting I'd like to be able to pass NaN to amber and make sure it's properly taken into account for ordered and unordered instructions.

### DIFF
--- a/src/tokenizer_test.cc
+++ b/src/tokenizer_test.cc
@@ -15,6 +15,7 @@
 #include "src/tokenizer.h"
 
 #include <limits>
+#include <cmath>
 
 #include "gtest/gtest.h"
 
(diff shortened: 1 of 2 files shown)


rg3igalia:
I signed it!

src/tokenizer_test.cc
```diff
@@ -77,6 +78,18 @@ TEST_F(TokenizerTest, ProcessDouble) {
   EXPECT_TRUE(next->IsEOS());
 }
 
+TEST_F(TokenizerTest, ProcessNaN) {
+  Tokenizer t("NaN");
```
dj2: Can you make this a parameterized test and try different combinations? NaN, nan, NAN, etc?

src/tokenizer_test.cc
```diff
@@ -77,6 +78,18 @@ TEST_F(TokenizerTest, ProcessDouble) {
...
+  auto next = t.NextToken();
+  ASSERT_TRUE(next != nullptr);
+  EXPECT_TRUE(next->IsDouble());
+  EXPECT_TRUE(std::isnan(next->AsDouble()) != 0);
```
dj2: std::isnan returns a bool, so just EXPECT_TRUE(std::isnan(next->AsDouble()))

src/executor_test.cc
```diff
@@ -202,7 +202,7 @@ class VkScriptExecutorTest : public testing::Test {
     auto engine = MakeUnique<EngineStub>();
     engine->Initialize(nullptr, nullptr, features, instance_extensions,
                        device_extensions);
-    return std::move(engine);
+    return engine;
```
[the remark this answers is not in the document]
rg3igalia: Yes, and moreover it appears some CI runs are failing after removing this, so I'm dropping the commit.

src/tokenizer_test.cc
rg3igalia: Will do.

src/tokenizer_test.cc
rg3igalia: Ah, you're right. isnan from C returns an integer but std::isnan returns a bool. Thanks!

Issue #600, opened 2019-07-26
AmberScript: Fuzzy buffer comparisions


dj2:
There are times when we know the results written into a buffer may be slightly different (framebuffers where there can be slight colour differences due to floating point instability).

We need to a way to do fuzzy comparisons against buffers in AmberScript. Aside from the syntax in AmberScript we need to determine how fuzzy comparison will work. Do image comparisons work the same as non-image comparisons? I can see a few different ways  (maybe we need to implement all of them?)

  1. Absolute value comparison. Given a set of values (I'd guess the format must match the buffer format so a vec3 requires 3 values, a float 1) the absolute value of the difference at each point must be less or equal to the given value.
  1. Percentage value comparison. As above but each value given is a % difference
  1. Total difference less then x. Sum all values for each buffer, absolute difference must be less then the given value.
  1. Fancy image comparison thingy using fancy image stuff ....
  1. Something else?

For the syntax, this could be an extension along the lines of `BUFFER_EQ`. Something like `BUFFER_FUZZY_EQ` if there is only one option picked or more specific things like `BUFFER_ABS_VALUE_COMPARE` or something.

dj2:
@dneto0 @afd @dnovillo

dj2:
Something more complicated like Root Mean Square Error? (https://www.statisticshowto.datasciencecentral.com/rmse/)

dnovillo:
RMSE has been giving decent results in our image comparison tests. Using a
threshold of .02 seems like a good compromise for us.

dj2:
The RMSE patch has landed. Going to close this. If we want other kinds of fuzzy comparisons we can open new bugs.

PR #602, opened 2019-07-27
Add root mean square error buffer comparison

dj2:
This CL adds a `RMSE_BUFFER` comparator which can be used with two
buffers to verify their values are within a given tolerance using the
root mean square error comparison method.

Issue #600


dj2:
Thoughts? This needs some tests around `Buffer::CalculateRMSE` and `Buffer::CalculateDiffs` but I think does what we want? It should also be extensible as we can add new comparators to the CompareBufferCommand as needed.

dj2:
@afd fyi

@dnovillo can you verify I did the RMSE calculation in `Buffer::CalculateRMSE` correctly?

hevrard:
@paulthomson can you have a look?
@dj2 Paul has been working on this in GraphicsFuzz, and he has more bandwidth than me for this now.

src/buffer.cc
```diff
@@ -53,6 +54,41 @@ T* ValuesAs(uint8_t* values) {
...
+    return Sub<uint8_t>(buf1, buf2);
+  if (comp.IsUint16())
+    return Sub<uint16_t>(buf1, buf2);
+  if (comp.IsUint32())
+    return Sub<uint32_t>(buf1, buf2);
+  if (comp.IsUint64())
+    return Sub<uint64_t>(buf1, buf2);
+  // TOOD(dsinclair): Handle float16 ...
+  if (comp.IsFloat16())
+    return 0.f;
```
paulthomson: Should we assert false here? Or was this supposed to be implemented before merging?

dj2: Should probably assert. I don't understand how float16 works at the moment so I'm skipping it for now. We have to verify all the float16 code in amber in general anyway.

dj2: done

docs/amber_script.md
```diff
@@ -446,6 +447,10 @@ EXPECT {buffer_name} IDX _x_in_pixels_ _y_in_pixels_ \
 
 # Checks that |buffer_1| contents are equal to those of |buffer_2|
 EXPECT {buffer_1} EQ_BUFFER {buffer_2}
+
+# Checks that the Root Mean Square Error when comparing |buffer_1| to
+# |buffer_2| is less than or equal too |tolerance|
```
dneto0: The syntax looks ok to me.  I'd clarify that |tolerance| is unitless.
E.g. when one buffer is an unbiased estimater of the other, then RMSE is the standard deviation, which is unitless.

src/buffer.cc
```diff
@@ -111,6 +150,58 @@ Result Buffer::IsEqual(Buffer* buffer) const {
...
+    return Result{"Buffers have a different size"};
+  if (buffer->width_ != width_)
+    return Result{"Buffers have a different width"};
+  if (buffer->height_ != height_)
+    return Result{"Buffers have a different height"};
+  if (buffer->ValueCount() != ValueCount())
+    return Result{"Buffers have a different number of values"};
+
+  auto diffs = CalculateDiffs(buffer);
+  float sum = 0.0;
```
dneto0: I'd do this in double.
Sums of squares can get pretty big.

We _could_ go even further and normalize.:   E.g. divide  each diff by the _mean_ of the diffs, and then compute RMSE, then remultiply by the absolute value of the mean.
But that's probably excessive for now.

Issue #19, opened 2018-11-14
Dawn: Implement DoEntryPoint


dj2:
Implement the DoEntryPoint command in the Dawn engine.

dneto0:
Fixed with https://github.com/google/amber/pull/605

PR #605, opened 2019-07-29
[Dawn] DoPatchParameterVertices and DoEntryPoint are not supported in Dawn

### LINKED ISSUE #24: Dawn: Implement DoPatchParameterVertices
Implement DoPatchParameterVertices in the dawn engine.

sarahM0:
fix #24  #19

### DIFF
--- a/src/dawn/engine_dawn.cc
+++ b/src/dawn/engine_dawn.cc
@@ -1169,12 +1169,12 @@ Result EngineDawn::DoCompute(const ComputeCommand*) {
 }
 
 Result EngineDawn::DoEntryPoint(const EntryPointCommand*) {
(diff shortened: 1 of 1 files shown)


src/dawn/engine_dawn.cc
```diff
@@ -1169,12 +1169,12 @@ Result EngineDawn::DoCompute(const ComputeCommand*) {
 }
 
 Result EngineDawn::DoEntryPoint(const EntryPointCommand*) {
-  return Result("Dawn:DoEntryPoint not implemented");
+  return Result("Dawn: Entry point must be \"main\" in Dawn");
```
dneto0: I think this is true now but will be relaxed in the future.
I'll file an issue to enhance this later.

PR #606, opened 2019-07-30
Enable shader optimizations

alan-baker:
* During shader compilation, if SPIRV-Tools are enabled run requested
optimizations for the shader
  * these optimizations are pipeline specific
  * added a test

### DIFF
--- a/src/shader_compiler.h
+++ b/src/shader_compiler.h
@@ -34,9 +34,13 @@ class ShaderCompiler {
   ~ShaderCompiler();
 
   /// Returns a result code and a compilation of the given shader.
-  /// If the |shader| has a corresponding entry in the |shader_map|, then the
-  /// compilation result is copied from that entry. Otherwise a compiler is
-  /// invoked to produce the compilation result.
+  /// If the shader in |shader_info| has a corresponding entry in the
+  /// |shader_map|, then the compilation result is copied from that entry.
+  /// Otherwise a compiler is invoked to produce the compilation result.
+  ///
+  /// If |shader_info| specifies shader optimizations to run and there is no
+  /// entry in |shader_map| for that shader, then the SPIRV-Tools optimizer will
+  /// be invoked to produce the shader binary.
   std::pair<Result, std::vector<uint32_t>> Compile(
       Pipeline::ShaderInfo* shader_info,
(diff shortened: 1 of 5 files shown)


src/shader_compiler_test.cc
```diff
@@ -165,6 +165,60 @@ TEST_F(ShaderCompilerTest, InvalidHex) {
...
+)";
+
+  Shader shader(kShaderTypeCompute);
+  shader.SetName("TestShader");
+  shader.SetFormat(kShaderFormatSpirvAsm);
+  shader.SetData(spirv);
+
+  Pipeline::ShaderInfo unoptimized(&shader, kShaderTypeCompute);
+  Pipeline::ShaderInfo optimized(&shader, kShaderTypeCompute);
+  optimized.SetShaderOptimizations({"--eliminate-dead-code-aggressive"});
```
dneto0: This conflicts with the example in https://github.com/google/amber/blob/master/docs/amber_script.md  
There, the option is named without the leading hyphens.  Please update the example in the doc.

src/shader_compiler.h
```diff
@@ -34,9 +34,13 @@ class ShaderCompiler {
...
 
   /// Returns a result code and a compilation of the given shader.
-  /// If the |shader| has a corresponding entry in the |shader_map|, then the
-  /// compilation result is copied from that entry. Otherwise a compiler is
-  /// invoked to produce the compilation result.
+  /// If the shader in |shader_info| has a corresponding entry in the
+  /// |shader_map|, then the compilation result is copied from that entry.
+  /// Otherwise a compiler is invoked to produce the compilation result.
+  ///
+  /// If |shader_info| specifies shader optimizations to run and there is no
```
dneto0: This is unclear what those optimization strings should look like. ...

Please add a comment to src/pipeline.h to the SetShaderOptimizations method to say what that strings should look like. I'd reference the command line options of spirv-opt, or   SPIRV-Tools include/spirv-tools/optimizer.hpp method Optimizer::RegisterPassesFromFlags

dj2:
Thanks a lot for adding this. I'd thought I'd file a bug that this was missing but see I forgot ... sorry about that.

PR #609, opened 2019-07-31
[Dawn] Resolve remaining test failures

### LINKED ISSUE #608: Dawn: resolve remaining test failures
See https://github.com/google/amber/blob/master/tests/run_tests.py#L102

My first thoughts:

- specialization: I think this is planned for WebGPU but not yet designed.  I suspect Dawn does not support it.

- non-default entry-point:  current limitation of Dawn.  See #607 

- OpenCL support.  That uses Clspv.  _Generally_ Clspv requires the VariablePointers or VariablePointersStorageBuffer SPIR-V feature, and the corresponding Vulkan feature.  But in many cases the compute shader (kernel) code is simple enough that it's not needed. We should look at those examples to see if the compiler generates code that should be usable in Dawn/WebGPU.

sarahM0:
issue #608


tests/cases/clear_with_depth.amber
```diff
@@ -40,7 +40,7 @@ END
 # We would need a second pipeline or a depth test on the pipeline to observe the effects of the
 # depth-stencil clear.
 
-CLEAR_COLOR my_pipeline 10 20 30 40
+CLEAR_COLOR my_pipeline 255 0 0 255
```
dj2: Why did this need to change?

tests/cases/draw_triangle_list_with_depth.vkscript
```diff
@@ -43,44 +43,40 @@ void main() {
...
 
 #         Mixed color for quarter frame
-          0    0  0  127    127 127 127   0
-        127  127  0  127    127 127 127   0
-          0  127  0  127    127 127 127   0
-
-          0    0  0  127    127 127 127   0
-        127  127  0  127    127 127 127   0
-        127    0  0  127    127 127 127   0
+          0    0  16  127    127 127 127   0
```
dj2: Ditto, why bump up the blues?

tests/run_tests.py
```diff
@@ -89,26 +88,28 @@
   "probe_no_compute_with_ssbo.vkscript",
   # Max number of descriptor sets is 4 in Dawn
   "multiple_ssbo_with_sparse_descriptor_set_in_compute_pipeline.vkscript",
-  # DoEntryPoint is not supported in Dawn backend, yet
+  # Dawn entry point has to be "main" as a result it does not support
+  # doEntryPoint or opencl (in opencl using "main" as entry point is invalid).
+  # issue #607 (temp dawn limitation)
```
dj2: Awesome, thanks for linking these back to bugs.

tests/cases/draw_triangle_list_with_depth.vkscript
sarahM0: this was part an attempt to figure out why this test does not work in dawn. I'll revert it.

dneto0: The blues here are being used as Z values.
We wanted to make sure there was no flakiness with precision of the depth values.  So we separated them by wide intervals.   This might not have been strictly necessary, we did this while trying to debug the depth flakiness in the Dawn flow.

tests/cases/clear_with_depth.amber
sarahM0: 10 20 30 40 is not a color that we could verify by looking at it.

dneto0: True, this probably didn't have to change.

dneto0: ah.  :-)

PR #610, opened 2019-07-31
Avoid uninitialized-var warning

dneto0:
The warning was on dawn_buffer.  
Initialized it with nullptr and add an early return with message in case it was in fact not initialized

### DIFF
--- a/src/dawn/engine_dawn.cc
+++ b/src/dawn/engine_dawn.cc
@@ -1260,26 +1260,25 @@ Result EngineDawn::DoPatchParameterVertices(
 Result EngineDawn::DoBuffer(const BufferCommand* command) {
   Result result;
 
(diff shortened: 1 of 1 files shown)


src/dawn/engine_dawn.cc
```diff
@@ -1260,26 +1260,25 @@ Result EngineDawn::DoPatchParameterVertices(
 Result EngineDawn::DoBuffer(const BufferCommand* command) {
   Result result;
 
-  ::dawn::Buffer* dawn_buffer;
+  ::dawn::Buffer* dawn_buffer = nullptr;
+
+  const auto descriptor_set = command->GetDescriptorSet();
```
dneto0: I also did a drive-by simplification of the code around looking up the descriptorset/binding in the maps.

PR #613, opened 2019-08-01
DEPS: Downgrade glslang to match Dawn

dneto0:
When using Amber with Dawn, both Dawn and Amber pull in Glslang.
If the versions of Glslang differ in ABIs to which Shaderc is
sensitive, then we can't use Dawn with Amber.  The symptom is
that Amber won't link.

In this case:
Glslang commit 137d2ba changed the signature of TShader::mapIO,
and that was merged into Glslang master with f04f1f93 (after
an earlier commit and revert pair).

So downgrade the glslang commit to the one currently supported
in Dawn.

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -13,7 +13,7 @@ vars = {
   'clspv_revision': '64a51387967ea2a26e4a52a22ad8640148194edc',
   'cpplint_revision': '9f41862c0efa7681e2147910d39629c73a2b2702',
   'dxc_revision': 'ec912b2ec95feb50925704dd631ef7abee1a5f09',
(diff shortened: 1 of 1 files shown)


dneto0:
FYI.  This consideration will affect anybody rolling forward the Glslang commit in Amber.
Cc: @sarahM0 @dj2

dneto0:
With this, I can build (especially link) an Amber with a current Dawn.

dj2:
Will the debug-dawn bot catch this for us going forward? Or, is there something we can do to make sure this doesn't happen in the future?

dneto0:
I think any bot which tests Amber-with-Dawn would find this issue.

PR #614, opened 2019-08-01
Add test case for --graphics-robust-access transform

dneto0:


dneto0:
cc: @Kangz

dneto0:
I tested this via Vulkan and Dawn backends.

Kangz:
Nice!

PR #615, opened 2019-08-02
Add option to disable SPIR-V validation

paulthomson:
This is useful for GraphicsFuzz as we already validate our shaders, often with a newer version of the validator (and we don't want a validator bug to stop us from running the shader). We can't just build without SPIRV-Tools because we still want the assembler.

I am not sure how you feel about passing a bool down through so many functions, and the use of default arguments. I could have used a field (member variable) instead of a parameter perhaps. Or I could pass the `Options` down, but the existing code seems to stop passing `Options` down at a certain point, so I followed that.

### DIFF
--- a/include/amber/amber.h
+++ b/include/amber/amber.h
@@ -105,6 +105,9 @@ struct Options {
   bool pipeline_create_only;
   /// Delegate implementation
   Delegate* delegate;
+  /// If true, disables SPIR-V validation. If false, SPIR-V shaders will be
+  /// validated using the Validator component (spirv-val) from SPIRV-Tools.
+  bool disable_spirv_validation;
 };
 
 /// Main interface to the Amber environment.
--- a/include/amber/amber.h
+++ b/include/amber/amber.h
@@ -38,6 +38,13 @@ enum EngineType {
   kEngineTypeDawn,
 };
 
+enum class ExecutionType {
+  /// Execute as normal.
+  kExecute = 0,
+  /// Only create the pipelines and then exit.
+  kPipelineCreateOnly
+};
+
 /// Override point of engines to add their own configuration.
 struct EngineConfig {
   virtual ~EngineConfig();
@@ -101,13 +108,14 @@ struct Options {
   std::string spv_env;
   /// Lists the buffers to extract at the end of the execution
   std::vector<BufferInfo> extractions;
-  /// Terminate after creating the pipelines.
-  bool pipeline_create_only;
-  /// Delegate implementation
-  Delegate* delegate;
+  /// The type of execution. For example, execute as normal or just create the
+  /// piplines and exit.
+  ExecutionType execution_type;
   /// If true, disables SPIR-V validation. If false, SPIR-V shaders will be
   /// validated using the Validator component (spirv-val) from SPIRV-Tools.
   bool disable_spirv_validation;
+  /// Delegate implementation
+  Delegate* delegate;
 };
 
 /// Main interface to the Amber environment.
--- a/src/amber.cc
+++ b/src/amber.cc
@@ -68,7 +68,8 @@ Options::Options()
     : engine(amber::EngineType::kEngineTypeVulkan),
       config(nullptr),
       pipeline_create_only(false),
-      delegate(nullptr) {}
+      delegate(nullptr),
+      disable_spirv_validation(false) {}
 
 Options::~Options() = default;
 
@@ -171,7 +172,8 @@ amber::Result Amber::ExecuteWithShaderData(const amber::Recipe* recipe,
   Result executor_result = executor.Execute(
       engine.get(), script, opts->delegate, shader_data,
       opts->pipeline_create_only ? ExecutionType::kPipelineCreateOnly
-                                 : ExecutionType::kExecute);
+                                 : ExecutionType::kExecute,
+      opts->disable_spirv_validation);
   // Hold the executor result until the extractions are complete. This will let
   // us dump any buffers requested even on failure.
 
--- a/src/amber.cc
+++ b/src/amber.cc
@@ -67,9 +67,9 @@ EngineConfig::~EngineConfig() = default;
 Options::Options()
     : engine(amber::EngineType::kEngineTypeVulkan),
       config(nullptr),
-      pipeline_create_only(false),
-      delegate(nullptr),
-      disable_spirv_validation(false) {}
+      execution_type(ExecutionType::kExecute),
+      disable_spirv_validation(false),
+      delegate(nullptr) {}
 
 Options::~Options() = default;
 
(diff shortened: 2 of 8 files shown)


include/amber/amber.h
```diff
@@ -105,6 +105,9 @@ struct Options {
   bool pipeline_create_only;
   /// Delegate implementation
   Delegate* delegate;
+  /// If true, disables SPIR-V validation. If false, SPIR-V shaders will be
+  /// validated using the Validator component (spirv-val) from SPIRV-Tools.
+  bool disable_spirv_validation;
```
dj2: Move this up with pipeline_create_only.

src/amber.cc
```diff
@@ -171,7 +172,8 @@ amber::Result Amber::ExecuteWithShaderData(const amber::Recipe* recipe,
   Result executor_result = executor.Execute(
       engine.get(), script, opts->delegate, shader_data,
       opts->pipeline_create_only ? ExecutionType::kPipelineCreateOnly
```
dj2: I think we should just pass options into `Execute` here as this is getting long ....

src/executor.h
```diff
@@ -39,10 +39,13 @@ class Executor {
                  const Script* script,
                  Delegate* delegate,
                  const ShaderMap& map,
-                 ExecutionType executionType);
+                 ExecutionType executionType,
+                 bool disable_spirv_validation = false);
```
dj2: I don't like default args, I find they hide when parameters are no longer used.

src/executor.cc
```diff
@@ -30,14 +30,16 @@ Executor::Executor() = default;
...
+                                bool disable_spirv_validation) {
   for (auto& pipeline : script->GetPipelines()) {
     for (auto& shader_info : pipeline->GetShaders()) {
       ShaderCompiler sc(script->GetSpvTargetEnv());
 
       Result r;
       std::vector<uint32_t> data;
-      std::tie(r, data) = sc.Compile(&shader_info, shader_map);
+      std::tie(r, data) =
+          sc.Compile(&shader_info, shader_map, disable_spirv_validation);
```
dj2: Let's pass disable_spirv_validation into the ShaderCompiler constructor instead since it will be global to the compiler

src/shader_compiler.cc
```diff
@@ -142,10 +143,13 @@ std::pair<Result, std::vector<uint32_t>> ShaderCompiler::Compile(
     return {Result("Invalid shader format"), results};
   }
 
+  (void)disable_spirv_validation;
```
dj2: The linter won't like this. Is there some way we can restructure the code so we don't need have a fake access?

src/shader_compiler.h
```diff
@@ -43,7 +43,8 @@ class ShaderCompiler {
   /// be invoked to produce the shader binary.
   std::pair<Result, std::vector<uint32_t>> Compile(
       Pipeline::ShaderInfo* shader_info,
-      const ShaderMap& shader_map) const;
+      const ShaderMap& shader_map,
+      bool disable_spirv_validation = false) const;
```
dj2: Remove default value please.

src/amber.cc
paulthomson: I agree. I have done this. However, `Options` includes:

```c++
EngineType engine;
std::string spv_env;
std::vector<BufferInfo> extractions;
```

none of which are used in `Execute`. I guess that is OK though, although it could be a bit confusing if future code ends up trying to rely on these fields.

I also need to update about 29 tests which currently do this:

```c++
Executor ex;
Result r = ex.Execute(engine.get(), script.get(), nullptr, ShaderMap(),
                        ExecutionType::kExecute);
```

to something like this:

```c++
Options options;
Executor ex;
Result r = ex.Execute(engine.get(), script.get(), ShaderMap(), &options);
```

`options->engine` will be `amber::EngineType::kEngineTypeVulkan` but the actual engine is `EngineStub`. But again, this works, as `Execute` does not access `options->engine`. Could be a bit confusing though.

src/shader_compiler.cc
paulthomson: The checks passed, but perhaps you mean other linters. I think this can be removed once it becomes a member variable.

include/amber/amber.h
paulthomson: Done

src/executor.h
paulthomson: Yes, makes sense. Removed.

src/executor.cc
paulthomson: Done.

src/shader_compiler.cc
paulthomson: Removed.

src/shader_compiler.h
paulthomson: Removed, as it is now passed in the constructor.

src/executor.cc
```diff
@@ -76,13 +77,14 @@ Result Executor::Execute(Engine* engine,
...
 
-  if (executionType == ExecutionType::kPipelineCreateOnly)
+  if (options->execution_type == ExecutionType::kPipelineCreateOnly)
     return {};
 
   // Process Commands
   for (const auto& cmd : script->GetCommands()) {
-    if (delegate && delegate->LogExecuteCalls())
-      delegate->Log(std::to_string(cmd->GetLine()) + ": " + cmd->ToString());
+    if (options->delegate && options->delegate->LogExecuteCalls())
```
dj2: nit: this should have {}'s as the body is more then 1 line.

src/shader_compiler.h
```diff
@@ -30,7 +30,8 @@ namespace amber {
 class ShaderCompiler {
  public:
   ShaderCompiler();
-  explicit ShaderCompiler(const std::string& env);
+  explicit ShaderCompiler(const std::string& env,
```
dj2: nit: explicit not needed now that there is more then 1 argument.

src/shader_compiler.h
```diff
@@ -53,6 +54,7 @@ class ShaderCompiler {
                         std::vector<uint32_t>* result) const;
 
   std::string spv_env_;
+  bool disable_spirv_validation_;
```
dj2: Default this to false here.

src/executor.cc
paulthomson: Fixed

src/shader_compiler.h
paulthomson: Fixed

src/shader_compiler.h
paulthomson: Oops. I have now set it to false in the default constructor. My IDE even warned me about this and I missed it.

dj2: This is fine for now, for future reference you can do:

```
bool disabled_spirv_validation_ = false;
```
 in the header file which will get set as the default in all constructors (it can still be overridden)

paulthomson:
Ah well I may as well change it to that! Thanks!

PR #625, opened 2019-08-30
CTS unused variable error fix

jarisiru:
When trying to update amber within VK-GL-CTS, the sanity script fails due to unused private variable when building with clang. These changes remove that variable.

### DIFF
--- a/src/executor.cc
+++ b/src/executor.cc
@@ -34,8 +34,7 @@ Result Executor::CompileShaders(const amber::Script* script,
                                 Options* options) {
   for (auto& pipeline : script->GetPipelines()) {
     for (auto& shader_info : pipeline->GetShaders()) {
-      ShaderCompiler sc(script->GetSpvTargetEnv(),
-                        options->disable_spirv_validation);
+      ShaderCompiler sc(script->GetSpvTargetEnv());
 
       Result r;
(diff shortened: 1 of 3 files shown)


jarisiru:
@googlebot I signed it!

dj2:
This will fix the CTS which compiles with spirv-tools disabled, but will break the build where it is enabled.  Either the inclusion of the variable needs to be #ifdef'd, or we move the usage out to wrap around the #define use of spirv-tools later.

@dneto0, I'd lean towards the second option, do you have opinions?

jarisiru:
@dj2 I'm using CTS with spirv-tools enabled, and it comples (with visual studio)..

However, I was too hasty with the pull request, as there seem to be still other issues with the sanity script, at least one unused parameter.

dj2:
The CTS has spirv-tools in it, but the CTS build disables the spirv-tools support inside Amber itself.

dneto0:
> @dneto0, I'd lean towards the second option, do you have opinions? 

I prefer the second option as well:  Use the variable in all compilation scenarios.

dneto0:
I posted https://github.com/google/amber/pull/626 to implement option 2

dneto0:
Closing this PR, as it should be fixed by #626

PR #626, opened 2019-09-04
Avoid unused var warning when SPIRV-Tools not used

dneto0:

PR #627, opened 2019-09-04
Disable warning in shader_compiler.cc.

ben-clayton:
This warning is treated as an error in dEQP.


ben-clayton:
Just seen that this is a dupe of https://github.com/google/amber/pull/626.

PR #634, opened 2019-09-09
Build fixes for updating dEQP's version of amber

ben-clayton:
Successful build can be seen here: https://gerrit.khronos.org/#/c/4546/

Required moving clang-only warning flags out of the `if(${COMPILER_IS_LIKE_GNU})` block and into the `if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")` block. See:
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
https://clang.llvm.org/docs/DiagnosticsReference.html
for lists of accepted GCC and Clang warning flags.

This PR also include's @dneto0 's implicit cast fix, but I've used `static_cast` instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,18 +194,20 @@ function(amber_default_compile_options TARGET)
       -Wall
       -Werror
       -Wextra
-      -Wno-c++98-compat
-      -Wno-c++98-compat-pedantic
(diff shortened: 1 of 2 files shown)


ben-clayton:
@dneto0, @dj2 - please take a look.

Issue #618, opened 2019-08-19
Support dumping multiple framebuffers


asuonpaa:
Currently a command option -I is used for selecting which frame buffer gets dumped. Would be nice to be able to dump all framebuffers used in an amber file. Maybe just add an option that outputs framebuffername.png for each framebuffer used?

I could do a PR if people first agree how this should be implemented.

dj2:
```
  -i <filename>             -- Write rendering to <filename> as a PNG image if it ends with '.png',
                               or as a PPM image otherwise.
  -I <buffername>           -- Name of framebuffer to dump. Defaults to 'framebuffer'.
  -b <filename>             -- Write contents of a UBO or SSBO to <filename>.
  -B [<desc set>:]<binding> -- Descriptor set and binding of buffer to write.
                               Default is [0:]0.
```

Looking at our current flags and how the buffer dumping works. I think we should change the framebuffer name variable into a `vector`. For buffers we append to a single file, but for framebuffers that won't work. So, I'd propose we write to a filename based off the `-i` flag where we do something like (if -i out_file.png is provided) `out_file-fb_name.png` where fb_name is the framebuffer name. This would only happen if multiple framebuffers are requested. If a single buffer is requested just `out_file.png` would be used.

We could then add a flag like `-A` for all framebuffers which creates the full list for you if that's needed.

paulthomson:
Passing thought: what about some way of listing each buffer and its output file one-after-the-other, so that the behavior is very clear and explicit:

`buffer_1 output1.png buffer_2 output2.ppm buffer_3 output3.bin buffer_4 output4.bin`

With VkScript, I assume there will be (implicitly) `framebuffer` and maybe `0:0`, `0:1`, etc., but in AmberScript, all the buffers can be explicitly named. If the output files end with .png or .ppm, a conversion is done, but otherwise, the raw binary is dumped (although it might need to be error if the buffer format doesn't make sense for the output extension). This also somewhat unifies the dumping of framebuffers vs. buffers. Of course, this doesn't fit particularly well with the existing flags.

dj2:
We could do some thing like pair up -i and -I flags?

`-i first.png -I framebuffer1 -i second.ppm -I otherbuffer -i third.png -I some_other_buffer`

then they just match, so framebuffer1 goes to first.png and so on? If you have one more -i then -I it defaults to `framebuffer`. If you have more then 1 extra it's an error?

[...]

PR #636, opened 2019-09-11
Add support for dumping multiple frame buffers.

asuonpaa:
Added support for specifying multiple `-i` and `-I` command line parameters to allow dumping multiple frame buffers.

Fixes https://github.com/google/amber/issues/618


samples/amber.cc
```diff
@@ -431,9 +432,13 @@ int main(int argc, const char** argv) {
                                      options.buffer_to_dump.end());
   }
 
-  if (!options.image_filename.empty()) {
+  // Use default frame buffer name when not specified.
+  while (options.image_filenames.size() > options.fb_names.size())
```
dj2: Should we check if this is going to be more then 1 and, if so, log an error an exit? Folks may not expect to get the same framebuffer written out n times.

samples/amber.cc
```diff
@@ -451,14 +456,16 @@ int main(int argc, const char** argv) {
       // give clues as to the failure.
     }
 
-    if (!options.image_filename.empty()) {
+    auto fb_name = options.fb_names.begin();
+
+    for (const auto& image_filename : options.image_filenames) {
```
dj2: I think I'd prefer turning this into:

```
for (size_t i = 0; i < options.image_filenames.size(); ++i)
```

and then removign fb_name above and using options.fb_names[i] below.

samples/amber.cc
```diff
@@ -498,7 +505,10 @@ int main(int argc, const char** argv) {
         std::cerr << options.buffer_filename << std::endl;
       } else {
         for (const amber::BufferInfo& buffer_info : amber_options.extractions) {
-          if (buffer_info.buffer_name == options.fb_name)
+          // Skip frame buffers.
+          if (std::any_of(
+                  options.fb_names.begin(), options.fb_names.end(),
+                  [&](std::string s) { return s == buffer_info.buffer_name; }))
```
dj2: This should also make sure to check the default generated framebuffer name as it won't be in the fb_names if there were no image filenames specified.

samples/amber.cc
```diff
@@ -498,7 +505,10 @@ int main(int argc, const char** argv) {
...
             continue;
```
dj2: nit: Add {}'s around the if() body since the declaration spans multiple lines.

PR #657, opened 2019-09-18
[AmberScript] Fix parsing of matrix data into buffers.

### LINKED ISSUE #616: Possible problem in the way Amber initializes a matrix uniform buffer
[The attached AmberScript file](https://github.com/google/amber/files/3474532/tmp_shader_test.amber.txt) (rename .amber.txt to .amber) captures a simple shader that declares a mat2x2 uniform m, populates it with 1.0 in every entry, and then writes:

```
vec4(m[1][0], 0.0, 0.0, 1.0)
```

to the output color.

This should give a red image.

On both SwiftShader and Intel Mesa (Vulkan) I get a black image.

The SPIR-V code that glslang has produced and that spirv-opt has optimized looks correct.

So I wonder, whether there might be an issue either in the way Amber is initializing this buffer, or perhaps in the way I've expressed the desired initialization in AmberScript:

```
BUFFER m DATA_TYPE mat2x2<float> DATA
 1.0 1.0 1.0 1.0
END
```

dj2:
Currently parsing of matrix data in a buffer DATA block would trigger
memory corrpution issues. This was due to several things:

 * Incorrect interpretation of mat2x4 for rows and columns
 * Not clamping the amount of data read from the DATA block
 * Handling of padding values.

With this CL we correctly handle mat2x4 as 2 columns and 4 rows. We do
not require padding values in the DATA block for matrices, it will be
inserted automatically as needed. We error out if the number of DATA
elements is not a multiple of the amount of input needed for the number
of elements.

Fixes #616


dj2:
@afd with this change, your good_fix_initialization.amber.txt will error out and say there is too much data. If you remove the two padding 0.0's you added it should then pass (it does here on a Nvidia device).

dj2:
@afd do you have any objections to me adding your good_fix_initialization test to the amber repo as test case?

afd:
@dj2 Please do feel free to add that as a test case.

afd:
[examples.zip](https://github.com/google/amber/files/3631629/examples.zip)

Based on some experiments I don't think the fix works for all matrix dimensions.

The attached archive has two amber files - matrices_simple_prepared.amber and matrices_uniforms_prepared.amber.  They are both generated from corresponding .frag files.  The .json files indicate the values with which uniforms will be set (this is GraphicsFuzz's input format for uniforms).

The shaders consider all 9 matrix types - from 2x2 through 4x4.  For each matrix type they sum the entries of a given matrix that is populated with all 1s - so for 4x4 the sum will be 16.  The result is then divided by 16 and used to colour in greyscale on section of a 3x3 grid, so that the top left ends up being near-black (4.0 / 16.0 from the 2x2 matrix), and the bottom right ends up being white (16.0 / 16.0 from the 4x4 matrix).

This is the image I expect to see, and it is what I get from the "simple" case:

image_0

For the "uniforms" case I should get the same image but I get this image:

image_0

If this is a pain to debug from these large shaders I can try to trim down to something smaller.

(And apologies in advance if it turns out to be a bug in the shaders, but I did cross-check them on GLES.)

dj2:
Thanks for the example. Looks like the problem is, I don't write the data in std140 for the uniform. I'm going to need to update AmberScript slightly to deal with that as we need a way on the buffer to say STD140. We don't know we need it in 140 until we get to the uniform otherwise, which is too late as the data is already in the buffer.

So, does landing this as is make sense since it fixes part of the issue and I'll start working on allowing layout modifiers on the buffers so we can make the new test case work?

afd:
@dj2 Yes, that makes sense.  Once I know what the changes to AmberScript are, we can update GraphicsFuzz to emit the appropriate code.

PR #661, opened 2019-09-23
Use Format for args instead of DatumType.

dj2:
This CL converts the arg holder to store a Format instead of a
DatumType.


dj2:
Dawn error looks unrelated.

PR #670, opened 2019-09-25
Use enum for available layouts.

dj2:
This CL removes the is_std_140 boolean from the Format class and uses an
enum instead.

### DIFF
--- a/src/format.cc
+++ b/src/format.cc
@@ -22,7 +22,7 @@ Format::Format() = default;
 
 Format::Format(const Format& b) {
   type_ = b.type_;
-  is_std140_ = b.is_std140_;
+  layout_ = b.layout_;
   pack_size_in_bytes_ = b.pack_size_in_bytes_;
   column_count_ = b.column_count_;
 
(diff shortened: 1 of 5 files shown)


src/format.cc
```diff
@@ -125,7 +125,7 @@ void Format::RebuildSegments() {
     // which rounds up to a vec4.
     //
     // In std140 and std430 a vector of size 3N will round up to a vector of 4N.
```
jaebaek: This is an independent issue, but I am a little bit confused. Based on this comment, it should check `if layout_ == Layout::kStd140 or layout_ == Layout::kStd430` ?

src/format.h
```diff
@@ -24,7 +24,7 @@
 
 namespace amber {
 
-/// The format class describes requested image formats. (eg. R8G8B8A8_UINT).
+/// The format class describes requested  data formats. (eg. R8G8B8A8_UINT).
```
jaebaek: If this is a format of buffer, we cannot use image formats e.g., `R8G8B8A8_UINT`. I think this example `(eg. R8G8B8A8_UINT)` can be interpreted as "we can use image formats for buffer resource". Please correct me if I misunderstand this.

src/format.cc
dj2: Maybe when we have more layouts. Right now only 140 and 430 exist.

src/format.h
dj2: You can define a BUFFER in amberscript with a FORMAT, you then use it as an image buffer.

PR #671, opened 2019-09-25
[AmberScript] Add STD140 and STD430 layout modifiers to BUFFER.

### LINKED ISSUE #659: [AmberScript] Can not create a std140 buffer.
When parsing buffer data, we set the data early in AmberScript. But, at the point we set the data we assume the buffer is in std430. We don't know it's std140 until we parse the `uniform` command on the buffer.

For an example see PR #657 

In order to fix this, we're going to add layout qualifiers to the `BUFFER` commands. So, you'll have `STD140`, `STD430`, `RELAXED`, `SCALAR`. If you don't add a qualifier the buffer will be in `STD430`.

dj2:
This CL allows specifying a STD140 or STD430 layout modifier for a
BUFFER in AmberScript. When using a buffer as a uniform specifying
STD140 will inject the correct padding.

Fixes #659


dj2:
@afd this fixes the matrices_uniform_draw test you'd provided. I added a STD140 qualifier onto the buffers in the example which is what makes it work.

Please let me know if there are other issues in this area that need to be looked at.

dj2:
The added .amber test case pads the Yx2 matrices, fails otherwise.

dj2:
I could possibly add unit tests for padding as well, can look tomorrow.

src/amberscript/parser_buffer_test.cc
```diff
@@ -323,6 +399,42 @@ END
...
+  std::vector<int32_t> results0 = {2, 3, 4, 5, 6, 7, 8, 9};
+  const auto* data0 = buffer->GetValues<int32_t>();
+  for (size_t i = 0; i < results0.size(); ++i) {
+    EXPECT_EQ(results0[i], data0[i]);
+  }
+}
+
+TEST_F(AmberScriptParserTest, BufferDataMultiRowStd140) {
+  std::string in = R"(
+BUFFER my_index_buffer DATA_TYPE vec2<int32> STD140 DATA
```
dj2: Now I feel like this test is wrong .... I think it should pad?

dj2: Sorry, this is right, it's the matrix that pads, will add a matrix example.

dj2:
K, I added a matrix test in 140 and 430 where the 140 has to pad and the 430 doesn't.

PR #674, opened 2019-09-26
Remove GetComponents from the Format class.

dj2:
With this change, it is no longer possible to get a list of components
from a Format. Instead, the segments should be used and the padding
taken into account.

A GetOnlyComponent was added for the usage in the amberscript parser but
requires there to only be a single component in the format.

### DIFF
--- a/src/format.h
+++ b/src/format.h
@@ -216,7 +216,7 @@ class Format {
   bool IsDouble() const { return AreAllComponents(FormatMode::kSFloat, 64); }
 
   /// Generates the image format name for this format if possible. Returns
-  /// name if generated or "" otherwise.
+  /// the name if generated or "" otherwise.
   std::string GenerateName() const;
 
  private:
--- a/src/format.h
+++ b/src/format.h
@@ -170,8 +170,8 @@ class Format {
            type_ == FormatType::kS8_UINT;
   }
 
-  /// Returns true if the format components are scaled
-  bool IsScaled() const {
+  /// Returns true if the format components are normalized.
+  bool IsNormalized() const {
     FormatMode mode = components_.back()->mode;
     return mode == FormatMode::kUNorm || mode == FormatMode::kSNorm ||
            mode == FormatMode::kSRGB;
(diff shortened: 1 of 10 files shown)


src/format.h
```diff
@@ -164,6 +170,13 @@ class Format {
            type_ == FormatType::kS8_UINT;
   }
 
+  /// Returns true if the format components are scaled
```
s-perron: Is there somewhere I can look up the definition of scaled?  My intuition tells me this is used backwards.  The value "1.0" would be the scaled on, but I don't know graphics jargon all that well.

src/format.h
```diff
@@ -202,6 +215,10 @@ class Format {
   /// Returns true if all components of this format are a 64 bit float.
   bool IsDouble() const { return AreAllComponents(FormatMode::kSFloat, 64); }
 
+  /// Generates the image format name for this format if possible. Returns
+  /// name if generated or "" otherwise.
```
s-perron: s/Returns name/Returns the name/

src/format.h
dj2: @jaebaek added this code originally ....

src/format.h
dj2: done.

src/format.h
dj2:
```
    float scale_factor_for_error_report = 1.0f;	
    if (component->mode == FormatMode::kUNorm ||	
        component->mode == FormatMode::kSNorm ||	
        component->mode == FormatMode::kSRGB) {	
      scale_factor_for_error_report = 255.0f;	
    }
```

is where this came from.

s-perron: Okay, I think the name is correct.  The output want to construct a string that displays values that are not normalized.  If it is "scaled", in the range 0.0-1.0, then it needs to be multiplied by 255, otherwise, it remains the same.

sRGB stands for "standard RGB".

I this "normalized" would be a better term because UNorm and SNorm stand for unsigned and signed normalized values.

dj2: done.

PR #679, opened 2019-10-02
New type structure

### LINKED ISSUE #603: Update Format class to handle more complicated data structures.
Currently the `Format` class has a `vector` of `Component`s which make up the format. `Format` itself deals with figuring out rows/columns and std140 vs std430. This structure needs to be updated in order to support structures.

Structures are a little different as the other members of the struct can effect the padding for a given member. `float vec3` will pack differently then `vec3 float` with the first taking 32 bytes and the second 16 bytes.

Proposal, the `Format` class will remain but will be abstracted to a second level. The `Format` will contain a set of `Element` classes. Those `Element` classes will then contain the `Component` classes. A `Format` in the current system would have a single `Element` with the `Component` entries that the `Format` would have held.

The `Format` class will then have to develop the smarts to figure out the various packing rules based on what makes up a given `Element` and what the surrounding `Element` classes contain.

This should allow `Format` to continue to work as it does currently, but then be extended to support structures.

This is required before #544 can be completed.

dj2:
This CL replaces the Format;:Components with a new Type hierarchy in
Amber. This allows more flexibility in defining how types are created.

Fixes #603


dj2:
Sorry, this one ended up being big ....

src/format.cc
```diff
@@ -68,74 +89,166 @@ uint32_t Format::InputNeededPerElement() const {
...
+          seg_size += AddSegmentForType(member.type);
+      } else {
+        seg_size = AddSegmentForType(member.type);
+      }
+
+      if (member.HasArrayStride()) {
+        uint32_t array_stride =
+            static_cast<uint32_t>(member.array_stride_in_bytes);
+        assert(seg_size <= array_stride && "Array has more size then stride");
+        seg_size = array_stride;
```
alan-baker: This looks wrong for sized arrays. For int32[4], a valid array stride is 4, but you calculate the segment size as 16. You should probably capture a single element size.

Now, imagine I had int32[4] with an array stride of 8. The overall array requires 28 bytes. So seg_size should be array_stride * (ArraySize() - 1) + elem_size.

src/format.cc
```diff
@@ -68,74 +89,166 @@ uint32_t Format::InputNeededPerElement() const {
...
+      } else if (member.HasMatrixStride()) {
+        uint32_t matrix_stride =
+            static_cast<uint32_t>(member.matrix_stride_in_bytes);
+        assert(seg_size <= matrix_stride && "Matrix has more size then stride");
+        seg_size = matrix_stride;
```
alan-baker: Similarly here, matrix stride describes the distance between vectors not the distance between matrices.

src/format.cc
```diff
@@ -68,74 +89,166 @@ uint32_t Format::InputNeededPerElement() const {
...
+  if (type->IsStruct()) {
+    auto s = type->AsStruct();
+    uint32_t cur_offset = 0;
+    for (const auto& member : s->Members()) {
+      if (member.HasOffset() &&
+          static_cast<uint32_t>(member.offset_in_bytes) > cur_offset) {
+        AddPaddedSegment(static_cast<uint32_t>(member.offset_in_bytes) -
+                         cur_offset);
+        cur_offset = static_cast<uint32_t>(member.offset_in_bytes);
+      }
```
alan-baker: It looks like you're assuming lower members have lower offsets. That's completely reasonable, but not required of the SPIR-V decorations. You should probably add an assert that the member offset is >= cur_offset.

src/type_parser.h
```diff
@@ -12,43 +12,47 @@
...
-  static FormatType NameToType(const std::string& data);
+  static FormatType NameToFormatType(const std::string& data);
 
-  FormatParser();
-  ~FormatParser();
+  TypeParser();
+  ~TypeParser();
 
-  std::unique_ptr<Format> Parse(const std::string& fmt);
+  std::unique_ptr<type::Type> Parse(const std::string& fmt);
```
s-perron: This could probably use a comment to explain what format the string is expected to have.

src/format.cc
dneto0: Specifically Vulkan 14.5.4 "Offset and Stride Assignment"

> The numeric order of `Offset` decorations does not need to follow member declaration order.

src/amberscript/parser.cc
```diff
@@ -1042,17 +1028,21 @@ Result Parser::ParseBufferInitializerSeries(Buffer* buffer,
...
-  bool is_double_data = fmt->IsFloat() || fmt->IsDouble();
-
   Value counter;
-  if (is_double_data)
+
+  auto n = type->AsNumber();
+  FormatMode mode = n->GetFormatMode();
+  uint32_t num_bits = n->NumBits();
+  if (type::Type::IsFloat32(mode, num_bits) ||
+      type::Type::IsFloat64(mode, num_bits)) {
```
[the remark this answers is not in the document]
dj2: I had that originally, then got tired of fixing the duplication. Went with this version and am getting tired of the typing. I'm probably going to add the helpers, but will do it in a followup so I can see what it looks like.

src/format.cc
```diff
@@ -68,74 +89,166 @@ uint32_t Format::InputNeededPerElement() const {
...
+    return true;
+  if (t->IsVec3() || (t->IsMatrix() && t->RowCount() == 3))
+    return true;
+  return false;
+}
 
-  for (size_t i = 0; i < column_count_; ++i) {
-    for (size_t k = 0; k < components_.size(); ++k) {
-      segments_.push_back(Segment{components_[k].get()});
+uint32_t Format::AddSegmentForType(type::Type* type) {
```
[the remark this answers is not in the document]
dj2: done.

src/pipeline.h
```diff
@@ -18,6 +18,7 @@
 #include <map>
 #include <memory>
 #include <string>
+#include <tuple>
```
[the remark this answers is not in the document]
dj2: Removed, was from a previous change that I removed.

src/type_parser.h
dj2: done.

src/amberscript/parser.cc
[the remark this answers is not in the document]
s-perron: Okay.

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
...
+|----------------|-------------------------------------|
+| vec2\<float>[] | <kbd>[bbbb][bbbb]            </kbd> |
+| vec3\<float>[] | <kbd>[bbbb][bbbb][bbbb][----]</kbd> |
+| vec4\<float>[] | <kbd>[bbbb][bbbb][bbbb][bbbb]</kbd> |
+| vec2\<int8>[]  | <kbd>[bb]                    </kbd> |
+| vec2\<int32>[] | <kbd>[bbbb][bbbb]            </kbd> |
+| vec3\<int32>[] | <kbd>[bbbb][bbbb][bbbb][----]</kbd> |
+
+
+## Matrices (All matrices are column matrices)
```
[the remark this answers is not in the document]
dneto0: Also, the notation is that it's C x R matrices.

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
...
+| mat2x3\<float> | <kbd>[bbbb][bbbb][bbbb][----]<br/>[bbbb][bbbb][bbbb][----]                                                          </kbd> |
+| mat2x4\<float> | <kbd>[bbbb][bbbb][bbbb][bbbb]<br/>[bbbb][bbbb][bbbb][bbbb]                                                          </kbd> |
+| mat3x2\<float> | <kbd>[bbbb][bbbb][----][----]<br/>[bbbb][bbbb][----][----]<br/>[bbbb][bbbb][----][----]                             </kbd> |
+| mat4x2\<float> | <kbd>[bbbb][bbbb][----][----]<br/>[bbbb][bbbb][----][----]<br/>[bbbb][bbbb][----][----]<br/>[bbbb][bbbb][----][----]</kbd> |
+| mat4x3\<float> | <kbd>[bbbb][bbbb][bbbb][----]<br/>[bbbb][bbbb][bbbb][----]<br/>[bbbb][bbbb][bbbb][----]<br/>[bbbb][bbbb][bbbb][----]</kbd> |
+
+### STD430
+| Name           | Bytes                                                                                                                      |
+|----------------|----------------------------------------------------------------------------------------------------------------------------|
+| mat2x2\<fp16>  | <kbd>[bbbb][bbbb]                                                                                                   </kbd> |
```
dneto0: This reads misleadingly.
Would prefer a linebreak between the two clumps.

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
...
+| mat2x2\<float> | <kbd>[bbbb][bbbb][bbbb][bbbb]                                                                                       </kbd> |
```
dneto0: Similar here.  Would write it as two rows.

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
...
+| mat2x3\<float> | <kbd>[bbbb][bbbb][bbbb][----]<br/>[bbbb][bbbb][bbbb][----]                                                          </kbd> |
+| mat2x4\<float> | <kbd>[bbbb][bbbb][bbbb][bbbb]<br/>[bbbb][bbbb][bbbb][bbbb]                                                          </kbd> |
+| mat2x2\<int8>  | <kbd>[bbbb]                                                                                                         </kbd> |
+| mat3x2\<float> | <kbd>[bbbb][bbbb]<br/>[bbbb][bbbb]<br/>[bbbb][bbbb]                                                                 </kbd> |
+| mat3x3\<float> | <kbd>[bbbb][bbbb][bbbb][----]<br/>[bbbb][bbbb][bbbb][----]<br/>[bbbb][bbbb][bbbb][----]                             </kbd> |
+| mat4x2\<float> | <kbd>[bbbb][bbbb][bbbb][bbbb]<br/>[bbbb][bbbb][bbbb][bbbb]                                                          </kbd> |
+| mat4x3\<float> | <kbd>[bbbb][bbbb][bbbb][----]<br/>[bbbb][bbbb][bbbb][----]<br/>[bbbb][bbbb][bbbb][----]<br/>[bbbb][bbbb][bbbb][----]</kbd> |
+
+
+## Matrix Array (All matrices are column matrices)
```
dneto0: Probably worth noting that the arraystride equals the base alignment of the matrix.  So there is no padding what is shown here and the next element in the array.

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Structures
+```
+struct {
+  int32 w;
+  float x;
+}
+```
+
+The STD140 pads 8 bytes at the end to become a multiple of 16 bytes.
```
dneto0: But that extra padding only is used if this struct is the element of an array, and that multiple-of-16bytes is the array's stride.

For example, if this struct is nested as the first member of an enclosing struct, and that enclosing struct had a second member which was a float y, then even in std140 that y would be placed at overall offset 8, i.e. within what would normally be considered padding of the inner struct (if you get what I mean).

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
...
+
+```
+struct {
+  int32 w;
+  vec2<float> x;
+  float y;
+}
+```
+
+For both cases, the `vec2` is the largest item, so giving a base alignment of
```
dneto0: "largest item" -> "member with the largest base alignment requirement"  (to be perfectly explicit about it)

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
...
+8 bytes. The STD140 cases pads 8 bytes at the end to become a multiple of 16.
```
dneto0: again this is in the context of this struct being the array element type.

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
...
+```
+struct {
+  int32 w;
+  float x;
+}
+```
+
+The STD140 pads 8 bytes at the end to become a multiple of 16 bytes.
+
+| STD | Stride | Bytes                                                                    |
```
dneto0: Recommend changing "Stride" to "Stride when used as array element"  or "Array element stride", or say it in prose somewhere.

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
+# Memory Layout Examples
+
+Note, this is how we'll layout the data in Amber. Some of these types don't
+exist in GLSL (like the int8 and int16 types). We allow them in Amber to give
```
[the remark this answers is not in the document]
dj2: Updated.

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
...
+flexibility in the data provided, but attempt to follow the layout rules with
+all types for consistency.
+
```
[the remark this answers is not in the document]
dj2: Updated

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
...
+## Scalars
```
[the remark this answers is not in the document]
dj2: done

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
...
+
+|  Name  | Bytes                    |
+|--------|--------------------------|
+| int8   | <kbd>[b]          </kbd> |
+| uint8  | <kbd>[b]          </kbd> |
+| int16  | <kbd>[bb]         </kbd> |
+| uint16 | <kbd>[bb]         </kbd> |
+| int32  | <kbd>[bbbb]       </kbd> |
+| uint32 | <kbd>[bbbb]       </kbd> |
```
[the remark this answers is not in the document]
dj2: added

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
...
+| int64  | <kbd>[bbbb][bbbb]</kbd> |
+| uint64 | <kbd>[bbbb][bbbb]</kbd> |
+| double | <kbd>[bbbb][bbbb]</kbd> |
+
+## Vectors
+### STD140 & STD430
+
+| Name        | Bytes                               |
+|-------------|-------------------------------------|
+| vec2<float> | <kbd>[bbbb][bbbb]            </kbd> |
```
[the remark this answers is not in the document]
dj2: fixed

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Vector Arrays
+### STD140
+| Name           | Bytes                               |
+|----------------|-------------------------------------|
+| vec2\<float>[] | <kbd>[bbbb][bbbb][----][----]</kbd> |
+| vec3\<float>[] | <kbd>[bbbb][bbbb][bbbb][----]</kbd> |
+| vec4\<float>[] | <kbd>[bbbb][bbbb][bbbb][bbbb]</kbd> |
+| vec2\<int8>[]  | <kbd>[bb--][----][----][----]</kbd> |
+| vec2\<int32>[] | <kbd>[bbbb][bbbb][----][----]</kbd> |
```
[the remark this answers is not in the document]
dj2: done, please check.

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
...
+
+### STD430
+| Name           | Bytes                               |
+|----------------|-------------------------------------|
+| vec2\<float>[] | <kbd>[bbbb][bbbb]            </kbd> |
+| vec3\<float>[] | <kbd>[bbbb][bbbb][bbbb][----]</kbd> |
+| vec4\<float>[] | <kbd>[bbbb][bbbb][bbbb][bbbb]</kbd> |
+| vec2\<int8>[]  | <kbd>[bb]                    </kbd> |
```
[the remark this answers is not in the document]
dj2: done

docs/memory_layout.md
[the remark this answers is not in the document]
dj2: done

docs/memory_layout.md
```diff
@@ -0,0 +1,230 @@
...
+| vec2\<int32>[] | <kbd>[bbbb][bbbb]            </kbd> |
+| vec3\<int32>[] | <kbd>[bbbb][bbbb][bbbb][----]</kbd> |
+
+
+## Matrices (All matrices are column matrices)
+### STD140
+| Name           | Bytes                                                                                                                      |
+|----------------|----------------------------------------------------------------------------------------------------------------------------|
+| mat2x2\<int8>  | <kbd>[bb--][----][----][----]<br/>[bb--][----][----][----]                                                          </kbd> |
```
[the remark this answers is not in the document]
dj2: Added a note.

docs/memory_layout.md
dj2: done

docs/memory_layout.md
dj2: done

docs/memory_layout.md
dj2: done

docs/memory_layout.md
dj2: I added this example as another entry, it doesn't seem to pad the way you expect, heh.

docs/memory_layout.md
dj2: done

docs/memory_layout.md
dj2: done

docs/memory_layout.md
dj2: done

docs/memory_layout.md
```diff
@@ -0,0 +1,300 @@
...
+
+The STD140 pads 8 bytes at the end to become a multiple of 16 bytes.
+
+| STD | Array Stride | Bytes |
+|-----|:--------------------:|-------|
+| 140 | 16 | <kbd>{w [bbbb]}<br/>{x [bbbb]}<br/> [----][----]</kbd> |
+| 430 |  8 | <kbd>{w [bbbb]}<br/>{x [bbbb]}</kbd> |
+
+<hr>
+```
```
alan-baker: This isn't formatting correctly.

docs/memory_layout.md
```diff
@@ -0,0 +1,300 @@
...
+| 430 | 24 | <kbd>{w [bbbb][----]}<br/>{x [bbbb][bbbb]}<br/>{y [bbbb][----]}</kbd> |
+
+<hr>
+
+
+```
+struct {
+  int32 w;
+  vec3<float> x;
+  float y;
```
alan-baker: Probably worth a case where `y` is vec2<float>. Then `y` can't be packed.

docs/memory_layout.md
```diff
@@ -0,0 +1,300 @@
...
+
+In STD140 the `mat2x2<float>` has a base alignment of 16 bytes (the mat2x2 is, effectively,
+an array of vec2's The vec2's have a size of 8 bytes this size rounds up to a vec4, so 16 bytes).
+
+In STD430, the round up doesn't happen, so the base alignment is 8 bytes.
+
+| STD | Array Stride  | Bytes |
+|-----|:---------------------:|-------|
+| 140 | 64 | <kbd>{w [bbbb][----][----][----]}<br/>{x [bbbb][bbbb][----][----]<br/> &nbsp;&nbsp; [bbbb][bbbb][----][----]}<br/>{y [bbbb][----][----][----]}</kbd> |
+| 430 | 48 | <kbd>{w [bbbb][----]}<br/>{x [bbbb][bbbb][bbbb][bbbb]}<br/>{y [bbbb][----]}</kbd> |
```
alan-baker: Isn't the array stride for the 430 case only 32?

docs/memory_layout.md
alan-baker: > The Offset decoration of a member must not place it between the end of a structure or an array and the next multiple of the alignment of that structure or array.

That rule is why it gets padded.

docs/memory_layout.md
```diff
@@ -0,0 +1,300 @@
...
+## Vectors
+### STD140 & STD430
+
+| Name | Bytes |
+|------|-------|
+| vec2\<float> | <kbd>[bbbb][bbbb]</kbd> |
+| vec3\<float> | <kbd>[bbbb][bbbb][bbbb][----]</kbd> |
+| vec4\<float> | <kbd>[bbbb][bbbb][bbbb][bbbb]</kbd> |
+| vec2\<int8>  | <kbd>[bb]</kbd> |
+| vec3\<int8>  | <kbd>[bbb-][----][----][----]</kbd> |
```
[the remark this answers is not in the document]
dj2: Right, should be `[bbbb]`. Also added the int16 cases in to make it clearer.

docs/memory_layout.md
dj2: fixed ... i think

docs/memory_layout.md
dj2: done

docs/memory_layout.md
dj2: yes, yes it is.

docs/memory_layout.md
alan-baker: Looks correct when I view it now.

src/format.cc
dj2: Done, line 260

PR #680, opened 2019-10-02
Initialize data buffer.

dj2:
This CL does an explicit memset of the data buffer to make sure it is
initialized with zeros.

### DIFF
--- a/src/buffer.cc
+++ b/src/buffer.cc
@@ -239,6 +239,15 @@ Result Buffer::SetDataWithOffset(const std::vector<Value>& data,
   // this maybe the first time data is set into the buffer.
   bytes_.resize(GetSizeInBytes());
 
+  // Set the new memory to zero to be on the safe side.
+  int32_t new_space = static_cast<int32_t>(
+      (static_cast<uint32_t>(data.size()) / format_->InputNeededPerElement()) *
+      format_->SizeInBytes());
+  assert(static_cast<uint32_t>(new_space) + offset <= GetSizeInBytes());
+
+  if (new_space > 0)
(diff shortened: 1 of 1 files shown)


src/buffer.cc
```diff
@@ -239,6 +239,15 @@ Result Buffer::SetDataWithOffset(const std::vector<Value>& data,
...
   bytes_.resize(GetSizeInBytes());
 
+  // Set the new memory to zero to be on the safe side.
+  int32_t new_space = static_cast<int32_t>(
+      (static_cast<uint32_t>(data.size()) / format_->InputNeededPerElement()) *
+      format_->SizeInBytes());
+  assert(static_cast<uint32_t>(new_space) + offset <= GetSizeInBytes());
+
+  if (new_space > 0)
+    memset(bytes_.data() + offset, new_space, sizeof(uint8_t));
```
s-perron: This does not look correct.  The prototype for memset is:

```
void * memset ( void * ptr, int value, size_t num );
```

This will set the single byte (`sizeof(uint8_t)`) to the value `new_space`, which will probably not be zero.

I think you want something like:

```
memset(bytes_.data() + offset, 0, new_space);
```

Also the more I think about it, we cannot avoid the ansi aliasing violation here.  The call to "resize" might move data around, which will not alias the uses later on.  Very unlikely that this will cause a problem though.  In this particular case, I do not see any reasonable way for the compiler could do anything.  There are other reason for it remain more conservative.

dj2: You mean, I can't treat memset like it's malloc and have the compiler figure out what I'm doing? damn.

Good catch, fixed.

PR #688, opened 2019-10-10
[AmberScript] Adding struct support

### LINKED ISSUE #544: AmberScript: Support a buffer with values of multiple types
Since the current AmberScript has only the following way to initialize a buffer with specific data:

```
BUFFER {name} DATA_TYPE {type} DATA
_value_+
END
```

it is impossible test the following case:

```
#!amber

SHADER compute kComputeShader HLSL
struct struct_with_bool {
  bool2 elem_v2bool;
  int2 elem_v2int;
  float2 elem_v2float;
  bool elem_bool;
  int elem_int;
  float elem_float;
};

ConsumeStructuredBuffer<struct_with_bool> consume_struct_with_bool;
AppendStructuredBuffer<bool> append_bool;
AppendStructuredBuffer<float> append_float;
AppendStructuredBuffer<int> append_int;

void main() {
  append_bool.Append(consume_struct_with_bool.Consume().elem_bool);
  append_int.Append(consume_struct_with_bool.Consume().elem_int);
  append_float.Append(consume_struct_with_bool.Consume().elem_float);
  append_v2bool.Append(consume_struct_with_bool.Consume().elem_v2bool);
  append_v2int.Append(consume_struct_with_bool.Consume().elem_v2int);
  append_v2float.Append(consume_struct_with_bool.Consume().elem_v2float);
}
END

BUFFER kConsumeBuffer DATA_TYPE float DATA
4.0 19.0
END
BUFFER kConsumeBufferIndex DATA_TYPE uint32 DATA
2
END

...
```

dj2:
This CL adds support for the STRUCT syntax in AmberScript. STRUCTS can
be used in buffers. They can be compared as expected. If no OFFSET or
STRIDE data is provided we will calculate the padding and offsets based
on the current layout.

Fixes #544

### DIFF
--- a/src/amberscript/parser.cc
+++ b/src/amberscript/parser.cc
@@ -16,6 +16,7 @@
 
 #include <cassert>
 #include <limits>
+#include <map>
 #include <string>
 #include <utility>
 #include <vector>
@@ -966,6 +967,8 @@ Result Parser::ParseStruct() {
           return Result("missing value for STRUCT member ARRAY_STRIDE");
         if (!token->IsInteger())
           return Result("invalid value for STRUCT member ARRAY_STRIDE");
+        if (!member_type->IsArray())
+          return Result("ARRAY_STRIDE only valid on array members");
 
         m->array_stride_in_bytes = token->AsInt32();
       } else if (token->AsString() == "MATRIX_STRIDE") {
@@ -974,6 +977,8 @@ Result Parser::ParseStruct() {
           return Result("missing value for STRUCT member MATRIX_STRIDE");
         if (!token->IsInteger())
           return Result("invalid value for STRUCT member MATRIX_STRIDE");
+        if (!member_type->IsMatrix())
+          return Result("MATRIX_STRIDE only valid on matrix members");
 
         m->matrix_stride_in_bytes = token->AsInt32();
       } else {
--- a/src/amberscript/parser.cc
+++ b/src/amberscript/parser.cc
@@ -163,6 +163,8 @@ Result Parser::Parse(const std::string& data) {
       r = ParseSet();
     } else if (tok == "SHADER") {
       r = ParseShaderBlock();
+    } else if (tok == "STRUCT") {
+      r = ParseStruct();
     } else {
       r = Result("unknown token: " + tok);
     }
@@ -521,7 +523,7 @@ Result Parser::ParseShaderSpecialization(Pipeline* pipeline) {
 
   auto type = ToType(token->AsString());
   if (!type)
-    return Result("invalid data_type provided");
+    return Result("invalid data type '" + token->AsString() + "' provided");
   if (!type->IsNumber())
     return Result("only numeric types are accepted for specialization values");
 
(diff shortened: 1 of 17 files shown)


src/amberscript/parser.cc
```diff
@@ -1368,6 +1513,9 @@ Result Parser::ParseValues(const std::string& name,
 
       v.SetIntValue(token->AsUint64());
     }
+    ++seg_idx;
```
alan-baker: Just to confirm, vec4<float> is 4 segments? How are runtime arrays to be handled?

src/amberscript/parser_buffer_test.cc
```diff
@@ -762,5 +765,142 @@ INSTANTIATE_TEST_SUITE_P(
                     NameData{"mat2x2"},
                     NameData{"mat2x2<>"}));  // NOLINT(whitespace/parens)
 
+TEST_F(AmberScriptParserTest, BufferWithStructStd140) {
```
alan-baker: Need some tests with vectors, arrays and runtime arrays.

src/amberscript/parser.cc
alan-baker: This seems to allow partial initialization because I don't see a check that all the segments are intialized. Just want to confirm it is intentional. I haven't noticed tests for it.

src/amberscript/parser.cc
```diff
@@ -875,6 +878,120 @@ Result Parser::ParsePipelineSet(Pipeline* pipeline) {
...
+    while (token->IsString()) {
+      if (token->AsString() == "OFFSET") {
+        token = tokenizer_->NextToken();
+        if (token->IsEOL())
+          return Result("missing value for STRUCT member OFFSET");
+        if (!token->IsInteger())
+          return Result("invalid value for STRUCT member OFFSET");
+
+        m->offset_in_bytes = token->AsInt32();
+      } else if (token->AsString() == "ARRAY_STRIDE") {
```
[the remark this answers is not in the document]
dj2: done

src/amberscript/parser.cc
```diff
@@ -875,6 +878,120 @@ Result Parser::ParsePipelineSet(Pipeline* pipeline) {
...
+        token = tokenizer_->NextToken();
+        if (token->IsEOL())
+          return Result("missing value for STRUCT member ARRAY_STRIDE");
+        if (!token->IsInteger())
+          return Result("invalid value for STRUCT member ARRAY_STRIDE");
+
+        m->array_stride_in_bytes = token->AsInt32();
+      } else if (token->AsString() == "MATRIX_STRIDE") {
```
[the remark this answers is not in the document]
dj2: done

src/amberscript/parser.cc
dj2: A vec4 would be 4 segments, yes. My plan for runtime arrays is that you end up looping over that group of segments again. So, the start of the array will be flagged and you can have 0 or more of them as needed.

dj2: Added test. The buffer code will reject a mismatched number of items currently.

src/amberscript/parser_buffer_test.cc
dj2: Added vec test. Array and runtime array will be followups when we add the array syntax.

Issue #327, opened 2019-03-04
create AmberScript sample showing image processing flow


dneto0:
I'm not sure if all the features are planned yet, but I'd like to support a script like this:

  First pipeline:  Fills one or more images with preprogrammed data.

  Second pipeline:  An image processing kernel (e.g. blur).  This is the actual intended "shader under test"

  Third pipeline:  Render the image to an output framebuffer.  (So we can write it out with "amber -i foo.png")

This way we can test image processing flows without having to read image files.

asuonpaa:
I had a similar feature request in mind, but for graphics pipeline: the first pipeline renders something to a color buffer and the second would use that as an input texture. I'd be more than happy to contribute as it would open up lots of opportunities when using Amber in Vulkan CTS.

I'm seeing several issues related to image support. Would someone explain what's the current status for this work, and what I could do to help?

dj2:
This should all be possible now. You can have multiple pipelines and you can pass buffers between them. So, for the example you provide, someone just needs to create a .amber file (we have to use the AmberScript format for this) to show generating an image in one pipeline and then using it as an image in the second pipeline.

I haven't worked with images in Vulkan yet, when you attach them to the pipeline, are they just buffers or do you need to setup something more then a buffer? If so, we need to figure out some syntax to handle that, and I'd need to see an example of what we need to do.

asuonpaa:
Textures would work pretty much the same way the frame buffer does right now. It's a VkImage object with a view provided by VkImageView. But instead of using the image view as an attachment, the texture would be accessed using a descriptor like, for example, an ssbo would. The descriptor type would be a combined image sampler. Here's a  [Vulkan tutorial]( https://vulkan-tutorial.com/Texture_mapping/Combined_image_sampler) for that. Buffers can be used to copy data to and from the texture image the same way as with frame buffer.

My requirement for texture support is based on a CTS test I'm working on. There I would need to access different mip levels based on a vertex ID. So I'd need a way to specify how many mip levels the texture has. And in this case each level needs to be cleared to a different color. Someone else might use autogenerated mip levels, or maybe a pipeline that would bind a specific mip level as a frame buffer. In fact a graphics pipeline could be used for generating the regular mip maps where each level has a scaled down version of the previous level. Just specify correct size and level of the mip map and draw a filtered texture there.


Below is a draft amber file which shows how I would be using the texture support.

```
SHADER vertex vert_shader_clear PASSTHROUGH

SHADER fragment frag_shader GLSL
[...]

PR #694, opened 2019-10-17
Added storage image support.

asuonpaa:
Fixes https://github.com/google/amber/issues/327

### DIFF
--- a/src/amberscript/parser.cc
+++ b/src/amberscript/parser.cc
@@ -654,8 +654,8 @@ Result Parser::ToBufferType(const std::string& name, BufferType* type) {
     *type = BufferType::kUniform;
   else if (name == "storage")
     *type = BufferType::kStorage;
-  else if (name == "texture")
-    *type = BufferType::kTexture;
+  else if (name == "storage_image")
+    *type = BufferType::kStorageImage;
   else
     return Result("unknown buffer_type: " + name);
 
@@ -707,8 +707,8 @@ Result Parser::ParsePipelineBind(Pipeline* pipeline) {
       Result r = pipeline->SetPushConstantBuffer(buffer);
       if (!r.IsSuccess())
         return r;
-    } else if (token->AsString() == "texture") {
-      buffer->SetBufferType(BufferType::kTexture);
+    } else if (token->AsString() == "storage_image") {
+      buffer->SetBufferType(BufferType::kStorageImage);
     } else {
       BufferType type = BufferType::kColor;
       Result r = ToBufferType(token->AsString(), &type);
@@ -725,7 +725,7 @@ Result Parser::ParsePipelineBind(Pipeline* pipeline) {
   if (buffer->GetBufferType() == BufferType::kUnknown ||
       buffer->GetBufferType() == BufferType::kStorage ||
       buffer->GetBufferType() == BufferType::kUniform ||
-      buffer->GetBufferType() == BufferType::kTexture) {
+      buffer->GetBufferType() == BufferType::kStorageImage) {
     // If AS was parsed above consume the next token.
     if (buffer->GetBufferType() != BufferType::kUnknown)
       token = tokenizer_->NextToken();
--- a/src/amberscript/parser.cc
+++ b/src/amberscript/parser.cc
@@ -16,6 +16,7 @@
 
 #include <cassert>
 #include <limits>
+#include <map>
 #include <string>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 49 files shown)


src/vulkan/buffer_descriptor.h
```diff
@@ -31,14 +32,9 @@ namespace vulkan {
...
 
-enum class DescriptorType : uint8_t {
-  kStorageBuffer = 0,
-  kUniformBuffer,
-};
-
 /// Stores descriptor set and binding information for storage and uniform
 /// buffers.
-class BufferDescriptor {
+class BufferDescriptor : public Descriptor {
```
dj2: How difficult would it be to put the changes for splitting out the Descriptor class into a separate CL which is mostly just moving code around? (Similar for, I think, the image resource code that was split out? We should be able to split and land it pretty quickly.

asuonpaa: Most likely it sounds easier than it is and would require some kind of intermediate code that would then be discarded for the final result. I'd rather get the storage image support into good shape quickly and then add other descriptor types one by one as separate PRs.

src/vulkan/pipeline.cc
```diff
@@ -296,15 +304,17 @@ Result Pipeline::AddDescriptor(const BufferCommand* cmd) {
...
-  auto* buf_desc = static_cast<BufferDescriptor*>(desc);
-  if (cmd->GetValues().empty()) {
-    Result r = buf_desc->SetSizeInElements(cmd->GetBuffer()->ElementCount());
-    if (!r.IsSuccess())
-      return r;
-  } else {
-    Result r = buf_desc->AddToBuffer(cmd->GetValues(), cmd->GetOffset());
-    if (!r.IsSuccess())
-      return r;
+  if (desc->IsStorageBuffer() || desc->IsUniformBuffer()) {
```
dj2: Is the assumption here that htere is no buffer data for image buffers? I don't think that's a valid assumption to make as we may have image data input, yea?

src/amberscript/parser.cc
```diff
@@ -705,6 +707,8 @@ Result Parser::ParsePipelineBind(Pipeline* pipeline) {
       Result r = pipeline->SetPushConstantBuffer(buffer);
       if (!r.IsSuccess())
         return r;
+    } else if (token->AsString() == "storage_image") {
```
dj2: Can you add a parser unit test for storage_image?

src/amberscript/parser_bind_test.cc
```diff
@@ -1325,5 +1325,121 @@ END)";
...
+BUFFER texture FORMAT R8G8B8A8_UNORM
+
+PIPELINE compute pipeline
+  ATTACH compute_shader
+  BIND BUFFER texture AS storage_image DESCRIPTOR_SET 0 BINDING 0
+END)";
+
+  Parser parser;
+  Result r = parser.Parse(in);
+  ASSERT_TRUE(r.IsSuccess());
```
dj2: Can you also pull out the buffer and verify it's set as `storage_image`?

dj2: ASSERT_TRUE(r.IsSuccess()) << r.Error();

src/amberscript/parser_bind_test.cc
```diff
@@ -1325,5 +1325,121 @@ END)";
...
+PIPELINE graphics pipeline
+  ATTACH vert_shader
+  ATTACH frag_shader
+  BIND BUFFER texture AS storage_image DESCRIPTOR_SET 0 BINDING 0
+  BIND BUFFER framebuffer AS color LOCATION 0
+END)";
+
+  Parser parser;
+  Result r = parser.Parse(in);
+  ASSERT_TRUE(r.IsSuccess());
```
dj2: ditto

src/vulkan/descriptor.cc
```diff
@@ -44,11 +44,9 @@ VkDescriptorType Descriptor::GetVkDescriptorType() const {
       return VK_DESCRIPTOR_TYPE_STORAGE_IMAGE;
     case DescriptorType::kCombinedImageSampler:
       return VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER;
-    case DescriptorType::kSampledImage:
-      return VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE;
     default:
-      assert(false && "unexpected descriptor type");
-      return VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER;
+      assert(type_ == DescriptorType::kSampledImage);
```
dj2: Why make this change instead of what was here previously? I think the original was clearer.

asuonpaa: CI clang build was failing because of having default even when all enums where covered.

dj2:
Looks like you need to add your new files to the Makefile and the BUILD.gn files.

asuonpaa:
> Looks like you need to add your new files to the Makefile.

Which makefile? Those are already in CMakeLists.txt.

PR #696, opened 2019-10-21
Specify comparison parameters after comparison algorithm

asuonpaa:
Prepare image_diff tool to allow different parameters per comparison algorithm. Now all comparison parameters are expected to be provided after the algorithm parameter.
This change removes the separate tolerance parameter as future comparators might not use it.

### DIFF
--- a/samples/image_diff.cc
+++ b/samples/image_diff.cc
@@ -71,8 +71,8 @@ bool ParseArgs(const std::vector<std::string>& args, Options* opts) {
         return false;
       }
       opts->tolerance = std::stof(std::string(args[i]));
-      if (opts->tolerance < 0) {
-        std::cerr << "Tolerance must be non-negative." << std::endl;
+      if (opts->tolerance < 0 || opts->tolerance > 255) {
+        std::cerr << "Tolerance must be in the range 0..255." << std::endl;
         return false;
       }
     } else if (arg == "--histogram_emd") {
--- a/samples/image_diff.cc
+++ b/samples/image_diff.cc
@@ -38,7 +38,10 @@ struct Options {
 };
 
 const char kUsage[] = R"(Usage: image_diff [options] image1.png image2.png
-options:
+
+Exactly one algorithm (and its parameters) must be specified.
+
+Algorithms:
 
   --rmse TOLERANCE
                Compare using the Root Mean Square Error (RMSE) algorithm with
@@ -54,6 +57,8 @@ options:
                different histograms for at least one of the color channels.
                E.g. an image with red=255 for every pixel vs. an image with
                red=0 for every pixel.
+
+Other options:
 
   -h | --help  This help text.
 )";
--- a/samples/image_diff.cc
+++ b/samples/image_diff.cc
@@ -36,12 +36,24 @@ struct Options {
 };
 
 const char kUsage[] = R"(Usage: image_diff [options] image1.png image2.png
+options:
 
- options:
-  --rmse                    -- Compare using RMSE algorithm (default).
-  --histogram_emd           -- Compare using histogram EMD algorithm.
-  -t | --tolerance <float>  -- Tolerance value for comparison.
-  -h | --help               -- This help text.
+  --rmse TOLERANCE
+               Compare using the Root Mean Square Error (RMSE) algorithm with
+               a floating point TOLERANCE value in the range 0..255, where 0
+               indicates identical images and 255 indicates images where every
+               color channel has the maximum difference.
+
+  --histogram_emd TOLERANCE
+               Compare the per-channel color histograms of the images using a
+               variant of the Earth Mover's Distance (EMD) algorithm with a
+               floating point TOLERANCE value in the range 0.0..1.0, where 0.0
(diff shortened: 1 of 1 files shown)


samples/image_diff.cc
```diff
@@ -36,12 +36,24 @@ struct Options {
...
-  -h | --help               -- This help text.
+options:
+
+  --rmse TOLERANCE
+               Compare using the Root Mean Square Error (RMSE) algorithm with
+               a floating point TOLERANCE value in the range 0..255, where 0
+               indicates identical images and 255 indicates images where every
+               color channel has the maximum difference.
+
+  --histogram_emd TOLERANCE
```
dj2: How is this going to work if there are multiple parameters? Listing them, something else?

asuonpaa: Not sure what you mean. Is there something preventing us to syntax below?

`--algorithm PARAM1 PARAM2`

I suppose if there were several parameters we would have to modify the layout.

samples/image_diff.cc
```diff
@@ -52,18 +64,28 @@ bool ParseArgs(const std::vector<std::string>& args, Options* opts) {
       return true;
     } else if (arg == "--rmse") {
       opts->compare_algorithm = CompareAlgorithm::kRMSE;
+      ++i;
+      if (i >= args.size()) {
+        std::cerr << "Missing tolerance value for RMSE comparison."
+                  << std::endl;
+        return false;
+      }
+      opts->tolerance = std::stof(std::string(args[i]));
```
asuonpaa: This won't catch an issue where user has forgotten to specify the tolerance value. If we have `--rmse image1.png` it would try to read image1.png as a number and then crash when trying to throw an exception. Compile options forbid the use of catch here.

samples/image_diff.cc
dj2: My concern was if this would get confusing. If I have a script which says:

`image_diff --foo 1 2 3 4 45 6 $1 $2` that's hard to read. Where:

`image_diff --foo --tolerance 1 --other 2 --thing 3 ...` is longer but potentially easier.

paulthomson: Yes, there is a trade-off here. We want the parameters to be specific to each algorithm. `--tolerance` looks optional, it looks like it can go anywhere, and it looks like it applies globally. We want to be more specific. But the downside is as you say. However, this is a common issue with many conventional command line interfaces (and indeed C/C++ IMO). If we want to go against convention, we could match AmberScript: 

```
image_diff \
  "EMD_WINDOW TOLERANCE 7 WINDOW_SIZE 4" \
  "EMD TOLERANCE 7" \
  "RMSE TOLERANCE 7"
```

But I think losing some readability might be OK here since this is not a scripting language and our help is pretty clear.

paulthomson: Another option:

```
image_diff \
  --histogram_emd_window=tolerance:7.0;window_size:4;
  --histogram_emd=tolerance:3.0;
  --rmse=tolerance:1.0;
```

dj2: I assumed this would do one type of image comparison at a time, are you thinking you want it to be able to do multiple? My assumption was you'd do something like:

```
image_diff --rmse --tolerance 2.3
image_diff --histogram_emd --tolerance 1.4 --window-size 4
```

but you want to do both of them in one command? What's the output in that case, if rmse passes and histogram_emd fails, does image_diff pass or fail?

paulthomson: Indeed, we don't need to run more than one algorithm. I think we agree to stick with the syntax in this PR, even though it may not always be obvious what the numbers mean.

samples/image_diff.cc
asuonpaa: Yeah, sscanf seems to work nicely.

samples/image_diff.cc
```diff
@@ -36,40 +38,79 @@ struct Options {
 };
 
 const char kUsage[] = R"(Usage: image_diff [options] image1.png image2.png
-
- options:
-  --rmse                    -- Compare using RMSE algorithm (default).
-  --histogram_emd           -- Compare using histogram EMD algorithm.
-  -t | --tolerance <float>  -- Tolerance value for comparison.
-  -h | --help               -- This help text.
+options:
```
paulthomson: I would maybe have:

```
Usage: image_diff [options] image1.png image2.png

Exactly one algorithm (and its parameters) must be specified.

Algorithms:

  --rmse ...

Other options:
   -h | --help ...
```

asuonpaa: Done

asuonpaa:
Looks like MSVC doesn't like sscanf, but insists of using sscanf_s instead. Do we want to do ifdefs to use that with MSVC?

dj2:
http://www.cplusplus.com/reference/cstdlib/strtod/ is what the tokenizer is using and it doesn't throw.

asuonpaa:
> http://www.cplusplus.com/reference/cstdlib/strtod/ is what the tokenizer is using and it doesn't throw.

That has the original issue of not detecting invalid strings. Paul suggested using string stream so I'll try that next.

PR #698, opened 2019-10-24
Add Vulkan device selection support

### LINKED ISSUE #619: Support choosing Vulkan device
Currently Amber picks the first VkPhysicalDevice that meets its feature and extension requirements. Setting `VK_ICD_FILENAMES` is sufficient in most cases but not when trying to choose between two devices using the same ICD.

My idea was to add two new command line options:

* List physical Vulkan devices with IDs and exit
* Specify Vulkan device ID to run with

mvainola:
Add the `-D` command line option for selecting a physical Vulkan device to run with. Only choose from enumerated physical devices in ChooseVulkanPhysicalDevice and move feature and extension checks to a new CheckVulkanPhysicalDeviceRequirements function.

I tried to add support for Dawn as well but had issues building Amber with Dawn support.

Fixes #619

### DIFF
--- a/samples/config_helper_vulkan.cc
+++ b/samples/config_helper_vulkan.cc
@@ -699,9 +699,85 @@ amber::Result ConfigHelperVulkan::CreateDebugReportCallback() {
   return {};
 }
 
-amber::Result ConfigHelperVulkan::ChooseVulkanPhysicalDevice(
+amber::Result ConfigHelperVulkan::CheckVulkanPhysicalDeviceRequirements(
+    const VkPhysicalDevice physical_device,
     const std::vector<std::string>& required_features,
     const std::vector<std::string>& required_extensions) {
+  VkPhysicalDeviceFeatures required_vulkan_features =
+      VkPhysicalDeviceFeatures();
+
+  if (use_physical_device_features2_) {
+    VkPhysicalDeviceVariablePointerFeaturesKHR var_ptrs =
+        VkPhysicalDeviceVariablePointerFeaturesKHR();
+    var_ptrs.sType =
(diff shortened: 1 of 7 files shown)


mvainola:
@googlebot I signed it!

samples/config_helper_vulkan.cc
```diff
@@ -716,72 +792,23 @@ amber::Result ConfigHelperVulkan::ChooseVulkanPhysicalDevice(
...
-          NamesToVulkanFeatures(required_features1, &required_vulkan_features);
-      if (!r.IsSuccess())
-        return r;
-
-    } else {
-      amber::Result r =
-          NamesToVulkanFeatures(required_features, &required_vulkan_features);
+  if (selected_device > -1) {
+    uint32_t deviceID = static_cast<uint32_t>(selected_device);
+    if (deviceID >= count)
```
dj2: nit: Add {}'s as the body is multi-line

samples/config_helper_vulkan.h
```diff
@@ -44,6 +44,7 @@ class ConfigHelperVulkan : public ConfigHelperImpl {
   amber::Result CreateConfig(
       uint32_t engine_major,
       uint32_t engine_minor,
+      int32_t selected_device,
```
dj2: This going to at least need to be added to config_helper_dawn or the build will fail, yea?

samples/config_helper_vulkan.cc
mvainola: Done

samples/config_helper_vulkan.h
mvainola: Added the parameter to Dawn's CreateConfig. I wasn't able to build with Dawn support so there might still be something else I missed.

PR #702, opened 2019-10-25
Define clear colour default in amberscript.md.

### LINKED ISSUE #701: Clear Color default
Is the clear color defaulted to `[0,0,0,0]`? If so is this at an Amber level, Engine level, AmberScript level, etc? It would be nice if that was defined instead of having to stumble across `clear.amber`

dj2:
Fixes #701


dj2:
TBR=dneto0

PR #703, opened 2019-10-28
Add tests for EMD comparison & histogram

Vihanakangas:
Created tests for GetHistogramForChannel & CompareHistogramEMD.


Vihanakangas:
@googlebot I signed it!

dj2:
The code LG to me, but I'd like @paulthomson to take a look before approving.

src/buffer_test.cc
```diff
@@ -124,4 +124,152 @@ TEST_F(BufferTest, SizeMatrixPaddedStd430) {
...
+
+  for (uint8_t i = 0; i < 4; i++) {
+    std::vector<uint64_t> bins = b.GetHistogramForChannel(i, 256);
+    for (uint64_t y = 0; y < values.size(); y++)
+      EXPECT_EQ(10, bins[255]);
+  }
+}
+
+TEST_F(BufferTest, CompareHistogramEMDToleranceFalse) {
+  TypeParser parser;
```
paulthomson: Please add something like this to the top of the function:

```cpp
// Creates two sets of equal pixel values, except for one pixel that has +50 in its red channel.
// Compares the histograms to see if they are equal with a low threshold, which we expect to fail.
```

src/buffer_test.cc
```diff
@@ -124,4 +124,152 @@ TEST_F(BufferTest, SizeMatrixPaddedStd430) {
...
+  auto type = parser.Parse("R8G8B8A8_UINT");
+  Format fmt(type.get());
+
+  std::vector<Value> values1(40);
```
paulthomson:
```cpp
// Creates 10 RGBA pixel values, with the red channels going from 0 to 255 (a gradient).
```

src/buffer_test.cc
```diff
@@ -124,4 +124,152 @@ TEST_F(BufferTest, SizeMatrixPaddedStd430) {
   EXPECT_EQ(12U * sizeof(int32_t), b.GetSizeInBytes());
 }
 
+TEST_F(BufferTest, GetHistogramForChannelGradient) {
+  TypeParser parser;
```
paulthomson:
```cpp
// Creates 10 RGBA pixel values, with the blue channels ranging from 0 to 255. Checks that the bin for each blue channel value contains 1, as expected.
```

src/buffer_test.cc
```diff
@@ -124,4 +124,152 @@ TEST_F(BufferTest, SizeMatrixPaddedStd430) {
...
+  auto type = parser.Parse("R8G8B8A8_UINT");
+  Format fmt(type.get());
+
+  std::vector<Value> values(40);
```
paulthomson:
```cpp
Creates 10 RBGA pixel values with the blue channels ranging from 0 to 255.
```

src/buffer_test.cc
```diff
@@ -124,4 +124,152 @@ TEST_F(BufferTest, SizeMatrixPaddedStd430) {
...
+
+  Buffer b2(BufferType::kColor);
+  b2.SetFormat(&fmt);
+  b2.SetData(values2);
+
+  EXPECT_TRUE(b1.CompareHistogramEMD(&b2, 0.0f).IsSuccess());
+}
+
+TEST_F(BufferTest, CompareHistogramEMDToleranceAllWhite) {
+  TypeParser parser;
```
paulthomson:
```cpp
// Creates two identical sets of RBGA pixel values and checks that the histograms are equal. 
```

paulthomson:
To fix the formatting, I recommend obtaining clang-format-5.0 as seen in the [CI script](https://github.com/google/amber/blob/master/kokoro/check-format/build.sh), and figuring out how to get it to run on file save for your preferred editor or IDE.

dj2:
There is also `./tools/format` which will run clang-format over all the files for you. `./tools/run-lint.sh` will run the linter.

src/buffer_test.cc
Vihanakangas: Done.

src/buffer_test.cc
Vihanakangas: Done.

src/buffer_test.cc
Vihanakangas: Done.

src/buffer_test.cc
Vihanakangas: Done.

src/buffer_test.cc
Vihanakangas: Done.

src/buffer_test.cc
```diff
@@ -124,4 +125,173 @@ TEST_F(BufferTest, SizeMatrixPaddedStd430) {
...
+  Buffer b2(BufferType::kColor);
+  b2.SetFormat(&fmt);
+  b2.SetData(values2);
+
+  EXPECT_FALSE(b1.CompareHistogramEMD(&b2, 0.001f).IsSuccess());
+}
+
+// Creates two sets of equal pixel values, except for one pixel that has +50 in
+// its red channel. Compares the histograms to see if they are equal with a high
+// threshold, which we expect to success.
```
paulthomson: typo: succeed.

src/buffer_test.cc
```diff
@@ -124,4 +125,173 @@ TEST_F(BufferTest, SizeMatrixPaddedStd430) {
...
+
+  for (uint8_t i = 0; i < 4; i++) {
+    std::vector<uint64_t> bins = b.GetHistogramForChannel(i, 256);
+    for (uint32_t y = 0; y < values.size(); y++)
+      EXPECT_EQ(10, bins[0]);
+  }
+}
+
+// Creates 10 RGBA pixel values, with all channels being the maximum value of 8
+// bit uint, and gets the histogram.
```
paulthomson: "gets the histogram"

Not a big deal, but since there is typo below that is worth fixing:

I would replace "gets the histogram" with "checks that..." for all of the above tests. E.g. in this case: checks that all channels have a count of 10 (all pixels) in the 255 (max uint8_t) bin.

src/buffer_test.cc
Vihanakangas: Done.

src/buffer_test.cc
Vihanakangas: Done. Hopefully it's correct.

paulthomson:
@Vihanakangas Please rebase

paulthomson:
Looks good to me. @dj2 can we merge?

dj2:
Go for it.

Dan

Issue #706, opened 2019-11-07
APK that mirrors samples/amber.cc


paulthomson:
We can build an `amber_ndk` executable that runs on Android from the ADB shell. We can also build an APK that bundles tests.

We would like an APK that is similar to `samples/amber.cc` and `amber_ndk`. I.e. it can run arbitrary tests. Ideally, we would also like to get the Amber log output and know whether the process (i.e. the graphics driver) crashed.

paulthomson:
These changes look great! And I have tried Ari's APK (not available on GitHub yet), which works really well. Here is my feedback.

* Write stdout and stderr to different files. Update example script accordingly.
* Related: I think we should have some more `-e` arguments to specify the paths for the stdout and stderr files. E.g. `-e stdout /sdcard/stdout.txt`, and this can be passed down to the C++ as another String argument. If they are not set, then they can default to e.g. /sdcard/stdout.txt.
* Try removing `AmberTestRunner` and use `InstrumentationRegistry.getArguments()` to get arguments.
* Fix the script to use this: `adb shell "am instrument -w -e args \"$*\" com.google.amber.test/androidx.test.runner.AndroidJUnitRunner"`
* Don’t add `-d` by default; I *think* we could add the validation layers to the APK, and specifying -d manually should work with the above fix to the script.
* Strings need to be free’d in JNI code. Probably not a big deal since the process will end anyway, but we may as well get it right. Here is an example I have.

```
std::string path_string;
  {
    const char* path_cstr = env->GetStringUTFChars(path_jstring, nullptr);
    path_string = path_cstr;
    env->ReleaseStringUTFChars(path_jstring, path_cstr);
  }
```

* Rename `useAppContext` and update comments. In fact, maybe create a new app in Android Studio based on the Empty Activity app (or whatever gets us an even simpler app, without the `amber/app/src/main/res/layout/activity_main.xml` UI file).
* Optional: we could try to support arguments with spaces by doing: `-e 1 arg1 -e 2 arg2 ...` and then in the Java code, checking for keys 1, 2, 3, etc., adding them to a `String[]` variable, and then passing them down to the C++ code in a `String[]` parameter. But I don't think this is particularly important.

paulthomson:
Nice work! I missed some details, so apologies that some of my feedback will be filling in those gaps.

[...]

PR #710, opened 2019-11-15
Don't add -d by default. Parse options for stdout/stderr files.

asuonpaa:
Fixes https://github.com/google/amber/issues/706


samples/android_main.cc
```diff
@@ -24,22 +24,44 @@ extern "C" JNIEXPORT JNICALL int Java_com_google_amber_Amber_androidMain(
...
-  for (auto s : argv_string)
-    argv.push_back(s.c_str());
+  std::string stdout_file = "/sdcard/amber_stdout.txt";
+  std::string stderr_file = "/sdcard/amber_stderr.txt";
+
+  for (size_t i = 0; i < argv_string.size(); i++) {
+    auto s = argv_string[i];
+
+    if (s == "--stdout") {
+      i++;
```
dj2: Need to validate i is < argv_string.size()

samples/android_main.cc
```diff
@@ -24,22 +24,44 @@ extern "C" JNIEXPORT JNICALL int Java_com_google_amber_Amber_androidMain(
...
+      stdout_file = argv_string[i];
+      continue;
+    }
+
+    if (s == "--stderr") {
+      i++;
```
dj2: ditto

samples/android_main.cc
```diff
@@ -23,23 +23,34 @@ extern int main(int argc, const char** argv);
...
   // Redirect std output to a file
-  freopen("/sdcard/amberlog.txt", "w", stdout);
-  freopen("/sdcard/amberlog.txt", "a", stderr);
+  freopen(stdout_file_cstr, "w", stdout);
+  freopen(stderr_file_cstr, "w", stderr);
+
+  env->ReleaseStringUTFChars(stdoutFile, stdout_file_cstr);
+  env->ReleaseStringUTFChars(stderrFile, stderr_file_cstr);
+
+  int arg_count = env->GetArrayLength(args);
```
paulthomson: Nit but could you use `jsize` here and below.

samples/android_main.cc
```diff
@@ -23,23 +23,34 @@ extern int main(int argc, const char** argv);
...
+    argv_string.push_back(arg_cstr);
+    env->ReleaseStringUTFChars(js, arg_cstr);
+  }
 
-  // Parse argument string and add -d by default
-  std::stringstream ss(args);
-  std::vector<std::string> argv_string{std::istream_iterator<std::string>{ss},
-                                       std::istream_iterator<std::string>{}};
   std::vector<const char*> argv;
   argv.push_back("amber");
```
paulthomson: Nit but could you push "amber" into `argv_string` above (before the loop).

samples/android_main.cc
```diff
@@ -23,23 +23,34 @@ extern int main(int argc, const char** argv);
...
-  argv.push_back("-d");
 
-  for (auto s : argv_string)
-    argv.push_back(s.c_str());
+  for (size_t i = 0; i < argv_string.size(); i++)
```
paulthomson: Nit but could you use a `for(const std::string& arg : argv_string)` loop instead?

asuonpaa: Sure

samples/android_main.cc
asuonpaa: done

samples/android_main.cc
asuonpaa: done

Issue #711, opened 2019-11-19
Add support for combined image sampler


asuonpaa:
Now that we have sampled image and sampler support, it should be fairly easy to add combined image samplers.

PR #713, opened 2019-11-20
Added combined image sampler support

asuonpaa:
Fixes https://github.com/google/amber/issues/711

### DIFF
--- a/docs/amber_script.md
+++ b/docs/amber_script.md
@@ -270,6 +270,11 @@ contain image attachment content, depth/stencil content, uniform buffers, etc.
   # be a `FORMAT` buffer.
   BIND BUFFER {buffer_name} AS sampled_image
 
+  # Attach |buffer_name| as a combined image image sampler. The provided
+  # buffer must be a `FORMAT` buffer. A sampler |sampler_name| must also
+  # be specified.
+  BIND BUFFER {buffer_name} AS combined_image_sampler SAMPLER {sampler_name}
+
   # Bind the sampler at the given descriptor set and binding.
   BIND SAMPLER {sampler_name} DESCRIPTOR_SET _id_ BINDING _id_
 
--- a/docs/amber_script.md
+++ b/docs/amber_script.md
@@ -262,17 +262,14 @@ contain image attachment content, depth/stencil content, uniform buffers, etc.
   BIND BUFFER {buffer_name} AS {buffer_type} DESCRIPTOR_SET _id_ \
        BINDING _id_
 
-  # Attach |buffer_name| as a storage image. The provided buffer must
-  # be a `FORMAT` buffer.
+  # Attach |buffer_name| as a storage image.
   BIND BUFFER {buffer_name} AS storage_image
 
-  # Attach |buffer_name| as a sampled image. The provided buffer must
-  # be a `FORMAT` buffer.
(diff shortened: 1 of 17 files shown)


docs/amber_script.md
```diff
@@ -270,6 +270,11 @@ contain image attachment content, depth/stencil content, uniform buffers, etc.
   # be a `FORMAT` buffer.
   BIND BUFFER {buffer_name} AS sampled_image
 
+  # Attach |buffer_name| as a combined image image sampler. The provided
+  # buffer must be a `FORMAT` buffer. A sampler |sampler_name| must also
```
dj2: I don't think this needs to restrict to FORMAT buffers. We can assemble a format name for things like a vec2 or vec3. You'll just end up with R32G32B32_SFLOAT. If you want something more specific, you can use FORMAT thought.

src/vulkan/image_descriptor.h
```diff
@@ -44,6 +45,8 @@ class ImageDescriptor : public BufferBackedDescriptor {
 
  private:
   std::unique_ptr<TransferImage> transfer_image_;
+  Sampler* amber_sampler_;
```
dj2: nit: move the null ptr here and remove the default value.

src/vulkan/pipeline.cc
```diff
@@ -279,7 +279,7 @@ Result Pipeline::AddBufferDescriptor(const BufferCommand* cmd) {
   if (cmd->IsPushConstant())
     return AddPushConstantBuffer(cmd->GetBuffer(), cmd->GetOffset());
   if (!cmd->IsSSBO() && !cmd->IsUniform() && !cmd->IsStorageImage() &&
-      !cmd->IsSampledImage())
+      !cmd->IsSampledImage() && !cmd->IsCombinedImageSampler())
```
dj2: nit: add {}'s

src/vulkan/sampler_descriptor.h
```diff
@@ -21,6 +21,8 @@
 namespace amber {
 namespace vulkan {
 
+VkSamplerCreateInfo GetSamplerCreateInfo(Sampler* sampler);
```
dj2: What about creating a amber::vulkan::Sampler class. It would have a `static VkSamplerCreateInfo GetSamplerCreateInfo(Sampler* sampler)` and then a the init and destructor methods to create and destroy the sampler. Then the image descriptor above would hold a vulkan::Sampler and not need to have the destructor code added?

docs/amber_script.md
asuonpaa: Done

src/vulkan/image_descriptor.h
asuonpaa: Ok, I've also added a setter for this so I don't have to specify the pointer for all image descriptor objects.

src/vulkan/pipeline.cc
asuonpaa: Done

src/vulkan/sampler_descriptor.h
asuonpaa: Done

dj2:
The Sampler class cleanup looks great, thanks for doing that.

The dawn error looks unrelated, once the format error is fixed this should be good to go.

PR #717, opened 2019-11-22
Android Gradle project files

asuonpaa:
Part of https://github.com/google/amber/issues/706


android_gradle/amber.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
dj2: Copyright missing

android_gradle/app/src/main/res/drawable-v24/ic_launcher_foreground.xml
```diff
@@ -0,0 +1,30 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:aapt="http://schemas.android.com/aapt"
+    android:width="108dp"
+    android:height="108dp"
+    android:viewportWidth="108"
+    android:viewportHeight="108">
+    <path android:pathData="M31,63.928c0,0 6.4,-11 12.1,-13.1c7.2,-2.6 26,-1.4 26,-1.4l38.1,38.1L107,108.928l-32,-1L31,63.928z">
```
dj2: Out of curiosity, what do the paths draw?

android_gradle/app/src/main/res/mipmap-anydpi-v26/ic_launcher_round.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dj2: Where do all the icon files come from? Do we need to add them? Should we just create a circle with an A in them so the copyright is amber authors?

android_gradle/amber.sh
asuonpaa: Added.

android_gradle/app/src/main/res/mipmap-anydpi-v26/ic_launcher_round.xml
asuonpaa: Those are default icons. It seems to work without, so I've removed all the resource files.

PR #724, opened 2019-11-27
Add ability to extract shader source.

dj2:
If the shader is compiled by Amber, the compiled shader data will now be
returned when querying for shader information. This allows adding a flag
'-w <filename>' to the amber sample program to dump the SPIR-V
disassembly of the compiled shaders to a file.

This also fixes up the shader name and optimization pass return values
in the shader_info to hold the correct information based on pipeline.

### DIFF
--- a/samples/amber.cc
+++ b/samples/amber.cc
@@ -330,7 +330,8 @@ class SampleDelegate : public amber::Delegate {
   bool log_execute_calls_ = false;
 };
 
-std::string disassemble(const std::string& env, const std::vector<uint32_t>& data) {
+std::string disassemble(const std::string& env,
+                        const std::vector<uint32_t>& data) {
 #if AMBER_ENABLE_SPIRV_TOOLS
   std::string spv_errors;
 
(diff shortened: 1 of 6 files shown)


samples/amber.cc
```diff
@@ -480,6 +540,29 @@ int main(int argc, const char** argv) {
       // give clues as to the failure.
     }
 
+    // Dump the shader assembly
+    if (!options.shader_filename.empty()) {
+#if AMBER_ENABLE_SPIRV_TOOLS
```
alan-baker: Makes me wonder whether you should just write the binary by default and have a second argument to control disassembly.

dj2: Writing binary is tricker as I have to write to N files. Currently I write all shaders to a single file ....

alan-baker: Right, the only other suggestion I have would be that -w specified a base file name and you output: <base>-<shader name>.[spv/svpasm]. At the same time this isn't a huge deal so feel free not to do it.

dj2: Right, but in the .spv case, you'd have to write out N files because you can have multiple shader,pipeline pairs. Right now, it just writes them all into a single file.

It's doable, I'm just not sure if it's worth it. Is the non-asm version of this _useful_?

alan-baker: Probably not enough to warrant the headaches.

PR #728, opened 2019-11-27
Remove BufferType from the Buffer class.

dj2:
The BufferType is a product of the <Pipeline, Buffer> pair. A given
buffer could be used as a colour buffer on one pipeline and as an image
on the second.

This Cl removes the BufferType from the Buffer and always uses the
BufferType on the Pipeline BufferInfo.


dj2:
Yup, all tests passed. CLSPV tests passed as well.

PR #741, opened 2019-12-05
Prefix pipeline name to shader name when searching shader map.

asuonpaa:
CTS amber integration was broken because of recent changes where shader names got pipeline name as prefix. Now the shader map is searched with that same prefix.

### DIFF
--- a/src/shader_compiler.cc
+++ b/src/shader_compiler.cc
@@ -58,7 +58,11 @@ std::pair<Result, std::vector<uint32_t>> ShaderCompiler::Compile(
     Pipeline::ShaderInfo* shader_info,
     const ShaderMap& shader_map) const {
   const auto shader = shader_info->GetShader();
-  auto it = shader_map.find(pipeline_name + "-" + shader->GetName());
+  std::string key = shader->GetName();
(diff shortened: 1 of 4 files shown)


src/shader_compiler.cc
```diff
@@ -54,10 +54,11 @@ ShaderCompiler::ShaderCompiler(const std::string& env,
 ShaderCompiler::~ShaderCompiler() = default;
 
 std::pair<Result, std::vector<uint32_t>> ShaderCompiler::Compile(
+    std::string pipeline_name,
     Pipeline::ShaderInfo* shader_info,
     const ShaderMap& shader_map) const {
   const auto shader = shader_info->GetShader();
-  auto it = shader_map.find(shader->GetName());
+  auto it = shader_map.find(pipeline_name + "-" + shader->GetName());
```
dj2: For the case where the shader is not in a pipeline, the "-" should not be present I think. It should never happen here as we call this with a pipeline all the time, but we should probably check if pipeline_name is "" to not add the "-".

asuonpaa: Done

PR #749, opened 2019-12-10
Don't checkout clspv and dxc deps by default

### LINKED ISSUE #707: Pull in only selected external dependencies unless otherwise specified
Some of the dependencies Amber pulls in by default - including clspv and SwiftShader - have large repositories but are not required by default.

Changing the setup so that these would be excluded if requested would help reduce the disk footprint of Amber, and speed up CI somewhat.

@dj2

alan-baker:
Fixes #707

* Add new arguments to git-sync-deps to control whether dxc and clspv
are checked out
* Update linux build to be able to sync extra deps

### DIFF
--- a/tools/git-sync-deps
+++ b/tools/git-sync-deps
@@ -50,6 +50,7 @@ from builtins import bytes
 
 with_clspv = False
 with_dxc = False
+with_swiftshader = False
 
 def git_executable():
   """Find the git executable.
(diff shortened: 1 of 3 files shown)


tools/git-sync-deps
```diff
@@ -37,14 +41,15 @@ Git Config:
...
-
 import os
 import re
 import subprocess
 import sys
 import threading
 from builtins import bytes
 
+with_clspv = False
+with_dxc = False
```
dj2: Did you skip the swiftshader bits here and below on purpose?

alan-baker: Nope adding now.

PR #750, opened 2019-12-10
Simplify amber_script spec document

dj2:
This CL combines some of the repeated entries in the spec and displays
all of the optional parameters together along with default values.


dj2:
@dneto0 @alan-baker does this seem clearer?

PR #751, opened 2019-12-10
Add Float 16 Support

### LINKED ISSUE #485: support 16bit float, a.k.a. half type
The tricky part of this is that C/C++ don't have native support for it.
There are various headers-only libraries for maniuplating them, including one in SPIRV-Tools that awoloszyn wrote

dj2:
This CL adds the `float16` data type into amber_script and the required plumbing to enable the feature.

Fixes #485

### DIFF
--- a/samples/config_helper_vulkan.cc
+++ b/samples/config_helper_vulkan.cc
@@ -46,6 +46,7 @@ const size_t kNumberOfRequiredValidationLayers =
 const char kVariablePointers[] = "VariablePointerFeatures.variablePointers";
 const char kVariablePointersStorageBuffer[] =
     "VariablePointerFeatures.variablePointersStorageBuffer";
+const char kShaderFloat16[] = "Float16Int8Features.shaderFloat16";
 
 const char kExtensionForValidationLayer[] = "VK_EXT_debug_report";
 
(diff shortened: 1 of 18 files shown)


samples/config_helper_vulkan.cc
```diff
@@ -46,6 +46,7 @@ const size_t kNumberOfRequiredValidationLayers =
 const char kVariablePointers[] = "VariablePointerFeatures.variablePointers";
 const char kVariablePointersStorageBuffer[] =
     "VariablePointerFeatures.variablePointersStorageBuffer";
+const char kFloat16Int8[] = "Float16Int8Features.shaderFloat16";
```
alan-baker: This should probably be kFloat16Int8Float16 or something equally ridiculous if you want to eventually support the int8 feature bit.

src/float16_helper.cc
```diff
@@ -0,0 +1,127 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "src/float16_helper.h"
+
+#include <cassert>
+#include <iostream>
```
alan-baker: Left over debugging?

samples/config_helper_vulkan.cc
dj2: Done.

src/float16_helper.cc
dj2: done

Issue #747, opened 2019-12-09
Support more image types


alan-baker:
It appears Amber only supports 2D images currently. It would be nice to support:
* 1D
* 3D
* Cube
* Arrayed

Amber probably needs a way to declare a buffer as a specific image type.

alan-baker:
Amberscript additions:
* add IMAGE command that takes image parameters, but creates a buffer

```
IMAGE <name> (1D|2D|3D|Cube) [ARRAYED] [MULTISAMPLED] \
  WIDTH <w> [HEIGHT <h> [DEPTH <d>]] [ARRAY_LAYERS <a>] \
  <buffer initializers>
```

Names subject to change, but this new command allows for simple parsing and image creation on the vulkan side. It is convenient to reuse buffer from the internal side since they are not fundamentally different.

PR #757, opened 2019-12-16
Image buffers

alan-baker:
Contributes to #747 

Add support for specifying buffers as images via a new command.
 * takes dimensionality and dimensions

Generalize image support to support 1D and 3D images.

Add some examples of the new functionality.


docs/amber_script.md
```diff
@@ -137,18 +137,38 @@ END
 BUFFER {name} DATA_TYPE {type} {STD140 | STD430} SIZE _size_in_items_ \
     {initializer}
 
-# Defines a buffer with width and height and filled by data as specified by the
```
dj2: Thinking about this, I think we should leave this here and just add a # Deprecated use IMAGE tag

docs/amber_script.md
```diff
@@ -137,18 +137,38 @@ END
...
+attributes.
+
+##### Dimensionality
+ * `DIM_1D` -- A 1-dimensional image
+ * `DIM_2D` -- A 2-dimensional image
+ * `DIM_3D` -- A 3-dimensional image
+
+```groovy
+# Specify an image buffer with a format. HEIGHT is necessary for DIM_2D and
+DIM_3D. DEPTH is necessary for DIM_3D.
```
dj2: nit: add # at start of line

docs/amber_script.md
```diff
@@ -137,18 +137,38 @@ END
...
+IMAGE {name} FORMAT {format_string} [ MIP_LEVELS _mip_levels_ (default 1) ] \
+    {dimensionality} \
+    WIDTH {w} [ HEIGHT {h} [ DEPTH {d} ] ] \
+    {initializer}
+
+# Specify an image buffer with a data type. HEIGHT is necessary for DIM_2D and
+DIM_3D. DEPTH is necessary for DIM_3D.
```
dj2: ditto

docs/amber_script.md
```diff
@@ -137,18 +137,38 @@ END
...
+IMAGE {name} DATA_TYPE {type} {dimensionality} \
```
dj2: Does this accept MIP_LEVELS?

src/amberscript/parser.cc
```diff
@@ -1205,6 +1219,151 @@ Result Parser::ParseBuffer() {
...
+
+  auto name = token->AsString();
+  if (name == "DATA_TYPE" || name == "FORMAT")
+    return Result("missing IMAGE name");
+
+  token = tokenizer_->NextToken();
+  if (!token->IsString())
+    return Result("invalid IMAGE command provided");
+
+  std::unique_ptr<Buffer> buffer;
```
dj2: Could probably just do:

```
auto buffer = MakeUnique<Buffer>();
```

we'd waste the allocation in case of error, but otherwise we always make the buffer this way.

src/amberscript/parser.cc
```diff
@@ -1205,6 +1219,151 @@ Result Parser::ParseBuffer() {
...
+      fmt = MakeUnique<Format>(type);
+      buffer->SetFormat(fmt.get());
+    } else {
+      auto new_type = ToType(token->AsString());
+      if (!new_type)
+        return Result("invalid data type '" + token->AsString() + "' provided");
+
+      fmt = MakeUnique<Format>(new_type.get());
+      buffer->SetFormat(fmt.get());
+      type = new_type.get();
```
dj2: Never used?

src/amberscript/parser.cc
```diff
@@ -1205,6 +1219,151 @@ Result Parser::ParseBuffer() {
...
+
+      if (!token->IsInteger())
+        return Result("invalid value for MIP_LEVELS");
+
+      buffer->SetMipLevels(token->AsUint32());
+    }
+  } else {
+    return Result("unknown IMAGE command provided: " + cmd);
+  }
+  buffer->SetName(name);
```
dj2: This could then move up with the buffer creation

docs/amber_script.md
alan-baker: Not currently.

PR #759, opened 2019-12-17
Auto-generate OpenCL literal samplers

alan-baker:
* Parse literal sampler descriptor map entries
  * add them to the pipeline sampler info
* Before generating engine pipelines, create the sampler objects for the
literal samplers
* Tests


src/pipeline.cc
```diff
@@ -723,4 +739,66 @@ Result Pipeline::GenerateOpenCLPodBuffers() {
...
+                             "." + std::to_string(info.binding));
+
+    // The values for addressing modes, filtering modes and coordinate
+    // normalization are all defined in the OpenCL header.
+    //
+    // LSB is coordinates mask.
+    literal_sampler->SetNormalizedCoords(info.mask & 0x1);
+
+    // Next 3 bits are addressing mode.
+    uint32_t addressing_bits = info.mask & 0xe;
```
dj2: Can we add named defines for these various magic masks?

src/pipeline.cc
```diff
@@ -723,4 +739,66 @@ Result Pipeline::GenerateOpenCLPodBuffers() {
...
+    AddressMode addressing_mode = AddressMode::kUnknown;
+    if (addressing_bits == 0 || addressing_bits == 2) {
```
dj2: Do these values have names in opencl that we can use here?

Issue #763, opened 2020-01-10
SwiftShader change tracker bug


ben-clayton:
This is a tracker for things that have changed in SwiftShader since the last time Amber updated, which need to be considered for the next update:

* SwiftShader has moved to C++14. [This CMake conditional](https://github.com/google/amber/blob/master/CMakeLists.txt#L31) will likely need to also consider `AMBER_ENABLE_SWIFTSHADER`.
* Several SwiftShader CMake options have been renamed. [This CL](https://swiftshader-review.googlesource.com/c/SwiftShader/+/39909/6/CMakeLists.txt) prefixes a number of options with `SWIFTSHADER_`. [This CMake logic](https://github.com/google/amber/blob/5acb622459aa850d63f345ab4b8220468dce035b/third_party/CMakeLists.txt#L68-L77) will have to be updated.

PR #768, opened 2020-01-17
Roll SwiftShader forward

ben-clayton:
Repoint SwiftShader to `2cb1db08`.

Update CMake build flags after a renaming spree.

SwiftShader has also moved to C++14, so set `CMAKE_CXX_STANDARD` to 14
when `AMBER_ENABLE_SWIFTSHADER` is enabled.

Fixes: #763


kokoro/scripts/linux/build.sh
```diff
@@ -31,8 +31,13 @@ if [ $COMPILER = "clang" ]
...
 else
-  # Use newer gcc than default.
-  CMAKE_C_CXX_COMPILER="-DCMAKE_C_COMPILER=/usr/bin/gcc-5 -DCMAKE_CXX_COMPILER=/usr/bin/g++-5"
+  # Specify we want to build with GCC 7 (which supports C++14)
+  sudo add-apt-repository ppa:ubuntu-toolchain-r/test
+  sudo apt-get update
+  sudo apt-get install -y gcc-7 g++-7
+  sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-7 100 --slave /usr/bin/g++ g++ /usr/bin/g++-7
+  sudo update-alternatives --set gcc "/usr/bin/gcc-7"
+  CMAKE_C_CXX_COMPILER="-DCMAKE_C_COMPILER=/usr/bin/gcc-7 -DCMAKE_CXX_COMPILER=/usr/bin/g++-7"
```
dj2: Oh, does this mean I can turn swiftshader on for the ubuntu tests now? I couldn't get it to compile before.

ben-clayton: I'm confused. I only made this change to get SwiftShader building.
Is there another kokoro config I'm not aware of?

dj2: There is nothing else you need to do. This is the build file for Linux (there is also one for osx and windows). Each kokoro bot calls this file with given flags, the ubuntu bot disables swiftshader as it wouldn't link for me (due to gcc5 I believe). So, I think with this gcc7 change it should work. I'll give it a try on Monday.

ben-clayton: I only made this change because the GCC Ubuntu builds started failing after rolling forward Swiftshader.
I can only conclude kokoro was already building Swiftshader, on Linux, using gcc5.

PR #774, opened 2020-01-29
tokenizer: Tokenize quoted strings

ben-clayton:
Handles the basic set of escape sequences: `'\a'`, `'\b'`, `'\t'`, `'\n'`, `'\v'`, `'\f'`, `'\r'`.
Also supports multi-line strings.

Added tests.

### DIFF
--- a/src/tokenizer.cc
+++ b/src/tokenizer.cc
@@ -31,7 +31,7 @@ Result Token::ConvertToDouble() {
   if (IsDouble())
     return {};
 
-  if (IsString() || IsEOL() || IsEOS())
+  if (IsIdentifier() || IsEOL() || IsEOS())
     return Result("Invalid conversion to double");
 
   if (IsInteger()) {
@@ -78,7 +78,7 @@ std::unique_ptr<Token> Tokenizer::NextToken() {
   // want to consume any other characters.
   if (data_[current_position_] == ',' || data_[current_position_] == '(' ||
       data_[current_position_] == ')') {
-    auto tok = MakeUnique<Token>(TokenType::kString);
+    auto tok = MakeUnique<Token>(TokenType::kIdentifier);
     std::string str(1, data_[current_position_]);
     tok->SetStringValue(str);
     ++current_position_;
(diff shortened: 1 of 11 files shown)


ben-clayton:
Note: I'm not sure about folding this into the existing `TokenType::kString` type. I would normally classify `kString` as `kIdentifier` as these are not quoted.
If we feel like it is worthwhile separating these two token types (rename the old 'string' type into 'identifier', add quoted strings as 'string'), then also I'm happy to do this.

src/tokenizer.cc
```diff
@@ -74,6 +74,86 @@ std::unique_ptr<Token> Tokenizer::NextToken() {
...
+    std::string tok_str;
+    bool escape = false;
+    for (; current_position_ < data_.length(); current_position_++) {
+      auto c = data_[current_position_];
+      switch (c) {
+        case '\\':
+          if (!escape) {
+            escape = true;
+            continue;
+          }
```
dj2: Does this handle a quoted \ correctly? If I give the input string of "\\n" this will get output as "\n" won't it? The first \ will mark as as escaped, the second \ will be output directly without escaping?

src/tokenizer_test.cc
```diff
@@ -298,6 +298,56 @@ TEST_F(TokenizerTest, StringStartingWithNum) {
...
+  EXPECT_TRUE(next->IsString());
+  EXPECT_EQ("Hello\n\nworld", next->AsString());
+
+  next = t.NextToken();
+  ASSERT_TRUE(next != nullptr);
+  EXPECT_TRUE(next->IsEOS());
+}
+
+TEST_F(TokenizerTest, StringQuotedEscapeSequences) {
+  Tokenizer t(R"("_\\_a\aa_b\bb_t\tt_n\nn_v\vv_f\ff_r\rr_")");
```
dj2: Throw a \" into this one to make sure embedded quotes work?

src/tokenizer.cc
ben-clayton: I suspect github may have mangled your text - that or I'm just being thick.

If the question was `Does this handle a quoted \\ correctly?` I believe so. `\\` is the escape sequence for writing a backslash. So `\\n` is `\n` (no escape).

src/tokenizer_test.cc
ben-clayton: Done. Split into two changes for sanity's sake.

PR #776, opened 2020-01-29
Add debugger testing support

ben-clayton:
An overview of the functionality and design can be found in the `docs/debugger.md` file, which is part of this change.

Note that this is all rather WIP, but this PR gives us something that we can start playing with.


ben-clayton:
Looking into presubmit failures now.

third_party/CMakeLists.txt
```diff
@@ -65,14 +65,25 @@ if (${AMBER_USE_LOCAL_VULKAN})
...
+
 if (${AMBER_ENABLE_SWIFTSHADER})
-  set(SWIFTSHADER_BUILD_EGL FALSE CACHE BOOL FALSE)
-  set(SWIFTSHADER_BUILD_GLESv2 FALSE CACHE BOOL FALSE)
-  set(SWIFTSHADER_BUILD_GLES_CM FALSE CACHE BOOL FALSE)
-  set(SWIFTSHADER_BUILD_VULKAN TRUE CACHE BOOL TRUE)
-  set(SWIFTSHADER_BUILD_SAMPLES FALSE CACHE BOOL FALSE)
-  set(SWIFTSHADER_BUILD_TESTS FALSE CACHE BOOL FALSE)
-  set(SWIFTSHADER_WARNINGS_AS_ERRORS FALSE CACHE BOOL FALSE)
+  set(SWIFTSHADER_BUILD_EGL FALSE)
```
dj2: Why remove the CACHE BOOL. Those seem to be needed to keep cmake happy. If you don't cache it then option overrides it, or something wonky.

src/vulkan/engine_vulkan_debugger.cc
```diff
@@ -0,0 +1,976 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "src/vulkan/engine_vulkan.h"
+
+#if AMBER_ENABLE_VK_DEBUGGING
```
dj2: I don't think this is needed. We should just not add this file in the cmakelist instead of if'defing the whole file.

src/vulkan/engine_vulkan_debugger.cc
```diff
@@ -0,0 +1,976 @@
...
+    printf(__VA_ARGS__);  \
+    printf("\n");         \
+  } while (false)
+#else
+#define DEBUGGER_LOG(...)
+#endif
+
+namespace {
+
+// TODO(bclayton): Make script configurable?
```
dj2: I think this is fine for now. File a bug so we can see if anyone pings it in the future?

src/vulkan/engine_vulkan_debugger.cc
```diff
@@ -0,0 +1,976 @@
...
+
+ private:
+  std::condition_variable cv_;
+  std::mutex mutex_;
+  bool signalled_ = false;
+};
+
+// Split slices str into all substrings separated by sep and returns a vector of
+// the substrings between those separators.
+// TODO(bclayton): Move to string utility file?
```
dj2: My usual thinking here is, if we have the same thing somewhere else, make ti common, otherwise leave it here for now and we can split it out later.

src/vulkan/engine_vulkan_debugger.cc
```diff
@@ -0,0 +1,976 @@
...
+    return true;
+  }
+
+  template <typename T>
+  bool Get(std::tuple<T, T, T>* out) const {
+    auto x = children.Find("x");
+    auto y = children.Find("y");
+    auto z = children.Find("z");
+    if (x != nullptr && y != nullptr && z != nullptr) {
+      T elX, elY, elZ;
```
dj2: nit: 1 declaration per line

src/vulkan/engine_vulkan_debugger.cc
```diff
@@ -0,0 +1,976 @@
...
+    return true;
+  }
+
+  // SourceContext retrieves the the SourceLines for the given source.
+  // Returns true on success, false on error.
+  bool SourceContent(const dap::Source& source, SourceLines* out) {
+    auto path = source.path.value("");
+    if (path != "") {
+      auto it = sourceCache_.by_path.find(path);
+      if (it != sourceCache_.by_path.end()) {
```
dj2: So, amber goes through here all the time as our files are embedded inside our scirpt, or does the DAP write temporary files out?

src/vulkan/engine_vulkan_debugger.cc
```diff
@@ -0,0 +1,976 @@
...
+        *out = it->second;
+        return true;
+      }
+
+      std::ifstream file(path);
```
dj2: This should, probably, call back to the embedder to load the file. The amber library doesn't know how to read files anywhere else, it assumes the embedder will handle that for it.

src/vulkan/engine_vulkan_debugger.cc
```diff
@@ -0,0 +1,976 @@
...
+
+// GlobalInvocationId holds a three-element unsigned integer index, used to
+// identifiy a single compute invocation.
+struct GlobalInvocationId {
+  size_t hash() const { return x << 20 | y << 10 | z; }
+  bool operator==(const GlobalInvocationId& other) const {
+    return x == other.x && y == other.y && z == other.z;
+  }
+
+  uint32_t x, y, z;
```
dj2: nit: one per line

src/vulkan/engine_vulkan_debugger.cc
```diff
@@ -0,0 +1,976 @@
...
+
+  // String returns a human-readable description of the key.
+  std::string String() const;
+
+  Type type;
+  Data data;
+};
+
+size_t InvocationKey::Hash::operator()(const InvocationKey& key) const {
+  size_t hash = 31 * static_cast<size_t>(key.type);
```
dj2: Should the enum have values set explicitly? Otherwise, couldn't' the key.type end up being up to the compiler and max_int or something which will overflow the size_t when multiplied?

src/vulkan/engine_vulkan_debugger.cc
```diff
@@ -0,0 +1,976 @@
...
+
+  const dap::integer threadId_;
+  const int lane_;
+  Client client_;
+  std::thread thread_;
+  Event done_;
+  Result error_;
+};
+
+}  // namespace
```
dj2: Move down to 664 so anonymous namespace is inside amber vulkan namespace.

src/vulkan/engine_vulkan_debugger.cc
```diff
@@ -0,0 +1,976 @@
...
+
+    for (int i = 0;; i++) {
+      auto lane_var = client.GetLane(locals, i);
+      if (!lane_var) {
+        break;
+      }
+      if (auto var = lane_var->Find(kGlobalInvocationId)) {
+        std::tuple<uint32_t, uint32_t, uint32_t> got;
+        if (var->Get(&got)) {
+          uint32_t x, y, z;
```
dj2: nit: one per line

src/vulkan/engine_vulkan_debugger.cc
```diff
@@ -0,0 +1,976 @@
...
+  std::mutex threads_mutex_;
+  PendingThreadsMap pendingThreads_;  // guarded by threads_mutex_
+  ThreadVector runningThreads_;       // guarded by threads_mutex_
+  std::mutex error_mutex_;
+  Result error_;  // guarded by error_mutex_
+};
+
+Result EngineVulkan::GetDebugger(Debugger** out) {
+  if (!debugger_) {
+    debugger_ = new VkDebugger();
```
dj2: Can debugger_ be a unique_ptr? We've been trying to avoid new/delete in amber.

src/vulkan/engine_vulkan_debugger.cc
```diff
@@ -0,0 +1,976 @@
...
+    auto res = debugger_->Connect();
+    if (!res.IsSuccess()) {
+      return res;
+    }
+  }
+  *out = debugger_;
+  return Result();
+}
+
+void EngineVulkan::DeleteDebugger() {
```
dj2: Which I think then means this isn't needed anymore as debugger will get cleaned up when the engine is destroyed

docs/debugger.md
```diff
@@ -0,0 +1,174 @@
...
+* `OpenCL.DebugInfo.100` insstructions are not currently preserved by many of the [SPIR-V Tools](https://github.com/KhronosGroup/SPIRV-Tools) optimization passes, so these optimizations should not currently be used.
+* `OpenCL.DebugInfo.100` may be incorrectly interpreted by the [SPIR-V Tools](https://github.com/KhronosGroup/SPIRV-Tools) validator, so Amber should currently be invoked with the `--disable-spirv-val` flag.
+
+---
+
+## Usage
+
+### Building
+
+The debugger testing functionality is disabled by default, and has to be enabled with the `AMBER_ENABLE_SWIFTSHADER` CMake flag.
```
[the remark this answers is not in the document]
ben-clayton: Yes, good spot. Done

docs/debugger.md
```diff
@@ -0,0 +1,174 @@
...
+
+  Declares a sequence of debugger commands to run when the shader invocation for the vertex with the given index is executed.
+
+* `THREAD FRAGMENT_WINDOW_SPACE_POSITION` \<x\> \<y\>
+
+  Defines a sequence of debugger commands to run when the shader invocation for the fragment with the given window space position is executed.
+
+Each of the `THREAD` commands begins a block that must be terminated with an `END`.
+
+Within each `THREAD` command blocks, you may use any of the following commands to control execution of the shader, and to verify the debugger's behaviour:
```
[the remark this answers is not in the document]
ben-clayton: Done

docs/debugger.md
```diff
@@ -0,0 +1,174 @@
...
+        EXPECT LOCATION "simple_vs.hlsl" 9 "  VS_OUTPUT vout;"
+        EXPECT LOCAL "pos.x" EQ -1.007874
+        EXPECT LOCAL "pos.y" EQ 1.000000
+        EXPECT LOCAL "pos.z" EQ 0.000000
+        EXPECT LOCAL "color.x" EQ 1.000000
+        EXPECT LOCAL "color.y" EQ 0.000000
+        EXPECT LOCAL "color.z" EQ 0.000000
+        STEP_IN
+        EXPECT LOCATION "simple_vs.hlsl" 10 "  vout.pos = pos;"
+        STEP_IN
```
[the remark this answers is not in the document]
ben-clayton: The first set of expectations are at the function entry, which assumes the parameters have been initialized. We enter with the following line highlighted: `/*  9 */   VS_OUTPUT vout;`

We then `STEP_IN` to have the `/* 10 */   vout.pos = pos;` line highlighted (but not yet executed).

Then one more `STEP_IN` to actually perform the assignment.

Then we can check the locals with `EXPECT LOCAL`.

Make sense, or have I screwed up somewhere?

docs/debugger.md
```diff
@@ -0,0 +1,174 @@
...
+        EXPECT LOCAL "vout.pos.x" EQ -1.007874
+        EXPECT LOCAL "vout.pos.y" EQ 1.000000
+        EXPECT LOCAL "vout.pos.z" EQ 0.000000
+        EXPECT LOCATION "simple_vs.hlsl" 11 "  vout.color = color;"
+        STEP_IN
+        EXPECT LOCAL "vout.color.x" EQ 1.000000
+        EXPECT LOCAL "vout.color.y" EQ 0.000000
+        EXPECT LOCAL "vout.color.z" EQ 0.000000
+        EXPECT LOCATION "simple_vs.hlsl" 12 "  return vout;"
+        CONTINUE
```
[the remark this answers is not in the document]
ben-clayton: It's a good question. Something that occurred to me earlier today.
Currently, things will not automatically resume, and the test will fail. I'm sort of on the fence about adding magic instructions (especially if we want to test for inter-thread interactions later).
Something I'll think about, and probably base a decision on based on the tests we produce.

docs/debugger.md
```diff
@@ -0,0 +1,174 @@
...
+
+The `amber::debug::Events` interface has a number of `BreakOn`XXX`()` methods that have parameters for the shader invocation of interest (global invocation ID, vertex index, etc) and a `OnThread` function callback parameter.
+
+The `OnThread` callback has the signature `void(amber::debug::Thread*)` which is called to control the debugger thread of execution and perform test verifications.
+
+### `amber::debug::Thread`
+
+`amber::debug::Thread` is the interface used to control a debugger thread of execution, and represents the script commands within the `THREAD` script blocks.
+
+Implementations the `amber::debug::Thread` interface are returned by the `amber::debug::Script::BreakOn`XXX`()` methods:
```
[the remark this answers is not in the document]
ben-clayton: Yeah, I can't seem to write english. Fixed.

docs/debugger.md
```diff
@@ -0,0 +1,174 @@
...
+
+* `THREAD GLOBAL_INVOCATION_ID` \<x\> \<y\> \<z\>
+
+  Declares a sequence of debugger commands to run when the compute shader invocation with the given global invocation identifier is executed.
+
+* `THREAD VERTEX_INDEX` \<index\>
+
+  Declares a sequence of debugger commands to run when the shader invocation for the vertex with the given index is executed.
+
+* `THREAD FRAGMENT_WINDOW_SPACE_POSITION` \<x\> \<y\>
```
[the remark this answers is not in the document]
ben-clayton: Good spot. :) I've added a todo.

src/amberscript/parser.cc
```diff
@@ -1807,6 +1809,146 @@ Result Parser::ParseRun() {
...
+  }
+
+  command_list_.back()->SetDebugScript(std::move(dbg));
+
+  return Result();
+}
+
+Result Parser::ParseDebugThread(debug::Events* dbg) {
+  auto result = Result();
+  auto parseThread = [&](debug::Thread* thread) {
```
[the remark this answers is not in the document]
ben-clayton: That is quite a big step away from the record-and-replay design I have, as creating the `debug::ThreadScript` and passing in would require splitting the `debug::Thread` interface into the existing debugger interface, and one with a `now-replay-this-thread-I've-already-made` method.

FWIW, the original implementation had the `BreakOn` methods return a `debug::Thread`, instead of the callback, the the actual debugger implementation needs these threads to be driven concurrently, and I don't want to expose the hairy threading logic outside of the private debugger implementation.

I understand your concerns with lambda lifetimes here, but I do feel like the compact nature of the shared interface outweighs the additional code required to make the script callback-free. I'm happy to try and find a compromise though. 

Unrelated to what you asked, I've moved the bulk of the lambda into a `ParseDebugThreadBody` method as this function was a bit unruly to read. This also cleans up the `Result` returns quite a bit.

src/debug.cc
```diff
@@ -0,0 +1,115 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "src/debug.h"
+
+#include <memory>  // std::make_shared<>
```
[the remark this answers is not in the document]
ben-clayton: Removed

src/debug.cc
```diff
@@ -0,0 +1,115 @@
...
+#include <string>
+
+#include "src/make_unique.h"
+
+namespace {
```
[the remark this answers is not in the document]
ben-clayton: Done. Removed extra qualifications in the anonymous namespace that are now unnecessary.

docs/debugger.md
```diff
@@ -0,0 +1,174 @@
...
+/* 11 */   vout.color = color;
+/* 12 */   return vout;
+/* 13 */ }
+```
+
+The following performs a basic debugger test for the 3rd vertex in the triangle list:
+
+```groovy
+DEBUG pipeline DRAW_ARRAY AS TRIANGLE_LIST START_IDX 0 COUNT 6
+    THREAD VERTEX_INDEX 2
```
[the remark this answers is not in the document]
ben-clayton: If the invocation never runs, then the test fails. I've updated the docs to include this. Great catch!

src/engine.h
```diff
@@ -106,6 +114,11 @@ class Engine {
   /// This covers both Vulkan buffers and images.
   virtual Result DoBuffer(const BufferCommand* cmd) = 0;
 
+  /// GetDebugger retrieves the shader debugger from the engine.
+  /// If the engine does not support a shader debugger then the Result will be a
+  /// failure, and |out| will not be written to.
+  virtual Result GetDebugger(Debugger** out) = 0;
```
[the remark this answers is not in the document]
ben-clayton: Done

src/executor_test.cc
```diff
@@ -159,6 +159,10 @@ class EngineStub : public Engine {
     return {};
   }
 
+  Result GetDebugger(Debugger**) override {
+    return Result("EngineStub does not currenty support a debugger");
```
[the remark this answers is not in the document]
ben-clayton: Done

src/vulkan/engine_vulkan.h
```diff
@@ -87,6 +89,10 @@ class EngineVulkan : public Engine {
   std::unique_ptr<CommandPool> pool_;
 
   std::map<amber::Pipeline*, PipelineInfo> pipeline_map_;
+
+  class VkDebugger;
```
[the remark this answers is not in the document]
ben-clayton: Done

src/vulkan/engine_vulkan.h
```diff
@@ -87,6 +89,10 @@ class EngineVulkan : public Engine {
...
+  void DeleteDebugger();
```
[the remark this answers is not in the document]
ben-clayton: Done

third_party/CMakeLists.txt
ben-clayton: I changed SwiftShader so that it'll check for these to be set before promoting them up to options. This way they won't pollute the amber options, when they must not be changed.

src/vulkan/engine_vulkan_debugger.cc
ben-clayton: There are two functions that need a definition, which are stubbed at the bottom of this file:

```C++
std::pair<Engine::Debugger*, Result> EngineVulkan::GetDebugger() {
  return {nullptr,
          Result("Amber was not built with AMBER_ENABLE_VK_DEBUGGING enabled")};
}

void EngineVulkan::DeleteDebugger() {}
```

You want me to move their definition to `engine_vulkan.cc`?

I personally prefer to keep this sort of thing in the preprocessor if possible as it doesn't add additional complexity to each build system. I see that amber has CMake and Android build scripts. Anything file-conditional like this would need to be replicated carefully in both (plus any other build systems that might crop up in the future).

src/vulkan/engine_vulkan_debugger.cc
ben-clayton: Okay. Removed for now. You're right - if SwiftShader passes these, I think it's fairly safe to assume timeouts will not be an issue for other drivers. :)

src/vulkan/engine_vulkan_debugger.cc
ben-clayton: Removed the TODO.

src/vulkan/engine_vulkan_debugger.cc
ben-clayton: Done. Bit verbose though :)

src/vulkan/engine_vulkan_debugger.cc
ben-clayton: There's two ways the DAP can share source - by file path and by DAP protocol. For the embedded shader cases, it'll use the latter. However, we will probably want to write tests that exercise the by-path code path, in which case we'll need to check in the source files.
At no point will temporary files be written by the driver or amber.

src/vulkan/engine_vulkan_debugger.cc
ben-clayton: By "embedder", you mean something like dEQP?
Acknowledged for now - I'll have a think about this.

src/vulkan/engine_vulkan_debugger.cc
ben-clayton: Done

src/vulkan/engine_vulkan_debugger.cc
ben-clayton: I'm no c++ expert, but I believe that enums that do not have an explicit initializer are guaranteed to start at 0 and increment from there.

> "If the first enumerator has no initializer, the value of the corresponding constant is zero. An enumerator-definition without an initializer gives the enumerator the value obtained by increasing the value of the previous enumerator by one."

http://eel.is/c++draft/dcl.enum

Even if what you said were to happen, the hash would overflow, but would be still be stable.

src/vulkan/engine_vulkan_debugger.cc
ben-clayton: Done (I think this is what you wanted)

src/vulkan/engine_vulkan_debugger.cc
ben-clayton: Done

src/vulkan/engine_vulkan_debugger.cc
ben-clayton: Not while keeping everything [PImpl](https://en.cppreference.com/w/cpp/language/pimpl), as I'd then need to change the forward declaration to a true declaration in the header, due to the  unique_ptr needing to know about the destructor.

src/vulkan/engine_vulkan_debugger.cc
ben-clayton: See above

docs/debugger.md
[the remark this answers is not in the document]
dj2: That makes sense. Can you put those as comments in the above?

```
       # Debugger is at line 9, we can read the input variables
        EXPECT LOCATION "simple_vs.hlsl" 9 "  VS_OUTPUT vout;"
￼        EXPECT LOCAL "pos.x" EQ -1.007874
￼        EXPECT LOCAL "pos.y" EQ 1.000000
￼        EXPECT LOCAL "pos.z" EQ 0.000000
￼        EXPECT LOCAL "color.x" EQ 1.000000
￼        EXPECT LOCAL "color.y" EQ 0.000000
￼        EXPECT LOCAL "color.z" EQ 0.000000
￼        STEP_IN   # step to line 10
￼        EXPECT LOCATION "simple_vs.hlsl" 10 "  vout.pos = pos;"
￼        STEP_IN   # step to line 11 and read result of line 10
        EXPECT LOCAL "vout.pos.x" EQ -1.007874
￼        EXPECT LOCAL "vout.pos.y" EQ 1.000000
￼        EXPECT LOCAL "vout.pos.z" EQ 0.000000
￼        EXPECT LOCATION "simple_vs.hlsl" 11 "  vout.color = color;"
￼        STEP_IN  # step to line 12 and read result of line 11
￼        EXPECT LOCAL "vout.color.x" EQ 1.000000
￼        EXPECT LOCAL "vout.color.y" EQ 0.000000
￼        EXPECT LOCAL "vout.color.z" EQ 0.000000
￼        EXPECT LOCATION "simple_vs.hlsl" 12 "  return vout;"
￼        CONTINUE
```

Or something so it's a little clearer?

docs/debugger.md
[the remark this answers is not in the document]
dj2: sounds good.

src/amberscript/parser.cc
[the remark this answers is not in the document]
dj2: I think I didn't explain right. I think you can make the change I want and the external impact is in `BreakOnComputeGlobalInvocation` and `BreakOnVertexIndex` to just remove the first two lines. So, in ParseDebugThread we'd create the `debug::Thread` object. Call your ParseDebugThreadBody and pass in the thread and then pass the `thread` to the `BreakOn*` methods.

Basically, just makes the parsing straight line instead of taking a jog through the lambda. Does that makes sense?

(if what I"m saying still doesn't fit with what you want, please ignore me as this is technically fine, just took some thinking to figure out where the parse happens and what state the tokenizer would be in afterwards and I like to avoid thinking, heh)

src/vulkan/engine_vulkan_debugger.cc
dj2: The problem with reviewing lineally is you forget to remove comments when you realize they don't make sense. This bit is fine to do this way since we have a #else block later. If it was just conditionally blocking out all the code I'd say do it in the cmake file.

src/vulkan/engine_vulkan_debugger.cc
dj2: Yea, even the `amber` sample app is an 'embeddeder' of the amber library. Adding a `TODO` here is fine and file a bug and we can come back to it later.

src/vulkan/engine_vulkan_debugger.cc
dj2: sgtm. I just assume the compiler will do bad things, heh

src/vulkan/engine_vulkan_debugger.cc
dj2: yup

src/vulkan/engine_vulkan_debugger.cc
dj2: I think I'm ok with that if it isn't too much hassle. Removes the chance we can leak and removes the need for the custom delete method.

src/amberscript/parser.cc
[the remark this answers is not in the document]
ben-clayton: I think I understand you. Things just start unravelling at the:

 > Call your ParseDebugThreadBody and pass in the thread and then pass the thread to the BreakOn* methods.

We can expose the `ThreadScript` object (which is currently PImpl) - that's fine - and we can record the thread objects on this before handing it over to the `debug::Events`.

However, the symmetry and interface reuse falls apart when we actually come to doing the debugger replay. Currently the debugger internally creates a `std::thread` to concurrently run the threads, and on this `std::thread`, it calls the `BreakOn` callback, so that the recorded thread ops happen sequentially and block until they're complete. Given the current way things are structured, I don't see a trivial way to keep the interface the same after changing the parser to push the thread script.  

With all that said, I think I might have a way to do what you're after - but it requires a bit of restructuring. I'm hoping to get this into the hands of Jaebaek this week, so do you mind if I experiment a bit and come back to you with a follow up change?

src/vulkan/engine_vulkan_debugger.cc
ben-clayton: It occurred to me that we *can* use a `std::unique_ptr<Engine::Debugger>` field, which dodges the destructor problem.
Done.

src/vulkan/engine_vulkan_debugger.cc
ben-clayton: Done: https://github.com/google/amber/issues/777

src/amberscript/parser.cc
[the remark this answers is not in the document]
dj2: Sounds good. Please file a bug and add a TODO which references is.

src/vulkan/engine_vulkan_debugger.cc
```diff
@@ -0,0 +1,979 @@
...
+  std::mutex threads_mutex_;
+  PendingThreadsMap pendingThreads_;  // guarded by threads_mutex_
+  ThreadVector runningThreads_;       // guarded by threads_mutex_
+  std::mutex error_mutex_;
+  Result error_;  // guarded by error_mutex_
+};
+
+std::pair<Engine::Debugger*, Result> EngineVulkan::GetDebugger() {
+  if (!debugger_) {
+    auto debugger = new VkDebugger();
```
dj2: any reason to not `debugger_ = MakeUnique<Debugger>()` ? This currently leaks if you early return on line 954.

dj2:
This is pretty much lgtm once the bots are fixed. If you could add the comments about step_in, etc to the markdown file and the bug for the parsing bit. The only thing I see is the leak of the debugger if Connect fails.

docs/debugger.md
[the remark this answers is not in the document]
ben-clayton: Done. Thanks.

src/vulkan/engine_vulkan_debugger.cc
ben-clayton: Yeah, this downcasts and I'd need to upcast again to call the `Connect()` method.
Reshuffled things so that the leak shouldn't happen. Good spot!

src/amberscript/parser.cc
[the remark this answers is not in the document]
ben-clayton: It was actually easier to implement what I had in mind than I was expecting. Please see the new CL at the end of the stack.

This does use `std::shared_ptr` (frowned upon in google style), but given these `ThreadScripts` need to pass through `std::function` (which doesn't allow move-only captures), and potential reuse of the scripts, this seemed to pass the 'good reason' clause.

ben-clayton:
(Did another push to try to fix the kokoro flakes)

PR #783, opened 2020-02-08
Roll SwiftShader forward & fix debugger test

ben-clayton:
SwiftShader now ignores SPIR-V 'non-statement' opcodes for the debugger. These are no longer steppable.

`tests/cases/debugger_spirv_line_stepping.amber` has been updated to match this new behaviour.


ben-clayton:
(rerunning failed presubmits - I force pushed twice in rapid succession that broke the first pull)

PR #785, opened 2020-02-11
Debugger: Add support for EXPECT CALLSTACK

ben-clayton:
Checks the callstack is as expected.
Note: I've updated the .amber scripts with the current SwiftShader
output. It's clear that this is somewhat wonky, and will change.

### DIFF
--- a/src/vulkan/engine_vulkan_debugger.cc
+++ b/src/vulkan/engine_vulkan_debugger.cc
@@ -218,9 +218,20 @@ class Client {
 
   // TopStackFrame retrieves the frame at the top of the thread's call stack.
   // Returns true on success, false on error.
-  bool TopStackFrame(dap::integer threadId, dap::StackFrame* frame) {
+  bool TopStackFrame(dap::integer thread_id, dap::StackFrame* frame) {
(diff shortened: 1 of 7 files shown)


src/vulkan/engine_vulkan_debugger.cc
```diff
@@ -608,6 +619,53 @@ class Thread : public debug::Thread {
     }
   }
 
+  void ExpectCallstack(
+      const std::vector<debug::StackFrame>& callstack) override {
+    DEBUGGER_LOG("ExpectCallstack()");
+
+    std::vector<dap::StackFrame> got_stack;
+    if (!client_.Callstack(thread_id_, &got_stack)) {
+      return;
```
dj2: Should this be an error? We expected a callstack but didn't receive on?

ben-clayton: Callstack already records the error messages before returning `false`:
https://github.com/google/amber/blob/3bb1eea7f9e212e0bae0b5ae7cc1318ff4537217/src/vulkan/engine_vulkan_debugger.cc#L232-L246

PR #791, opened 2020-02-21
Implement DRAW_GRID

jarisiru:
Add support for draw_grid, which is fairly similar to draw_rect,
except it also takes parameters to split the rectangle into
desired number of columns and rows of cells.

This functionality is meant for the testing of vertex shaders,
giving an easy way to produce a lot of vertices.

The implementation is based on draw_rect. Dawn rendering is not
implemented.


src/vkscript/command_parser.cc
```diff
@@ -307,6 +309,74 @@ Result CommandParser::ProcessDrawRect() {
   return {};
 }
 
+Result CommandParser::ProcessDrawGrid() {
```
dj2: Drop this part. We aren't updating the vkscript format with new features. If you want to use `draw grid` then AmberScript is the way to go. We also don't want to diverge from the original vkscript as much as possible.

src/amberscript/parser.cc
```diff
@@ -1718,6 +1718,99 @@ Result Parser::ParseRun() {
     return ValidateEndOfStatement("RUN command");
   }
 
+  if (token->AsString() == "DRAW_GRID") {
+    if (!pipeline->IsGraphics())
+      return Result("RUN command requires graphics pipeline");
+
+    if (pipeline->GetVertexBuffers().size() > 1) {
```
dj2: The comment above says the index and vertex data are ignored. Why do we allow one vertex buffer to be attached if it will be ignored?

src/command.h
```diff
@@ -183,6 +187,52 @@ class DrawRectCommand : public PipelineCommand {
   float height_ = 0.0;
 };
 
+/// Command to draw a grid of recrangles on screen.
+class DrawGridCommand : public PipelineCommand {
+ public:
+  explicit DrawGridCommand(Pipeline* pipeline, PipelineData data);
```
dj2: explicit is only needed if there is a single parameter

src/command.h
```diff
@@ -183,6 +187,52 @@ class DrawRectCommand : public PipelineCommand {
...
+  ~DrawGridCommand() override;
+
+  const PipelineData* GetPipelineData() const { return &data_; }
```
dj2: This is always empty, can we remove it?

src/command.h
```diff
@@ -183,6 +187,52 @@ class DrawRectCommand : public PipelineCommand {
...
+
+  void EnableOrtho() { is_ortho_ = true; }
```
dj2: Ortho is only ever true, can we remove the member and just set it as needed?

src/command.h
```diff
@@ -183,6 +187,52 @@ class DrawRectCommand : public PipelineCommand {
...
+  bool IsOrtho() const { return is_ortho_; }
+
+  void EnablePatch() { is_patch_ = true; }
```
dj2: Is this needed? Can it be removed?

src/command.h
```diff
@@ -183,6 +187,52 @@ class DrawRectCommand : public PipelineCommand {
...
+  void SetY(float y) { y_ = y; }
+  float GetY() const { return y_; }
+
+  void SetWidth(float w) { width_ = w; }
+  float GetWidth() const { return width_; }
+
+  void SetHeight(float h) { height_ = h; }
+  float GetHeight() const { return height_; }
+
+  void SetColumns(float c) { columns_ = c; }
```
dj2: Columns and Rows are used as uint32_t, should these just be uint32_t instead of float?

src/vulkan/engine_vulkan.cc
```diff
@@ -450,6 +450,85 @@ Result EngineVulkan::DoDrawRect(const DrawRectCommand* command) {
...
+
+  auto* graphics = info.vk_pipeline->AsGraphics();
+
+  float x = command->GetX();
+  float y = command->GetY();
+  float width = command->GetWidth();
+  float height = command->GetHeight();
+  const uint32_t columns = static_cast<uint32_t>(command->GetColumns());
+  const uint32_t rows = static_cast<uint32_t>(command->GetRows());
+  const uint32_t vertices = columns * rows * 3 * 2;
```
dj2: Let's move the 3 and 2 to constants which will make this a lot more obvious

```
kTrianglesPerCell = 2
kVerticesPerTriangle = 3
```

src/vulkan/engine_vulkan.cc
```diff
@@ -450,6 +450,85 @@ Result EngineVulkan::DoDrawRect(const DrawRectCommand* command) {
...
+      // Top left
+      values[c + 8].SetDoubleValue(static_cast<double>(x + (width / columns) * j));
+      values[c + 9].SetDoubleValue(static_cast<double>(y + (height / rows) * i));
+      // Top right
+      values[c + 10].SetDoubleValue(static_cast<double>(x + (width / columns) * (j + 1)));
+      values[c + 11].SetDoubleValue(static_cast<double>(y + (height / rows) * i));
+    }
+  }
+
+
```
dj2: nit: remove blank line

src/vulkan/engine_vulkan.cc
```diff
@@ -450,6 +450,85 @@ Result EngineVulkan::DoDrawRect(const DrawRectCommand* command) {
...
+    const float frame_width = static_cast<float>(graphics->GetWidth());
+    const float frame_height = static_cast<float>(graphics->GetHeight());
+    x = ((x / frame_width) * 2.0f) - 1.0f;
+    y = ((y / frame_height) * 2.0f) - 1.0f;
+    width = (width / frame_width) * 2.0f;
+    height = (height / frame_height) * 2.0f;
+  }
+
+  std::vector<Value> values(vertices * 2);
+
```
dj2: Lets pull out the cell size calculations:

```
double cell_width = width / columns;
double cell_height = height / rows;
```

src/amberscript/parser.cc
jarisiru: Copied off from draw_rect. Fixed here, but not in draw_rect.

src/command.h
jarisiru: copied off draw_rect. Since the number of parameters here dropped to one, I left the explicit. Not fixed in draw_rect (or other calls here, which also have the explicit keyword).

src/command.h
jarisiru: Copied off draw_rect. Removed here, left in draw_rect.

src/command.h
jarisiru: Copied off draw_rect. Removed here, left in draw_rect.

src/command.h
jarisiru: Copied off draw_rect. This is related to tesselation, and may be useful in draw_rect. Removed here.

src/command.h
jarisiru: Done

src/vkscript/command_parser.cc
jarisiru: Removed all changes to vkscript.

src/vulkan/engine_vulkan.cc
jarisiru: Done. Why are these constants prefixed with k, but not others?

src/vulkan/engine_vulkan.cc
jarisiru: Done

src/vulkan/engine_vulkan.cc
jarisiru: Done

src/vulkan/engine_vulkan.cc
dj2: Mostly I was thinking they would move up to the anonymous namespace at the top of the file making them constants, which we prefix with a k. Leaving them in the method you could drop the k if you want, either way is fine.

PR #792, opened 2020-02-24
Add support for 8-bit int shader/storage features

rg3igalia:
This pull request adds support for 8-bit integers to be used in shaders, both as a basic type and for input/output as part of push constants and uniform/storage blocks. It adds 4 new features understood by Amberscript which are needed to request the corresponding Vulkan features. The original structure is called "8BitStorageFeatures" but it has been renamed to "Storage8BitFeatures" due to the leading digit causing problems to the parser.

### DIFF
--- a/tests/cases/int8.amber
+++ b/tests/cases/int8.amber
@@ -1,0 +1,53 @@
+#!amber
+# Copyright 2020 The Amber Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+INSTANCE_EXTENSION VK_KHR_get_physical_device_properties2
+DEVICE_EXTENSION VK_KHR_shader_float16_int8
+DEVICE_EXTENSION VK_KHR_storage_buffer_storage_class
+DEVICE_EXTENSION VK_KHR_8bit_storage
+DEVICE_FEATURE Float16Int8Features.shaderInt8
+DEVICE_FEATURE Storage8BitFeatures.uniformAndStorageBuffer8BitAccess
+DEVICE_FEATURE Storage8BitFeatures.storagePushConstant8
+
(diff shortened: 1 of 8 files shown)


tests/cases/int8.amber
```diff
@@ -0,0 +1,58 @@
...
+layout(push_constant) uniform PushConstantsBlock {
+  int8_t factor;
+} pushConstants;
+
+void main() {
+  data.value = data.value * pushConstants.factor;
+}
+END
+
+BUFFER buf DATA_TYPE int8 DATA
```
dj2: minor nit: this could be written as:

```
BUFFER buf DATA_TYPE int8 DATA 63 END
```

dj2:
Looks like the NDK doesn't have the 8bit storage feature? Does this require a higher NDK version?

```
../src/vulkan/device.cc:419:18: error: use of undeclared identifier 'VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_8BIT_STORAGE_FEATURES'; did you mean 'VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_16BIT_STORAGE_FEATURES'?
                 VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_8BIT_STORAGE_FEATURES) {
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                 VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_16BIT_STORAGE_FEATURES
/tmpfs/src/android-ndk-r20/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/include/vulkan/vulkan_core.h:213:5: note: 'VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_16BIT_STORAGE_FEATURES' declared here
    VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_16BIT_STORAGE_FEATURES = 1000083000,
    ^
```

rg3igalia:
I'm not an Android developer so I'm a bit lost here. That's from the VK_KHR_8bit_storage extension that was promoted to Vulkan 1.2. I'll change that line to use the equivalent VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_8BIT_STORAGE_FEATURES_KHR enum value (note the _KHR suffix) which may fix the build problem.

tests/cases/int8.amber
rg3igalia: Will fix.

dj2:
I've found that if there is a KHR and non-KHR version the KHR is always safer.

rg3igalia:
Yes, it was actually an oversight on my part.

PR #793, opened 2020-02-25
Switch to rsplit for repo path in git-sync-deps

### LINKED ISSUE #786: Support ssh in git-sync-deps
The git-sync-deps script doesn't currently support ssh URLs like "ssh://git@server" because of how it splits strings by '@' character:

```
    if '@' in dependencies[directory]:
      repo, checkoutable = dependencies[directory].split('@', 1)
```

I'd suggest checking the string for two @ characters and recombining the first two into the repo string.

Some Khronos gitlab servers only(?) support ssh, so it would be nice for Amber to support this.

dj2:
This CL changes git-sync-deps to split from the right side when looking
for the @ sign. This should allow SSH URLs which have multiple @'s in
them to work properly.

Fixes #786

### DIFF
--- a/tools/git-sync-deps
+++ b/tools/git-sync-deps
@@ -224,7 +224,7 @@ def git_sync_deps(deps_file_path, command_line_os_requests, verbose):
   list_of_arg_lists = []
   for directory in sorted(dependencies):
     if '@' in dependencies[directory]:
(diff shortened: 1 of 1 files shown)


dj2:
@jeffbolznv does this work for your use case? (I don't have any ssh hosts to test against)

jeffbolznv:
Works great, thanks!

PR #798, opened 2020-02-26
Fix float16 zeros and denormals handling

rg3igalia:
Handling of 0.0 as float16 was broken. It can be easily tested by replacing 2.4 in tests/cases/float16.amber with 0.0, which makes amber assert in float16_helper.cc:48, unable to convert 0.0 to float16. This PR handles this special case and flushes denormals to zero.

### DIFF
--- a/src/float16_helper.cc
+++ b/src/float16_helper.cc
@@ -43,7 +43,11 @@ uint16_t FloatSign(const uint32_t hex_float) {
 
 // Return exponent value of 32 bits float.
 uint16_t FloatExponent(const uint32_t hex_float) {
-  uint32_t exponent = ((hex_float >> 23U) & ((1U << 8U) - 1U)) - 112U;
+  uint32_t exponent_bits = ((hex_float >> 23U) & ((1U << 8U) - 1U));
+  // Handle zero and denormals.
+  if (exponent_bits == 0U)
+    return 0;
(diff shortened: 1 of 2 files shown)


src/float16_helper.cc
```diff
@@ -59,11 +63,18 @@ uint32_t FloatMantissa(const uint32_t hex_float) {
 // based on IEEE-754.
 float HexFloat16ToFloat(const uint8_t* value) {
   uint32_t sign = (static_cast<uint32_t>(value[1]) & 0x80) << 24U;
-  uint32_t exponent = (((static_cast<uint32_t>(value[1]) & 0x7c) >> 2U) + 112U)
-                      << 23U;
-  uint32_t mantissa = ((static_cast<uint32_t>(value[1]) & 0x3) << 8U |
-                       static_cast<uint32_t>(value[0]))
-                      << 13U;
+  uint32_t exponent_bits = (static_cast<uint32_t>(value[1]) & 0x7c) >> 2U;
+  uint32_t exponent, mantissa;
```
dj2: nit: Put these on separate lines. Could also do:

```
uint32_t exponent = 0;
uint32 _t mantissa = 0;
if (exponent_bits != 0U) {
  ...
}
```

rg3igalia: No problem. Fixed!

PR #806, opened 2020-03-03
Add polygon mode support for AmberScript

asuonpaa:
Polygon mode was only supported by VkScript. This adds the support for AmberScript.


src/amberscript/parser.cc
```diff
@@ -1884,6 +1906,7 @@ Result Parser::ParseRun() {
     cmd->SetTopology(topo);
     cmd->SetFirstVertexIndex(start_idx);
     cmd->SetVertexCount(count);
+    cmd->SetPolygonMode(pipeline->GetPolygonMode());
```
dj2: Does the DrawGridCommand and DrawRectCommand also need to have the PolygonMode set to be consistent?

asuonpaa: Propably yes. I'll add it there too.

PR #812, opened 2020-03-12
Add support for push constants generated by Clspv

jrprice:
When the -work-dim or -global-offset flags are passed to Clspv, it
will generate push constants. The sizes and offsets of the push
constant values are provided in the descriptor map. Amber needs to add
these to the pipeline layout and create a buffer for them.


src/pipeline.cc
```diff
@@ -825,4 +826,61 @@ Result Pipeline::GenerateOpenCLLiteralSamplers() {
...
+  for (const auto& pc : shader_info.GetPushConstants()) {
+    if (pc.offset + pc.size > bytes.size()) {
+      bytes.resize(pc.offset + pc.size);
+    }
+
+    switch (pc.type) {
+      case Pipeline::ShaderInfo::PushConstant::kPushConstantDimensions:
+        // All compute kernel launches are 3D.
+        assert(pc.size == 4);
+        *reinterpret_cast<uint32_t*>(bytes.data() + pc.offset) = 3;
```
dj2: Would it be simpler to just make the buffer a `uint32_t` and then multiply the size  on line 872 by `sizeof(uint32_t)` ?

src/pipeline.h
```diff
@@ -112,6 +112,23 @@ class Pipeline {
       return descriptor_map_;
     }
 
+    /// Push constant information for an OpenCL-C shader.
+    struct PushConstant {
+      enum {
```
dj2:
```
enum class PushConstantType {
  kDimensions = 0;
  kGlobalOffset;
}
PushConstantType type;
```

src/pipeline_test.cc
```diff
@@ -821,4 +821,38 @@ TEST_F(PipelineTest, OpenCLGenerateLiteralSamplers) {
...
+  pc2.size = 12;
+  p.GetShaders()[0].AddPushConstant(std::move(pc2));
+
+  auto r = p.GenerateOpenCLPushConstants();
+  ASSERT_TRUE(r.IsSuccess());
+
+  const auto& buf = p.GetPushConstantBuffer();
+  EXPECT_EQ(28U, buf.buffer->GetSizeInBytes());
+
+  uint32_t* bytes = reinterpret_cast<uint32_t*>(buf.buffer->ValuePtr()->data());
```
dj2: The buffer has a templated GetValues, so I think `buf.buffer->GetValues<uint32_t>()` would work here?

src/pipeline.cc
jrprice: Sort of. I'm trying to make it simple for us to support values that are not 32-bits in the future (since we may use push constants for kernel arguments at some point). I've made something that resembles your suggested change though, as it does simplify some logic.

PR #823, opened 2020-03-23
Added a support for loading a png file into a buffer

asuonpaa:


include/amber/amber.h
```diff
@@ -116,6 +117,9 @@ struct Options {
   bool disable_spirv_validation;
   /// Delegate implementation
   Delegate* delegate;
+
+  std::function<amber::Result(std::string, amber::BufferInfo*)>
```
dj2: Move the callback method up to the Delegate class above instead of as an option. (Please add a comment as well)

samples/png.cc
```diff
@@ -81,4 +81,22 @@ amber::Result ConvertToPNG(uint32_t width,
   return {};
 }
 
+amber::Result LoadPNG(const std::string file_name,
+                      uint32_t* width,
+                      uint32_t* height,
+                      std::vector<amber::Value>* values) {
+  std::vector<uint8_t> decoded_buffer;
+  if (lodepng::decode(decoded_buffer, *width, *height, file_name,
+                      LodePNGColorType::LCT_RGBA, 8) != 0)
```
dj2: nit: Add {}'s since the condition is multiple lines.

src/executor.cc
```diff
@@ -88,6 +88,21 @@ Result Executor::Execute(Engine* engine,
 
   Engine::Debugger* debugger = nullptr;
 
+  // Load data to buffers
+  for (const auto& buf : script->GetBuffers()) {
+    if (buf->GetDataFile() != "") {
```
dj2:
```
if (buf->GetDataFile().empty())
   continue;
```

src/executor.cc
```diff
@@ -88,6 +88,21 @@ Result Executor::Execute(Engine* engine,
...
+      BufferInfo info;
+      Result r = options->loadBufferDataFunc(buf->GetDataFile(), &info);
+
```
dj2: nit: remove blank line

tests/run_tests.py
```diff
@@ -43,14 +43,20 @@
     "multiple_ubo_update_with_graphics_pipeline.vkscript",
     # DXC not currently building on bot
     "draw_triangle_list_hlsl.amber",
+    # Texture file not found from this path
```
dj2: If this is suppressed on all platforms, where does this work? Why is the path not found?

docs/amber_script.md
```diff
@@ -156,6 +156,10 @@ BUFFER {name} DATA_TYPE {type} {STD140 | STD430} WIDTH {w} HEIGHT {h} \
 # The buffer will be sized based on the `RENDER_SIZE` of the `PIPELINE`.
 BUFFER {name} FORMAT {format_string} \
     [ MIP_LEVELS _mip_levels_ (default 1) ]
+
+# Load buffer data from a PNG image with file name specified by `FILE`.
```
dj2: Can you please comment what the file path is? Is it relative to the amber file? Something else? Should I be using `/` or `\` between path elements?

tests/run_tests.py
asuonpaa: I'm considering it more as a sample than a test since it doesn't have any EXPECT commands. The texture file is under tests/cases, but the script is run under tests/ so it doesn't find the file. I could change the file name in the amber file, but then it wouldn't work when running manually from the same path as the amber file.

docs/amber_script.md
asuonpaa: The path string is cosumed by LodePNG. / is working on Linux, but I haven't tried this on Windows (yet).

tests/run_tests.py
dj2: Right, I think we need to setup everything so the file path is relative to the `.amber` file. This means that `amber.cc` will need to store the path to the `.amber` and prepend that to the filename before calling the loader.

asuonpaa: Ok, sounds reasonable.

PR #825, opened 2020-03-25
Enable DXC and kokoro test for it

jaebaek:
This commit makes amber successfully run
tests/cases/draw_triangle_list_hlsl.amber and add it to kokoro test


tests/run_tests.py
```diff
@@ -42,15 +44,15 @@
     "multiple_ssbo_update_with_graphics_pipeline.vkscript",
     "multiple_ubo_update_with_graphics_pipeline.vkscript",
     # DXC not currently building on bot
-    "draw_triangle_list_hlsl.amber",
+    DXC_TEST_CASE,
```
dj2: Remove these from this set of SUPPRESSIONS and make a SUPPRESSIONS_DXC similar to SUPPRESSIONS_DAWN

jaebaek: I followed the behavior of `--use-opencl`. PTAL.

PR #831, opened 2020-04-02
Add support for VIRTUAL_FILE

ben-clayton:
`VIRTUAL_FILE` declares a file in a new virtual file system.
These files can now be referenced by the `SHADER` commands instead of using inline source.

The DXC compiler will first look in the virtual file system for any `#include`s before falling back to the standard file system.

This allows us to write tests that exercise the compiler and debugger handling of multiple files.

Added `relative_includes_hlsl.amber` which checks that DXC correctly includes relative to the current file, not the root file.

### DIFF
--- a/docs/amber_script.md
+++ b/docs/amber_script.md
@@ -68,7 +68,62 @@ set of data types.
 SET ENGINE_DATA {engine data variable} {value}*
 ```
 
+### Virtual File Store
+
+Each amber script contains a virtual file system that can store files of textual
+data. This lets you bundle multiple source files into a single, hermetic amber
+script file.
+
+Virtual files are declared using the `VIRTUAL_FILE` command:
+
+```groovy
+VIRTUAL_FILE {path}
+ {file-content}
+END
+```
+
+Paths must be unique.
+
+Shaders can directly reference these virtual files for their source. \
+HLSL shaders that `#include` other `.hlsl` files will first check the virtual
+file system, before falling back to the standard file system.
+
 ### Shaders
+
+Shader programs are declared using the `SHADER` command. \
+Shaders can be declared as `PASSTHROUGH`, with inlined source or using source
+from a `VIRTUAL_FILE`.
+
+Pass-through shader:
+
+```groovy
+# Creates a passthrough vertex shader. The shader passes the vec4 at input
+# location 0 through to the `gl_Position`.
+SHADER vertex {shader_name} PASSTHROUGH
+```
+
+Shader using inlined source:
+
+```groovy
+# Creates a shader of |shader_type| with the given |shader_name|. The shader
+# will be of |shader_format|. The shader source then follows and is terminated
+# with the |END| tag.
+SHADER {shader_type} {shader_name} {shader_format}
+{shader_source}
+END
+```
+
+Shader using source from `VIRTUAL_FILE`:
+
+```groovy
+# Creates a shader of |shader_type| with the given |shader_name|. The shader
+# will be of |shader_format|. The shader will use the virtual file with |path|.
+SHADER {shader_type} {shader_name} {shader_format} VIRTUAL_FILE {path}
+```
+
+`{shader_name}` is used to identify the shader to attach to `PIPELINE`s,
+
+`{shader_type}` and `{shader_format}` are described below:
 
 #### Shader Type
  * `vertex`
@@ -92,23 +147,10 @@ types, but in that case must only provide a single shader type in the module.
 
 #### Shader Format
  * `GLSL`  (with glslang)
- * `HLSL`  (with dxc or glslang if dxc disabled)  -- future
+ * `HLSL`  (with dxc or glslang if dxc disabled)
  * `SPIRV-ASM` (with spirv-as)
  * `SPIRV-HEX` (decoded straight to SPIR-V)
  * `OPENCL-C` (with clspv)
-
-```groovy
-# Creates a passthrough vertex shader. The shader passes the vec4 at input
-# location 0 through to the `gl_Position`.
-SHADER vertex {shader_name} PASSTHROUGH
-
-# Creates a shader of |shader_type| with the given |shader_name|. The shader
-# will be of |shader_format|. The shader should then be inlined before the
-# |END| tag.
-SHADER {shader_type} {shader_name} {shader_format}
-...
-END
-```
 
 ### Buffers
 
@@ -471,7 +513,7 @@ RUN {pipeline_name} \
 
 ```groovy
 # Run the given |pipeline_name| which must be a `graphics` pipeline. The
-# grid at |x|, |y|, |width|x|height|, |columns|x|rows| will be rendered. 
+# grid at |x|, |y|, |width|x|height|, |columns|x|rows| will be rendered.
 # Ignores VERTEX_DATA and INDEX_DATA on the given pipeline.
 # For columns, rows of (5, 4) a total of 5*4=20 rectangles will be drawn.
 RUN {pipeline_name} \
--- a/src/amberscript/parser.cc
+++ b/src/amberscript/parser.cc
@@ -200,6 +200,8 @@ Result Parser::Parse(const std::string& data) {
       r = ParseStruct();
     } else if (tok == "SAMPLER") {
       r = ParseSampler();
+    } else if (tok == "VIRTUAL_FILE") {
+      r = ParseVirtualFile();
     } else {
       r = Result("unknown token: " + tok);
     }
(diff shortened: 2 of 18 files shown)


docs/amber_script.md
```diff
@@ -68,8 +68,63 @@ set of data types.
...
+
+```groovy
+VIRTUAL_FILE {path}
+ {file-content}
+END
+```
+
+Paths must be unique.
+
+Shaders can directly reference these virtual files for their source. \
```
dj2, line 87: What's with the \'s at the end of some of these?

src/amberscript/parser.cc
```diff
@@ -370,19 +372,42 @@ Result Parser::ParseShaderBlock() {
...
+      return r;
 
-  shader->SetData(data);
+    auto path = token->AsString();
 
-  token = tokenizer_->NextToken();
-  if (!token->IsIdentifier() || token->AsString() != "END")
-    return Result("SHADER missing END command");
+    std::string data;
+    r = script_->GetVirtualFiles()->Get(path, &data);
```
dj2, line 390: What about just `script_->GetVirtualFile(path, &data);` keep the implementation hidden

src/amberscript/parser.cc
```diff
@@ -2896,5 +2921,24 @@ Result Parser::ParseTolerances(std::vector<Probe::Tolerance>* tolerances) {
   return {};
 }
 
+Result Parser::ParseVirtualFile() {
+  auto token = tokenizer_->NextToken();
+  if (!token->IsIdentifier() && !token->IsString())
+    return Result("invalid virtual file path");
+  auto path = token->AsString();
```
dj2, line 2928: nit: add blank line above

src/amberscript/parser.cc
```diff
@@ -2896,5 +2921,24 @@ Result Parser::ParseTolerances(std::vector<Probe::Tolerance>* tolerances) {
...
+  if (!r.IsSuccess())
+    return r;
+
+  auto data = tokenizer_->ExtractToNext("END");
+
+  token = tokenizer_->NextToken();
+  if (!token->IsIdentifier() || token->AsString() != "END")
+    return Result("VIRTUAL_FILE missing END command");
+
+  return script_->GetVirtualFiles()->Add(path, data);
```
dj2, line 2940: `script_->SetVirtualFile(path, data)`

src/amberscript/parser_shader_test.cc
```diff
@@ -235,6 +235,26 @@ END)";
   EXPECT_EQ("2: extra parameters after SHADER command: INVALID", r.Error());
 }
```
dj2, line 237: Add a test for duplicate virtual file name

src/script.cc
```diff
@@ -18,7 +18,7 @@
 
 namespace amber {
 
-Script::Script() = default;
+Script::Script() : virtual_files_(new VirtualFileStore()) {}
```
dj2, line 21: nit: `virtual_files_(std::make_unique<VirtualFileStore>())`

src/virtual_file_store.cc
```diff
@@ -0,0 +1,48 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "src/virtual_file_store.h"
+
+namespace amber {
+namespace {
+
+bool HasPrefix(const std::string& str, const std::string& prefix) {
+  return str.rfind(prefix, 0) == 0;
```
dj2, line 21: Why `rfind`? Is this just `str.compare(0, prefix.size(), prefix);`

src/virtual_file_store.h
```diff
@@ -0,0 +1,64 @@
...
+      return Result("Virtual file '" + path + "' already declared");
+    }
+    files_by_path_.emplace(canonical, content);
+    return {};
+  }
+
+  /// Look up the virtual file by path. If the file was found, the content is
+  /// assigned to content.
+  Result Get(const std::string& path, std::string* content) const {
+    auto canonical = GetCanonical(path);
```
dj2, line 48: Should we either `assert(content)` or check and return failure if it's nullptr?

docs/amber_script.md
ben-clayton: Forces a new line in markup. Double newline starts a new paragraph (larger blank space)

src/amberscript/parser.cc
ben-clayton: `VirtualFileStore` is handed down to the compilers in places the full script is not available. Also gives me a clean place to add the virtual file tests, without cluttering the script further.

src/amberscript/parser.cc
ben-clayton: Ack.

src/amberscript/parser.cc
ben-clayton: See above. I could add helpers to the script to do forward on to the `VirtualFileStore` if you prefer?

src/amberscript/parser_shader_test.cc
ben-clayton: Ack

src/amberscript/parser.cc
dj2: That's fine, and `Get` and `Add` make sense as the API into the virtual file system. I'm just thinking if the parser should care, or if it should just tell the script has files to add/get.

src/script.cc
ben-clayton: `make_unique` was introduced in C++14. I thought amber targeted 11?

src/amberscript/parser.cc
dj2: Yea, that's what I was thinking.

src/script.cc
dj2: Been working in tint too long, `virtual_files_(MakeUnique<VirtualFileStore>())`

ben-clayton: Ack.

src/virtual_file_store.h
ben-clayton: Do you mean zero length? I could certainly do this in `Set`. I think the `Get` handler in this case is probably fine (it won't be found).

src/virtual_file_store.cc
ben-clayton: No good reason. First example I found on the web. Obviously not a good one.

src/virtual_file_store.h
dj2: Content is a pointer, so what happens if I call `Get("foo", nullptr)` ?

src/amberscript/parser.cc
ben-clayton: Done.

src/amberscript/parser.cc
ben-clayton: Done.

src/amberscript/parser.cc
ben-clayton: Done.

src/amberscript/parser_shader_test.cc
ben-clayton: Done. Also added one for blank paths.

src/script.cc
ben-clayton: Done.

src/virtual_file_store.cc
ben-clayton: Done.

src/virtual_file_store.h
ben-clayton: Done on both.

ben-clayton: Oh you mean `content`. Yeah, could `assert()` - not sure it'll be _that much_ clearer than the null-pointer segfault you'll get. :)

If this is a standard thing done in amber for out variables, I can add.

ben-clayton: I saw some other places that does indeed `assert()` on out parameters. Done.

dj2: I don't remember what the standard is in amber, but `assert(content && "content should not be nullptr")` can be clearer, heh

ben-clayton:
Thanks. Will address style issue and anything else that's flagged by presubmits.

src/virtual_file_store.h
ben-clayton: I've seen some compilers warn about `&&`'ing with string literals, so I'll opt to keep as a basic `assert(content)` if that's okay.

dj2: Works for me.

ben-clayton:
`CI-macos-clang-debug` failed on:
* `./tests/cases/copy_format_buffer.amber`
* `./tests/cases/magfilter_nearest.amber`

With the test failures:

```
[ VUID-vkFreeMemory-memory-00677 ] Object: 0x7f874efb3808 (Type = 3) | Cannot call vkFreeMemory on VkDeviceMemory 0x1a000000001a[] that is currently in use by a command buffer. The Vulkan spec states: All submitted commands that refer to memory (via images or buffers) must have completed execution (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-vkFreeMemory-memory-00677)
[ERROR] validation layer (Validation):
[ VUID-vkDestroyBuffer-buffer-00922 ] Object: 0x190000000019 (Type = 9) | Cannot free VkBuffer 0x190000000019[] that is in use by a command buffer. The Vulkan spec states: All submitted commands that refer to buffer, either directly or via a VkBufferView, must have completed execution (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-vkDestroyBuffer-buffer-00922)
[ERROR] validation layer (Validation):
[ VUID-vkDestroyRenderPass-renderPass-00873 ] Object: 0x7f874efb3808 (Type = 3) | Cannot call vkDestroyRenderPass on VkRenderPass 0x60000000006[] that is currently in use by a command buffer. The Vulkan spec states: All submitted commands that refer to renderPass must have completed execution (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-vkDestroyRenderPass-renderPass-00873)
[ERROR] validation layer (Validation):
...
```

@dj2 thinks that this is a timeout teardown issue, which makes sense, as this particular CI bot is running **really** slowly.

I see nothing to suggest that this change has affected the build state, so @dj2 suggests landing this PR, and following up with suppressing/fixing the flaky tests.

PR #835, opened 2020-04-03
Debugger: Scan for VIRTUAL_FILEs and wait for steps

ben-clayton:
If the `dap::Source` does not provide the embedded source, scan the virtual file system for the file by path.

Also fix a race condition where we would instruct the debugger to step, but not actually wait for it to notify us that the step had completed before making further checks.

Updated the `debugger_hlsl_line_stepping.amber` to:
* Use `VIRTUAL_FILE` for the file content
* Fix the scoping of `CompilationUnit` <- `Function` <- `LexicalBlock`
* Remove the unused and incorrect debug info from the fragment shader. We'll have proper testing for this once DXC produces correct debug info.

### DIFF
--- a/tests/cases/debugger_hlsl_line_stepping.amber
+++ b/tests/cases/debugger_hlsl_line_stepping.amber
@@ -15,6 +15,21 @@
 
 SET ENGINE_DATA fence_timeout_ms 1000000
 
+VIRTUAL_FILE "vertex-shader.hlsl"
+/* 1*/ struct VS_OUTPUT {
+/* 2*/   float4 pos : SV_POSITION;
+/* 3*/   float4 color : COLOR;
+/* 4*/ };
+/* 5*/
+/* 6*/ VS_OUTPUT main(float4 pos : POSITION,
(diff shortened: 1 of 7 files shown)


tests/cases/debugger_hlsl_line_stepping.amber
```diff
@@ -25,35 +40,7 @@ SHADER vertex vtex_shader SPIRV-ASM
...
-
-VS_OUTPUT main(float4 pos : POSITION,
-               float4 color : COLOR) {
-  VS_OUTPUT vout;
-  vout.pos = pos;
-  vout.color = color;
-  return vout;
-}
-"
-               OpSource HLSL 600 %src "
```
dj2, line 43: This lost the OpSource, was that intentional?

ben-clayton: Yes. The content is now in the `VIRTUAL_FILE`.
We want to test that we're grabbing the file content from the virtual file, not the embedded source.

ben-clayton:
Rebasing...

PR #837, opened 2020-04-08
Add depth/stencil image readback to Amber buffer

asuonpaa:
Depth/stencil attachments were not copied back to Amber buffers like color buffers were. With these changes the data is copied, which allows for probing the depth and stencil data as well as sharing the depth/stencil buffer between pipelines.

Also moved polygon mode under pipeline data as a preparation for depth and stencil parameter parsing.


src/amberscript/parser.cc
```diff
@@ -1807,7 +1807,7 @@ Result Parser::ParseRun() {
 
     auto cmd = MakeUnique<DrawGridCommand>(pipeline);
     cmd->SetLine(line);
-    cmd->SetPolygonMode(pipeline->GetPolygonMode());
```
dj2, line 1810: The other two no-longer set polygon mode, why does this one? Seems inconsistent. I think the DrawGridCommand should be changed to accept the pipeline and SetPolygonMode removed from cmd.

src/vulkan/graphics_pipeline.cc
```diff
@@ -767,42 +772,41 @@ Result GraphicsPipeline::Clear() {
   colour_clear.color = {
       {clear_color_r_, clear_color_g_, clear_color_b_, clear_color_a_}};
 
-  Result r = ClearBuffer(colour_clear, VK_IMAGE_ASPECT_COLOR_BIT);
```
dj2, line 770: Why was the method removed and this merged into one big function?

src/vulkan/transfer_image.cc
```diff
@@ -203,12 +219,17 @@ VkBufferImageCopy TransferImage::CreateBufferImageCopy(uint32_t mip_level) {
...
+  const VkImageAspectFlagBits aspects[] = {VK_IMAGE_ASPECT_COLOR_BIT,
+                                           VK_IMAGE_ASPECT_DEPTH_BIT,
+                                           VK_IMAGE_ASPECT_STENCIL_BIT};
   std::vector<VkBufferImageCopy> copy_regions;
   uint32_t last_mip_level = used_mip_levels_ == VK_REMAINING_MIP_LEVELS
                                 ? mip_levels_
                                 : base_mip_level_ + used_mip_levels_;
   for (uint32_t i = base_mip_level_; i < last_mip_level; i++)
-    copy_regions.push_back(CreateBufferImageCopy(i));
+    for (auto aspect : aspects)
```
dj2, line 230: nit: Both these for()'s should have {}'s around ther bodies as they're more then 1 line long.

src/vulkan/transfer_image.cc
```diff
@@ -182,15 +190,23 @@ Result TransferImage::CreateVkImageView() {
...
 
-VkBufferImageCopy TransferImage::CreateBufferImageCopy(uint32_t mip_level) {
+VkBufferImageCopy TransferImage::CreateBufferImageCopy(
+    VkImageAspectFlags aspect,
+    uint32_t mip_level) {
   VkBufferImageCopy copy_region = VkBufferImageCopy();
-  copy_region.bufferOffset = 0;
+  if (aspect == VK_IMAGE_ASPECT_STENCIL_BIT) {
+    // Store stencil data at the end of the buffer after depth data.
+    copy_region.bufferOffset =
```
dj2, line 199: Does the buffers size already account for depth + stencil?

src/amberscript/parser.cc
asuonpaa: I agree. Let me try that.

src/vulkan/graphics_pipeline.cc
asuonpaa: Because it contained code that should've been done only once before copying color and depth/stencil buffer. Even the original implementation copied color buffers twice if depth buffer was present. Basically there was nothing left in the function when fixing this.

src/vulkan/transfer_image.cc
asuonpaa: It does. The size is queried from the driver.

src/command.h
```diff
@@ -192,9 +190,11 @@ class DrawRectCommand : public PipelineCommand {
 /// Command to draw a grid of recrangles on screen.
 class DrawGridCommand : public PipelineCommand {
  public:
-  explicit DrawGridCommand(Pipeline* pipeline);
+  explicit DrawGridCommand(Pipeline* pipeline, PipelineData data);
```
dj2, line 193: nit: explicit not needed here as there are 2 parameters.

PR #838, opened 2020-04-08
Added implementation for loading buffer data from a binary file.

asuonpaa:


samples/amber.cc
```diff
@@ -337,9 +337,21 @@ class SampleDelegate : public amber::Delegate {
...
-    // TODO(asuonpaa): Try to load a binary format.
+    // Try to load a binary format.
+    auto data = ReadFile(path_ + file_name);
+    if (data.empty())
+      return amber::Result("Failed to load buffer data " + file_name);
 
-    return amber::Result("Failed to load buffer data " + file_name);
+    for (auto d : data) {
+      amber::Value v;
+      v.SetIntValue(d);
```
dj2, line 347: Does this not need to match the DATA_TYPE of the buffer? If the buffer is all float data, why is this written as all Int's?

samples/amber.cc
```diff
@@ -337,9 +337,21 @@ class SampleDelegate : public amber::Delegate {
...
+      buffer->values.push_back(v);
+    }
+
+    buffer->width = 1;
```
dj2, line 351: Why are these set?

tests/cases/buffer_load_binary.amber
```diff
@@ -0,0 +1,55 @@
...
+  float out_data[12];
+};
+
+void main() {
+    for (int i = 0; i < 12; i++)
+        out_data[i] = in_data[i] * 2.0;
+}
+END
+
+BUFFER buf0 DATA_TYPE vec4<float> SIZE 3 FILE vec4data.bin
```
dj2, line 34: Looking at the test case, I'm not sure if binary is what we want here. The `vec4data.bin` file seems like it would be much more useful if it was actually text? So, then I could look at it and edit it? Right now, how do you create/edit one of these binary files?

src/executor.cc
```diff
@@ -99,7 +99,16 @@ Result Executor::Execute(Engine* engine,
     if (!r.IsSuccess())
       return r;
 
-    buf->SetData(info.values);
+    std::vector<uint8_t>* data = buf->ValuePtr();
+
+    data->clear();
```
dj2, line 104: This seems like a separate fix?

samples/amber.cc
asuonpaa: Data is always loaded as byte format. There's no point converting the data here and then converting it back to bytes when loading the data into a buffer.

samples/amber.cc
asuonpaa: That's the width and height a buffer gets when initializing it inline. I'm matching it to avoid errors in buffer comparison.

src/executor.cc
asuonpaa: It's just a more general solution that doesn't assume the destination buffer is in uint8 format. It works both with the png images and buffers of any format. Basically we keep the destination buffer format but directly copy the bytes and then adjust the element size of the destination buffer based on the format.

PR #839, opened 2020-04-15
Added buffer data loading from a text file

asuonpaa:
Also moved delegate from options to Amber object to be able to access it from a parser.


src/amberscript/parser.h
```diff
@@ -35,6 +35,7 @@ namespace amberscript {
 class Parser : public amber::Parser {
  public:
   Parser();
```
dj2, line 37: Do we need to keep this constructor or can we remove it?

src/amberscript/parser_buffer_test.cc
```diff
@@ -1092,15 +1092,9 @@ TEST_F(AmberScriptParserTest, BufferDataFilePng) {
...
-
-  auto script = parser.GetScript();
-  const auto& buffers = script->GetBuffers();
-  ASSERT_EQ(1U, buffers.size());
+  ASSERT_FALSE(r.IsSuccess());
 
-  ASSERT_TRUE(buffers[0] != nullptr);
-  EXPECT_EQ("foo.png", buffers[0]->GetDataFile());
-  EXPECT_EQ(BufferDataFileType::kPng, buffers[0]->GetDataFileType());
+  EXPECT_EQ("1: missing delegate", r.Error());
```
dj2, line 1097: These tests fee like they should set the delegate so we get the same result out?

src/vkscript/parser.h
```diff
@@ -32,6 +32,7 @@ namespace vkscript {
 class Parser : public amber::Parser {
  public:
   Parser();
```
dj2, line 34: ditto, can we remove this constructor?

android_test/test.cc
```diff
@@ -17,5 +17,5 @@
 #include "amber/amber.h"
 
 void android_main(struct android_app* /*state*/) {
-  amber::Amber amber;
+  amber::Amber amber(nullptr);
```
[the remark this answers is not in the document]
asuonpaa: It expects a delegate pointer. Do you mean I should create a dummy delegate object here?

src/amberscript/parser.h
asuonpaa: We can remove this but it would require passing a nullptr in every parsing unit test. That's why I added a default constructor that would pass that nullptr for you.

src/amberscript/parser_buffer_test.cc
asuonpaa: Both GetDataFile and GetDataFileType were removed now that the parser can instantly load the buffer contents. I can still try to add a dummy delegate to allow the parsing to pass.

android_test/test.cc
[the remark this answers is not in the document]
dj2: Right was thinking it was an AmberOptions object. No need to create a delegate here.

src/amberscript/parser_buffer_test.cc
```diff
@@ -1090,17 +1118,14 @@ TEST_F(AmberScriptParserTest, BufferMissingDataFilePng) {
 TEST_F(AmberScriptParserTest, BufferDataFilePng) {
   std::string in = "BUFFER my_buffer FORMAT R8G8B8A8_UNORM FILE PNG foo.png";
 
-  Parser parser;
+  DummyDelegate delegate;
+  Parser parser(&delegate);
   Result r = parser.Parse(in);
-  ASSERT_TRUE(r.IsSuccess()) << r.Error();
-
+  ASSERT_TRUE(r.IsSuccess());
```
dj2, line 1124: Can we update the `DummyDelegate` to store a flag when the load is called and verify that we call into `LoadBufferData` for each one of these?

asuonpaa: The member function is const, so I cannot store anything in the delegate itself. What I did was store the file type in buffer data.

PR #843, opened 2020-04-21
Added DEPTH support for Amber Script.

asuonpaa:


src/vulkan/graphics_pipeline.cc
```diff
@@ -113,9 +113,10 @@ VkCompareOp ToVkCompareOp(CompareOp op) {
       return VK_COMPARE_OP_GREATER_OR_EQUAL;
     case CompareOp::kAlways:
       return VK_COMPARE_OP_ALWAYS;
+    default:
```
dj2, line 116: We've seen issues on compilers where there is no return even if the switch is fully covered. Please move this back out of the switch.

asuonpaa: But then I'm getting an error of not covering the new kUnknown value in a switch. Should I do both then?

dj2: Yea, add
```
case CompareOp::kUnknown:
  break;
```

and let the external return handle it.

alan-baker:
FYI, The linux bot failures will need #844 to work.

PR #846, opened 2020-04-22
Update cmake

dj2:


kokoro/scripts/linux/build.sh
```diff
@@ -53,12 +53,12 @@ fi
 echo y | sudo apt-get purge --auto-remove cmake
 
 # Installing the 3.10.2 version
```
ben-clayton, line 55: Is this comment still correct?

dj2: nope, removed

dj2:
PTAL. Updating this as Swiftshader is moving to a later CMake so want to make sure our bots are ready for the change.

kokoro/scripts/linux/build_dawn.sh
```diff
@@ -43,12 +43,13 @@ fi
 echo y | sudo apt-get purge --auto-remove cmake
 
 # Installing the 3.10.2 version
```
alan-baker, line 45: Bad comment now?

dj2: done

PR #847, opened 2020-04-24
Added STENCIL support for Amber Script.

asuonpaa:


src/amberscript/parser.cc
```diff
@@ -1307,6 +1330,150 @@ Result Parser::ParsePipelineDepth(Pipeline* pipeline) {
...
+Result Parser::ParsePipelineStencil(Pipeline* pipeline) {
+  auto token = tokenizer_->NextToken();
+  if (!token->IsIdentifier())
+    return Result("STENCIL missing face");
+
+  bool setFront = false;
+  bool setBack = false;
+
+  if (token->AsString() == "front") {
+    setFront = true;
```
dj2, line 1342: Should these be an enum similar to the compare_op and stencil_op?

src/amberscript/parser.cc
```diff
@@ -1307,6 +1330,150 @@ Result Parser::ParsePipelineDepth(Pipeline* pipeline) {
...
+      else
+        return Result("STENCIL invalid value for TEST: " + token->AsString());
+    } else if (token->AsString() == "FAIL_OP") {
+      token = tokenizer_->NextToken();
+
+      if (!token->IsIdentifier())
+        return Result("STENCIL invalid value for FAIL_OP");
+
+      StencilOp stencil_op = StrToStencilOp(token->AsString());
+      if (stencil_op != StencilOp::kUnknown) {
```
dj2, line 1382: For this and the below let's handle the error case first, then we don't need the else.

```
if (stencil_op == StencilOp::kUnknown) {
  return Result ....;
}
```

src/amberscript/parser.cc
asuonpaa: I don't think so, because one or both booleans can be true. With enums we would later need to do this:

```
if (face == kFront || face == kFrontAndBack)
   doFront()
else if (face == kBack || face == kFrontAndBack)
   doBack()
```

src/amberscript/parser.cc
asuonpaa: sure

PR #853, opened 2020-04-30
Moved Delegate destructor definition into header file

asuonpaa:
Integrating the latest Amber version into CTS gave linker errors when having Delegate destructor defined inside amber.cc.

### DIFF
--- a/include/amber/amber.h
+++ b/include/amber/amber.h
@@ -85,7 +85,7 @@ enum class BufferDataFileType : int8_t {
 /// Delegate class for various hook functions
 class Delegate {
  public:
-  virtual ~Delegate();
+  virtual ~Delegate() = default;
 
   /// Log the given message
   virtual void Log(const std::string& message) = 0;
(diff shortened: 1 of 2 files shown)


dj2:
This causes compile errors on our bots:

```
In file included from ../samples/amber.cc:15:
../include/amber/amber.h:86:7: error: 'Delegate' has no out-of-line virtual method definitions; its vtable will be emitted in every translation unit [-Werror,-Wweak-vtables]
class Delegate {
```

What's the error the CTS is producing? The delegate has been there for a long time without issue.

asuonpaa:
> This causes compile errors on our bots:
> 
> ```
> In file included from ../samples/amber.cc:15:
> ../include/amber/amber.h:86:7: error: 'Delegate' has no out-of-line virtual method definitions; its vtable will be emitted in every translation unit [-Werror,-Wweak-vtables]
> class Delegate {
> ```
> 
> What's the error the CTS is producing? The delegate has been there for a long time without issue.

Here's the error:

```
/usr/bin/ld: external/vulkancts/modules/vulkan/amber/libdeqp-vk-amber.a(vktAmberTestCase.cpp.o):(.rodata+0x1b0): undefined reference to `typeinfo for amber::Delegate'
```

Until now CTS has passed a null pointer as delegate.

PR #855, opened 2020-05-06
Move delegate destructor to header file.

dj2:
This CL moves the delegate virtual destructor to the header fix to fix
translation unit issues.


dj2:
Didn't work.

PR #860, opened 2020-05-13
Added support for specifying target environment per shader

asuonpaa:


docs/amber_script.md
```diff
@@ -155,6 +155,10 @@ types, but in that case must only provide a single shader type in the module.
  * `SPIRV-HEX` (decoded straight to SPIR-V)
  * `OPENCL-C` (with clspv)
 
+Specifying `{target_env}` is optional and can be used to select a target
+SPIR-V environment. This overwrites the default environment (spv1.0) or
+one specified in command line parameters.
+
```
paulthomson, line 161: I suggest modifying the bullet point above:

"`SPIRV-ASM` (with spirv-as)"
->
"`SPIRV-ASM` (with spirv-as; specifying `TARGET_ENV` is _highly recommended_ in this case, as explained below)"

And then changing the text to:

#### Target environment

Specifying `TARGET_ENV {target_env}` is optional and can be used to select a target
SPIR-V environment. For example:

* `spv1.0`
* `spv1.5`
* `vulkan1.0`
* `vulkan1.2`

Check the help text of the corresponding tool (e.g. spirv-as, glslangValidator) for the full list. The `SPIRV-HEX` shader format is not affected by the target environment.

The specified target environment for the shader overrides the default (`spv1.0`) or
the one specified on the command line.

Specifying the target environment when using the `SPIRV-ASM` shader format
is _highly recommended_, otherwise the SPIR-V version of the final SPIR-V binary shader that is passed to the
graphics device might not be what you expect.
Typically, SPIR-V assembly text will contain a comment near the beginning similar
to `; Version: 1.0` but this is _ignored_ by the spirv-as assembler.
Thus, you should specify the equivalent target environment (e.g. `spv1.0`)
in the `SHADER` command.

Specifying the target environment for other shader formats depends on whether
you want to vary the final SPIR-V shader binary based on the target environment
specified on the command line. For example, you could write one AmberScript file
that contains a GLSL shader without specifying a target environment.
You could then run the AmberScript file several times with different
target environments specified on the command line
(`spv1.0`, `spv1.1`, `spv1.2`, etc.) to test the different SPIR-V shader variants.

src/amberscript/parser.cc
```diff
@@ -475,6 +475,15 @@ Result Parser::ParseShaderBlock() {
...
 
+    token = tokenizer_->PeekNextToken();
+    if (token->IsIdentifier() && token->AsString() == "TARGET_ENV") {
+      tokenizer_->NextToken();
+      token = tokenizer_->NextToken();
+      if (!token->IsIdentifier() && !token->IsString())
+        return Result("expected target environment after TARGET_ENV");
+      shader->SetTargetEnv(token->AsString());
+    }
+
```
paulthomson, line 486: Actually, I think we should not allow TARGET_ENV when using a PASSTHROUGH shader. Instead, we should call shader->SetTargetEnv(...) and hard code the target environment, probably to "spv1.0" (because that is what the kPassThroughShader text represents). It turns out that mixing SPIR-V versions in a single pipeline IS allowed, so this should be fine.

src/amberscript/parser.cc
```diff
@@ -489,6 +498,15 @@ Result Parser::ParseShaderBlock() {
...
 
+  token = tokenizer_->PeekNextToken();
+  if (token->IsIdentifier() && token->AsString() == "TARGET_ENV") {
+    tokenizer_->NextToken();
+    token = tokenizer_->NextToken();
+    if (!token->IsIdentifier() && !token->IsString())
+      return Result("expected target environment after TARGET_ENV");
+    shader->SetTargetEnv(token->AsString());
+  }
+
```
paulthomson, line 509: I think we should probably give an error if we see TARGET_ENV and the shader type is SPIRV-HEX, as the target env parameter won't do anything. We should also change the documentation in this case. @dj2

src/amberscript/parser_shader_test.cc
```diff
@@ -122,6 +122,34 @@ TEST_F(AmberScriptParserTest, ShaderPassThroughExtraParameters) {
   EXPECT_EQ("1: extra parameters after SHADER PASSTHROUGH: INVALID", r.Error());
 }
 
+TEST_F(AmberScriptParserTest, ShaderPassThroughTargetEnv) {
```
paulthomson, line 125: I guess we will have to change this into negative test, or remove it.

src/amberscript/parser.cc
paulthomson: Actually, this is not true if you want to optimize the shader (using spirv-opt). So maybe ignore this or see what Dan thinks we should do.

paulthomson: Actually, I am kind of wrong again; spirv-opt with target-env spv1.5 will not change a shader to spv1.5; the SPIR-V version remains unchanged. But we could still allow TARGET_ENV and SPIRV-HEX.

docs/amber_script.md
```diff
@@ -121,7 +121,7 @@ Shader using source from `VIRTUAL_FILE`:
 ```groovy
 # Creates a shader of |shader_type| with the given |shader_name|. The shader
 # will be of |shader_format|. The shader will use the virtual file with |path|.
-SHADER {shader_type} {shader_name} {shader_format} VIRTUAL_FILE {path}
+SHADER {shader_type} {shader_name} {shader_format} TARGET_ENV {target_env} VIRTUAL_FILE {path}
```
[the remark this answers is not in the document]
asuonpaa: done

docs/amber_script.md
asuonpaa: done

src/amberscript/parser.cc
asuonpaa: done

src/amberscript/parser.cc
asuonpaa: ok, so there's nothing I should change here, right?

src/amberscript/parser_shader_test.cc
asuonpaa: done

src/amberscript/parser_shader_test.cc
```diff
@@ -235,6 +235,41 @@ END)";
...
+  gl_FragColor = vec3(2, 3, 4);
+}
+END)";
+
+  Parser parser;
+  Result r = parser.Parse(in);
+  ASSERT_FALSE(r.IsSuccess());
+  EXPECT_EQ("3: expected target environment after TARGET_ENV", r.Error());
+}
+
```
dj2, line 272: Add a test where `TARGET_ENV 12345` is specified?

asuonpaa: The parser doesn't validate the environment string since the list might change when the tools are updated. Do we still want to add a list of supported values to the parser?

dj2: I believe the target env is always a string correct? Was just looking for something that wasn't a string to make sure we had that validation.

asuonpaa: Ok, I see. Done.

Issue #861, opened 2020-05-14
Add support for uniform/storage texel buffers


asuonpaa:
These would map to Vulkan descriptor types VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER and VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER.

PR #862, opened 2020-05-15
Add support for uniform/storage texel buffers

asuonpaa:
Fixes https://github.com/google/amber/issues/861

### DIFF
--- a/docs/amber_script.md
+++ b/docs/amber_script.md
@@ -460,6 +460,8 @@ The following commands are all specified within the `PIPELINE` command.
 #### Buffer Types
  * `uniform`
  * `storage`
+ * `uniform_texel_buffer`
+ * `storage_texel_buffer`
 
 TODO(dsinclair): Sync the BufferTypes with the list of Vulkan Descriptor types.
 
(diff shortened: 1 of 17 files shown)


src/vulkan/buffer_descriptor.cc
```diff
@@ -47,12 +47,23 @@ Result BufferDescriptor::CreateResourceIfNeeded() {
...
+  VkBufferUsageFlags flags =
+      VK_BUFFER_USAGE_TRANSFER_SRC_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT;
+  if (IsUniformBuffer()) {
+    flags |= VK_BUFFER_USAGE_UNIFORM_BUFFER_BIT;
+  } else if (IsStorageBuffer()) {
+    flags |= VK_BUFFER_USAGE_STORAGE_BUFFER_BIT;
+  } else if (IsUniformTexelBuffer()) {
+    flags |= VK_BUFFER_USAGE_UNIFORM_TEXEL_BUFFER_BIT;
+  } else {
+    assert(IsStorageTexelBuffer());
```
dj2, line 62: Instead of `assert` this should return a failed Result.

src/vulkan/transfer_buffer.cc
```diff
@@ -19,10 +19,20 @@
...
+                               uint32_t size_in_bytes,
+                               Format* format)
+    : Resource(device, size_in_bytes) {
+  if (format)
+    format_ = device->GetVkFormat(*format);
+}
 
 TransferBuffer::~TransferBuffer() {
+  if (view_ != VK_NULL_HANDLE) {
+    device_->GetPtrs()->vkDestroyBufferView(device_->GetVkDevice(), view_,
```
dj2, line 32: I don't think you need the `if()` check. From the spec:

```
If bufferView is not VK_NULL_HANDLE, bufferView must be a valid VkBufferView handle
```

which sounds like you can just pass in the VK_NULL_HANDLE

src/vulkan/transfer_buffer.h
```diff
@@ -30,10 +31,13 @@ class Device;
 /// Wrapper around a Vulkan VkBuffer object.
 class TransferBuffer : public Resource {
  public:
-  TransferBuffer(Device* device, uint32_t size_in_bytes);
+  TransferBuffer(Device* device,
+                 uint32_t size_in_bytes,
+                 Format* format = nullptr);
```
dj2, line 36: nit: Remove the default value and pass `nullptr` at the various callers. I find default params makes refactoring a lot harder.

docs/amber_script.md
```diff
@@ -492,17 +494,17 @@ contain image attachment content, depth/stencil content, uniform buffers, etc.
   # Attach |buffer_name| as a storage image. The MIP level will have a base
   # value of |level|.
   BIND BUFFER {buffer_name} AS storage_image \
-      [ BASE_MIP_LEVEL _level_ (default 0) ]
+      DESCRIPTOR_SET _id_ BINDING _id_ [ BASE_MIP_LEVEL _level_ (default 0) ]
```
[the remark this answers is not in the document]
asuonpaa: It means the documentation was broken.

src/vulkan/buffer_descriptor.cc
asuonpaa: done

src/vulkan/transfer_buffer.cc
asuonpaa: Same applies to vkDestroyBuffer so I've removed the null check from there too.

src/vulkan/transfer_buffer.h
asuonpaa: done

Issue #870, opened 2020-06-03
Addition to instanced draw (amber script)


ilkkasaa:
For instanced drawing it would be very useful to be able to define vertex input rate for vertex buffer bindings: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkVertexInputRate.html

I propose the amber script syntax to be something like: 
```groovy
PIPELINE ...
  VERTEX_DATA {buffer_name} LOCATION _val_ RATE {vertex|instance} (default vertex)
...
END 
```

asuonpaa:
This is not as trivial as it seems. The input rate is defined per binding and not per location. Currently Amber has only binding 0 enabled and all locations has this binding. There are multiple TODOs in the code saying we should support multiple bindings for vertex data. I suppose this serves as a motivation to do that.

asuonpaa:
Amber interleaves the vertex data automatically to a single Vulkan buffer, which is probably the reason only one binding is used. This means currently the user has no choice of manually adding interleaved data to a single Amber buffer and the input rate cannot be controlled per Amber buffer.

I believe a better option is to use a separate binding for each Amber buffer bound as vertex data. This change would happen under the hood with no changes needed to Amber script. We could also allow manual data interleaving in an Amber buffer when the same buffer is bound to multiple locations in the same pipeline. Amber buffers would restrict this to a single format so I don't know if that would be useful in many cases, but it's better than nothing.

With these changes the original syntax suggestion for RATE would be possible to implement.

asuonpaa:
PR here: https://github.com/google/amber/pull/871
Interleaving vertex data inside Amber buffers can be done as a separate task if needed.

PR #871, opened 2020-06-06
Add input rate support for Amber script

asuonpaa:
Also modified Vulkan engine to use separate bindings for each vertex buffer as input rate is set per binding.

Fixes https://github.com/google/amber/issues/870

### DIFF
--- a/src/vulkan/graphics_pipeline.cc
+++ b/src/vulkan/graphics_pipeline.cc
@@ -546,31 +546,25 @@ Result GraphicsPipeline::CreateVkGraphicsPipeline(
         "null");
   }
 
+  std::vector<VkVertexInputBindingDescription> vertex_bindings;
+  std::vector<VkVertexInputAttributeDescription> vertex_attribs;
+  if (vertex_buffer != nullptr) {
+    vertex_bindings = vertex_buffer->GetVkVertexInputBinding();
+    vertex_attribs = vertex_buffer->GetVkVertexInputAttr();
+  }
+
   VkPipelineVertexInputStateCreateInfo vertex_input_info =
       VkPipelineVertexInputStateCreateInfo();
   vertex_input_info.sType =
       VK_STRUCTURE_TYPE_PIPELINE_VERTEX_INPUT_STATE_CREATE_INFO;
-  vertex_input_info.vertexBindingDescriptionCount = 1;
-
-  VkVertexInputBindingDescription vertex_binding_desc =
-      VkVertexInputBindingDescription();
-  if (vertex_buffer != nullptr) {
-    vertex_binding_desc = vertex_buffer->GetVkVertexInputBinding();
-    const auto& vertex_attr_desc = vertex_buffer->GetVkVertexInputAttr();
-
-    vertex_input_info.pVertexBindingDescriptions = &vertex_binding_desc;
-    vertex_input_info.vertexAttributeDescriptionCount =
-        static_cast<uint32_t>(vertex_attr_desc.size());
-    vertex_input_info.pVertexAttributeDescriptions = vertex_attr_desc.data();
-  } else {
-    vertex_binding_desc.binding = 0;
-    vertex_binding_desc.stride = 0;
-    vertex_binding_desc.inputRate = VK_VERTEX_INPUT_RATE_VERTEX;
-
-    vertex_input_info.pVertexBindingDescriptions = &vertex_binding_desc;
-    vertex_input_info.vertexAttributeDescriptionCount = 0;
-    vertex_input_info.pVertexAttributeDescriptions = nullptr;
-  }
+  vertex_input_info.vertexBindingDescriptionCount =
+      static_cast<uint32_t>(vertex_bindings.size());
+  vertex_input_info.pVertexBindingDescriptions =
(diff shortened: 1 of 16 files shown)


src/vulkan/vertex_buffer.cc
```diff
@@ -23,70 +23,78 @@
 
 namespace amber {
 namespace vulkan {
+namespace {
+
+VkVertexInputRate GetVkInputRate(InputRate rate) {
+  if (rate == InputRate::kVertex)
+    return VK_VERTEX_INPUT_RATE_VERTEX;
+  else
+    return VK_VERTEX_INPUT_RATE_INSTANCE;
```
dj2, line 32: nit: `return rate == InputRate::kVertex ? VK_VERTEX_INPUT_RATE_VERTEX : VK_VERTEX_INPUT_RATE_INSTANCE;`

src/vulkan/graphics_pipeline.cc
```diff
@@ -546,31 +546,25 @@ Result GraphicsPipeline::CreateVkGraphicsPipeline(
...
-    const auto& vertex_attr_desc = vertex_buffer->GetVkVertexInputAttr();
-
-    vertex_input_info.pVertexBindingDescriptions = &vertex_binding_desc;
-    vertex_input_info.vertexAttributeDescriptionCount =
-        static_cast<uint32_t>(vertex_attr_desc.size());
-    vertex_input_info.pVertexAttributeDescriptions = vertex_attr_desc.data();
-  } else {
-    vertex_binding_desc.binding = 0;
-    vertex_binding_desc.stride = 0;
-    vertex_binding_desc.inputRate = VK_VERTEX_INPUT_RATE_VERTEX;
```
dj2, line 568: These three values seem to be gone now, does that mean they're the defaults?

src/vulkan/vertex_buffer.cc
```diff
@@ -23,70 +23,78 @@
...
 }
 
 Result VertexBuffer::SendVertexData(CommandBuffer* command) {
   if (!is_vertex_data_pending_)
     return Result("Vulkan::Vertices data was already sent");
 
-  const uint32_t n_vertices = GetVertexCount();
-  if (n_vertices == 0)
-    return Result("Vulkan::Data for VertexBuffer is empty");
+  // Non-empty transfer_buffers_ means we are in a unit test.
```
dj2, line 76: Why does this mean unit test? We shouldn't have code in here which is unit test specific.

src/vulkan/vertex_buffer.cc
```diff
@@ -23,70 +23,78 @@
...
+      std::memcpy(transfer_buffers_.back()->HostAccessibleMemoryPtr(),
+                  buf->GetValues<void>(), bytes);
+      transfer_buffers_.back()->CopyToDevice(command);
 
-  FillVertexBufferWithData(command);
+      if (!r.IsSuccess())
+        return r;
+    }
+  } else {
+    // Only used in vertex buffer unit tests.
```
dj2, line 94: ditto

src/vulkan/vertex_buffer.cc
```diff
@@ -23,70 +23,78 @@
...
   vertex_attr_desc_.back().location = location;
-  vertex_attr_desc_.back().offset = stride_in_bytes_;
+  vertex_attr_desc_.back().offset = 0u;
   vertex_attr_desc_.back().format = device_->GetVkFormat(*format);
 
-  stride_in_bytes_ += format->SizeInBytes();
-  data_.push_back(buffer);
-}
-
-Result VertexBuffer::FillVertexBufferWithData(CommandBuffer* command) {
```
dj2, line 45: So, why can this just go away? If we were interleaving buffers previously, doesn't that mean there are tests out there which many require this functionality?

src/vulkan/graphics_pipeline.cc
asuonpaa: Previously a default binding was created when there was no vertex buffer present. It's not necessary so now we just set the size of bindings to zero instead and don't provide this default data (which is not used anyway) at all.

src/vulkan/vertex_buffer.cc
asuonpaa: I totally agree. The same structure was there before and I just added comment to clarify there was unit test specific code present.

I think it's trying to avoid using a device object which is NULL when in unit test. We had a similar issue before and we used a dummy object in unit test instead. I'll try to use similar solution here.

src/vulkan/vertex_buffer.cc
asuonpaa: The interleaving was all happening under the hood (in Vulkan engine) and there was no benefit of doing that. Using separate buffers works just as fine. And this is not visible to shaders either. And as a bonus the code got a bit more simple.

src/vulkan/vertex_buffer.cc
asuonpaa: done

src/vulkan/vertex_buffer.cc
asuonpaa: Managed to add a dummy device class and get rid of test specific code in vertex buffer class.

PR #880, opened 2020-06-29
Add support for Vulkan 1.2 feature structures

asuonpaa:
Vulkan 1.2 added support for defining non-core physical device features using VkPhysicalDeviceVulkan11Features and VkPhysicalDeviceVulkan12Features structures. Previously several different feature structures were linked to vkGetPhysicalDeviceFeatures2. This change updates Amber feature checking to also go through these structs if present.

### DIFF
--- a/src/vulkan/device.cc
+++ b/src/vulkan/device.cc
@@ -394,6 +394,7 @@ Result Device::Initialize(
     PFN_vkGetInstanceProcAddr getInstanceProcAddr,
     Delegate* delegate,
     const std::vector<std::string>& required_features,
+    const std::vector<std::string>& required_instance_extensions,
     const std::vector<std::string>& required_device_extensions,
     const VkPhysicalDeviceFeatures& available_features,
     const VkPhysicalDeviceFeatures2KHR& available_features2,
@@ -628,11 +629,24 @@ Result Device::Initialize(
   ptrs_.vkGetPhysicalDeviceMemoryProperties(physical_device_,
                                             &physical_memory_properties_);
 
+  bool use_physical_device_features_2 = false;
+  for (auto& ext : required_instance_extensions) {
+    if (ext == "VK_KHR_get_physical_device_properties2")
+      use_physical_device_features_2 = true;
+  }
+
   subgroup_size_control_properties_ = {};
   const bool needs_subgroup_size_control =
       std::find(required_features.begin(), required_features.end(),
                 kSubgroupSizeControl) != required_features.end();
+
   if (needs_subgroup_size_control) {
+    if (!use_physical_device_features_2) {
+      return Result(
+          "Vulkan: Device::Initialize subgroup size control feature also "
+          "requires VK_KHR_get_physical_device_properties2");
+    }
+
     PFN_vkGetPhysicalDeviceProperties2KHR vkGetPhysicalDeviceProperties2KHR =
         reinterpret_cast<PFN_vkGetPhysicalDeviceProperties2KHR>(
             getInstanceProcAddr(instance_,
--- a/src/vulkan/device.cc
+++ b/src/vulkan/device.cc
@@ -425,36 +425,47 @@ Result Device::Initialize(
   void* ptr = available_features2.pNext;
   while (ptr != nullptr) {
     BaseOutStructure* s = static_cast<BaseOutStructure*>(ptr);
-    if (s->sType ==
-        VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VARIABLE_POINTER_FEATURES_KHR) {
-      var_ptrs = static_cast<VkPhysicalDeviceVariablePointerFeaturesKHR*>(ptr);
-    } else if (s->sType ==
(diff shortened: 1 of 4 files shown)


src/vulkan/device.cc
```diff
@@ -403,121 +402,134 @@ Result Device::Initialize(
...
 
-    std::vector<std::string> required_features1;
-    for (const auto& feature : required_features) {
-      // No dot means this is a features1 feature.
-      if (feature.find_first_of('.') == std::string::npos) {
-        required_features1.push_back(feature);
-        continue;
-      }
+  // Compare the available additional (non-core) features against the
+  // requirements.
```
paulthomson, line 457: Could you add another comment that is similar to your PR description. E.g.

Vulkan 1.2 added support for defining non-core physical device features using VkPhysicalDeviceVulkan11Features and VkPhysicalDeviceVulkan12Features structures. If |vulkan11_ptrs| and/or |vulkan12_ptrs| are null, we must check for features using the old approach (by checking across various feature structs); otherwise, we can check features via the new structs.

src/vulkan/device.cc
```diff
@@ -403,121 +402,134 @@ Result Device::Initialize(
...
+  VkPhysicalDeviceVulkan11Features* vulkan11_ptrs = nullptr;
+  VkPhysicalDeviceVulkan12Features* vulkan12_ptrs = nullptr;
+  VkPhysicalDeviceSubgroupSizeControlFeaturesEXT*
+      subgroup_size_control_features = nullptr;
+  void* ptr = available_features2.pNext;
+  while (ptr != nullptr) {
+    BaseOutStructure* s = static_cast<BaseOutStructure*>(ptr);
+    if (s->sType ==
+        VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VARIABLE_POINTER_FEATURES_KHR) {
+      var_ptrs = static_cast<VkPhysicalDeviceVariablePointerFeaturesKHR*>(ptr);
```
paulthomson, line 430: I know this is not due to your change, but can we make this a switch?

src/vulkan/device.cc
```diff
@@ -578,7 +621,7 @@ Result Device::Initialize(
   const bool needs_subgroup_size_control =
       std::find(required_features.begin(), required_features.end(),
                 kSubgroupSizeControl) != required_features.end();
-  if (needs_subgroup_size_control && use_physical_device_features_2) {
+  if (needs_subgroup_size_control) {
     PFN_vkGetPhysicalDeviceProperties2KHR vkGetPhysicalDeviceProperties2KHR =
```
paulthomson, line 625: Can you explain why you were able to remove this?

src/vulkan/device.cc
asuonpaa: done

src/vulkan/device.cc
asuonpaa: sure

src/vulkan/device.cc
asuonpaa: Previously the usage of vkGetPhysicalDeviceFeatures2 was determined by an extension. But now that it's promoted to core I don't think we should require the scripts to enable the extension. And since we check for both feature structs we don't really need to check this. After all we don't call the vkGetPhysicalDeviceFeatures2 here.

dj2:
If I have a vulkan 1.0 driver Amber will still work after this yes? We've tried to make Amber build and run on 1.0 and up.

asuonpaa:
> If I have a vulkan 1.0 driver Amber will still work after this yes? We've tried to make Amber build and run on 1.0 and up.

Yes, It still supports the old way of specifying the features. I have 1.1 on my Linux and everything's fine there even though it doesn't support the new structs. And when testing a more recent driver with 1.2 support on my Windows it executes the new code paths.

src/vulkan/device.cc
paulthomson: Promoted to core? But in Vulkan 1 it will still be an extension and so `getInstanceProcAddr` might return null?

asuonpaa: You're right. I need to restore the check for the extension and also query the Vulkan API version and use the core function in case of Vulkan >= 1.1.

paulthomson: I think you might be able to always use the extension, even if it is core in the Vulkan being used.

asuonpaa: Which means the script needs to have that extension enabled if it wants to be compatible with Vulkan 1.0. 

Does the spec somewhere mention you have to still support the extensions after being promoted to core? If not, it might be safest to just check the version as it's quite easy to do.

paulthomson: The notice of promotion is like this: "All functionality in this extension is included in core Vulkan 1.1, with the KHR suffix omitted. The original type, enum and command names are still available as aliases of the core functionality."

I think Amber should (and already does, I think) implicitly add required instance extensions (like `VK_KHR_get_physical_device_properties2`) when it knows they will be required due to other required extensions.

asuonpaa: I've now restored the checking of that extension.

asuonpaa:
> LGTM! Thanks.

Thanks. Still needs to be merged though...

PR #881, opened 2020-07-02
Add support for FreeBSD

amshafer:
Create platform definitions required for building on FreeBSD.


amshafer:
@googlebot I fixed it.

PR #885, opened 2020-07-16
Fix vkscript buffers being added twice

ilkkasaa:
Some of the buffers were added twice if  vkscript format was used.

This patch prevents same buffers to be added multiple times.

### DIFF
--- a/src/vulkan/pipeline.cc
+++ b/src/vulkan/pipeline.cc
@@ -338,7 +338,12 @@ Result Pipeline::AddBufferDescriptor(const BufferCommand* cmd) {
           "Descriptors bound to the same binding needs to have matching "
           "descriptor types");
     }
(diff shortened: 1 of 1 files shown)


src/vulkan/pipeline.cc
```diff
@@ -338,7 +338,12 @@ Result Pipeline::AddBufferDescriptor(const BufferCommand* cmd) {
           "Descriptors bound to the same binding needs to have matching "
           "descriptor types");
     }
-    desc->AsBufferBackedDescriptor()->AddAmberBuffer(cmd->GetBuffer());
+    // Check that the buffer is not added already.
```
dj2, line 341: Can we add a unit test for this?

paulthomson:
I think the other PR is preferable.

PR #888, opened 2020-07-16
Fix vkscript buffers being added twice

ilkkasaa:
Some of the buffers were added twice if vkscript format was used.

This patch removes unnecessary calls to Pipeline::AddBufferDescriptor()
function and therefore prevents buffers to be added more than once.

### DIFF
--- a/src/vulkan/pipeline_test.cc
+++ b/src/vulkan/pipeline_test.cc
@@ -1,0 +1,56 @@
+// Copyright 2020 The Amber Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


ilkkasaa:
This is alternative and maybe better way of fixing the same problem as https://github.com/google/amber/pull/885

src/vulkan/pipeline_test.cc
```diff
@@ -0,0 +1,56 @@
...
+  ASSERT_FALSE(r.IsSuccess());
+}
+
+TEST_F(VulkanPipelineTest, AddBufferDescriptorAddBufferTwice) {
+  amber::Pipeline amber_pipeline(PipelineType::kCompute);
+  std::vector<VkPipelineShaderStageCreateInfo> create_infos;
+  ComputePipeline pipeline(nullptr, 0, create_infos);
+
+  auto cmd = MakeUnique<BufferCommand>(BufferCommand::BufferType::kPushConstant,
+                                       &amber_pipeline);
```
paulthomson, line 44: Did you miss this?

paulthomson: Removed.

Issue #890, opened 2020-07-19
Python warning in tools/git_sync_deps


hevrard:
Python3 emits these warnings when running `./tools/git_sync_deps`:

```
$ python3 tools/git-sync-deps                     
tools/git-sync-deps:231: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if not with_clspv and directory is 'third_party/clspv':                      
tools/git-sync-deps:234: SyntaxWarning: "is" with a literal. Did you mean "=="?       
  if not with_clspv and directory is 'third_party/clspv-llvm':                 
tools/git-sync-deps:237: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if not with_dxc and directory is 'third_party/dxc':
tools/git-sync-deps:240: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if not with_swiftshader and directory is 'third_party/swiftshader':
```

These warnings come from a change in Python 3.8, see second item of https://docs.python.org/3.8/whatsnew/3.8.html#changes-in-python-behavior:

> The compiler now produces a SyntaxWarning when identity checks (is and is not) are used with certain types of literals (e.g.
> strings, numbers). These can often work by accident in CPython, but are not guaranteed by the language spec. The warning
> advises users to use equality tests (== and !=) instead. (Contributed by Serhiy Storchaka in bpo-34850.)

PR #891, opened 2020-07-19
Fix Python warnings related to 'is' operator

hevrard:
Replace the 'is' operators with '==' when there is a literal among the
arguments. The Python 'is' operator is not meant to be used with
literals, and the code emitted warnings like:

tools/git-sync-deps:231: SyntaxWarning: "is" with a literal. Did you mean "=="?

For the origin of this warning in Python, see
https://bugs.python.org/issue34850

Fix: #890


paulthomson:
Looks like the CI has been relying on this bug. The kokoro scripts need to be updated. `--use-dxc` and `--use-swiftshader` should be `--with-dxc` and `--with-swiftshader`.

hevrard:
@paulthomson two CI tracks are timing out, is that a flake?

paulthomson:
Yes. At least, I don't think that is your fault.

Issue #725, opened 2019-11-27
SEGV on swiftshader for multiple_ssbo_update_with_graphics_pipeline.vkscript


dj2:
When running under swiftshader the multiple_ssbo_update_with_graphics_pipeline.vkscript test crashes.

`Thread 1 "amber" received signal SIGSEGV, Segmentation fault.`

```
#11 0x00007ffff0d75655 in std::unordered_map<VkDeviceMemory_T*, std::shared_ptr<ObjectUseData>, std::hash<VkDeviceMemory_T*>, std::equal_to<VkDeviceMemory_T*>, std::allocator<std::pair<VkDeviceMemory_T* const, std::shared_ptr<ObjectUseData> > > >::erase (this=0x555556abcbd0, __x=@0x7fffffffc970: 0x100000000010) at /usr/include/c++/8/bits/unordered_map.h:815
#12 0x00007ffff0d6a854 in vl_concurrent_unordered_map<VkDeviceMemory_T*, std::shared_ptr<ObjectUseData>, 6, std::hash<VkDeviceMemory_T*> >::erase (this=0x555556abc818, key=@0x7fffffffc970: 0x100000000010)
    at ../../third_party/vulkan-validationlayers/layers/vk_layer_utils.h:255
#13 0x00007ffff0d5a04e in counter<VkDeviceMemory_T*>::DestroyObject (this=0x555556abc800, object=0x100000000010) at ../../third_party/vulkan-validationlayers/layers/generated/thread_safety.h:139
#14 0x00007ffff0d50f6c in ThreadSafety::DestroyObject (this=0x555556a9af70, object=0x100000000010) at ../../third_party/vulkan-validationlayers/layers/generated/thread_safety.h:475
#15 0x00007ffff0d3fa7f in ThreadSafety::PostCallRecordFreeMemory (this=0x555556a9af70, device=0x555556731ec8, memory=0x100000000010, pAllocator=0x0)
    at ../../third_party/vulkan-validationlayers/layers/generated/thread_safety.cpp:749
#16 0x00007ffff09823a2 in vulkan_layer_chassis::FreeMemory (device=0x555556731ec8, memory=0x100000000010, pAllocator=0x0) at ../../third_party/vulkan-validationlayers/layers/generated/chassis.cpp:1259
#17 0x00007ffff7fc6b98 in vkFreeMemory (device=0x555556731ec8, mem=0x100000000010, pAllocator=0x0) at ../../third_party/vulkan-loader/loader/trampoline.c:887
[...]

PR #893, opened 2020-07-21
Fix tests that had OOB writes for ssbo buffer

ben-clayton:
`compute` was called before fully initializing the `ssbo` buffer.
This lead to writes beyond the end of the buffer, corrupting memory in exciting ways when using the swiftshader driver.

Fixes: #725


ben-clayton:
I've manually cancelled the kokoro builds as they weren't using the upgraded VMs.
Will push again in a bit.

PR #899, opened 2020-08-05
 Switch amber to parse clspv's embedded reflection

alan-baker:
* No longer uses the descriptor map, instead parses the embedded
  reflection instructions directly
* strips reflection instructions after parsing to prevent needing
  VK_KHR_non_semantic_info


src/clspv_helper.cc
```diff
@@ -14,117 +14,267 @@
...
+#include "spirv/unified1/spirv.hpp"
+#include "spirv-tools/libspirv.hpp"
+#include "spirv-tools/optimizer.hpp"
+
+using amber::Pipeline;
+
+namespace {
+
+struct ReflectionHelper {
+  Pipeline::ShaderInfo* shader_info;
```
dj2, line 33: Should these two default to nullptr?

src/clspv_helper.cc
```diff
@@ -14,117 +14,267 @@
...
+      break;
+    } break;
+  }
+
+  return SPV_SUCCESS;
+}
+
+spv_result_t ParseReflection(void* user_data,
+                             const spv_parsed_instruction_t* inst) {
+  ReflectionHelper* helper = reinterpret_cast<ReflectionHelper*>(user_data);
```
dj2, line 206: nit: this could be `auto*`

src/clspv_helper.h
```diff
@@ -18,6 +18,8 @@
 #include <string>
 #include <vector>
 
+#include "spirv-tools/libspirv.h"
+
```
dj2, line 22: nit: no blank line

src/shader_compiler.h
```diff
@@ -19,6 +19,8 @@
 #include <utility>
 #include <vector>
 
+#include "spirv-tools/libspirv.h"
+
```
dj2, line 23: ditto

alan-baker:
https://github.com/microsoft/DirectXShaderCompiler/pull/3074 is necessary to get DXC compatible with the updated SPIRV-Tools

PR #901, opened 2020-08-13
Use native vkGetPhysicalDeviceProperties2 for Vulkan 1.1 and up

asuonpaa:
We cannot assume vkGetPhysicalDeviceProperties2KHR is exposed when the native version is available. This change uses the native
function if available and searches for the extension function only when running Vulkan 1.0.


paulthomson:
@googlebot I fixed it.

PR #906, opened 2020-08-24
Handle bad integer args gracefully

dneto0:
Replace uses of std::stoi with std::sscanf.
When parsing fails, std::stoi throws an invalid-argument exception,
which is not caught.  The standard library then terminates with an abort
about the uncaught exception.  That's unhelpful to users.

Instead, use std::sscanf gracefully handle the bad user input.

### DIFF
--- a/samples/amber.cc
+++ b/samples/amber.cc
@@ -106,30 +106,26 @@ const char kUsage[] = R"(Usage: amber [options] SCRIPT [SCRIPTS...]
 )";
 
 // Parses a decimal integer from the given string, and writes it to |retval|.
-// Returns 1 if the parse was successful and consumed the whole string.
-static int ParseOneInt(const char* str, int* retval) {
+// Returns true if parsing succeeded and consumed the whole string.
+static bool ParseOneInt(const char* str, int* retval) {
   char trailing = 0;
-  return
 #if defined(_MSC_VER)
-      sscanf_s(str, "%d%c", retval, &trailing, 1)
+  return sscanf_s(str, "%d%c", retval, &trailing, 1) == 1;
 #else
-      std::sscanf(str, "%d%c", retval, &trailing)
+  return std::sscanf(str, "%d%c", retval, &trailing) == 1;
 #endif
-          ;
 }
 
 // Parses a decimal integer, then a period (.), then a decimal integer  from the
-// given string, and writes it to |retval|.
-// Returns 2 if the parse was successful and consumed the whole string.
+// given string, and writes it to |retval|.  Returns true if parsing succeeded
+// and consumed the whole string.
 static int ParseIntDotInt(const char* str, int* retval0, int* retval1) {
   char trailing = 0;
-  return
 #if defined(_MSC_VER)
-      sscanf_s(str, "%d.%d%c", retval, retval1, &trailing, 1)
+  return sscanf_s(str, "%d.%d%c", retval0, retval1, &trailing, 1) == 2;
 #else
-      std::sscanf(str, "%d.%d%c", retval0, retval1, &trailing)
+  return std::sscanf(str, "%d.%d%c", retval0, retval1, &trailing) == 2;
 #endif
-          ;
 }
 
 bool ParseArgs(const std::vector<std::string>& args, Options* opts) {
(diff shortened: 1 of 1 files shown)


samples/amber.cc
```diff
@@ -102,6 +105,33 @@ const char kUsage[] = R"(Usage: amber [options] SCRIPT [SCRIPTS...]
   -h                        -- This help text.
 )";
 
+// Parses a decimal integer from the given string, and writes it to |retval|.
+// Returns 1 if the parse was successful and consumed the whole string.
+static int ParseOneInt(const char* str, int* retval) {
+  char trailing = 0;
+  return
+#if defined(_MSC_VER)
+      sscanf_s(str, "%d%c", retval, &trailing, 1)
```
dj2, line 114: I'd rather just see the `returns` put into the if's as it's less confusing to read.

samples/amber.cc
```diff
@@ -102,6 +105,33 @@ const char kUsage[] = R"(Usage: amber [options] SCRIPT [SCRIPTS...]
...
+}
+
+// Parses a decimal integer, then a period (.), then a decimal integer  from the
+// given string, and writes it to |retval|.
+// Returns 2 if the parse was successful and consumed the whole string.
+static int ParseIntDotInt(const char* str, int* retval0, int* retval1) {
+  char trailing = 0;
+  return
+#if defined(_MSC_VER)
+      sscanf_s(str, "%d.%d%c", retval, retval1, &trailing, 1)
```
dj2, line 128: ditto

samples/amber.cc
```diff
@@ -168,7 +198,11 @@ bool ParseArgs(const std::vector<std::string>& args, Options* opts) {
         return false;
       }
 
-      int32_t val = std::stoi(std::string(args[i]));
+      int32_t val = 0;
+      if (1 != ParseOneInt(args[i].c_str(), &val)) {
```
dj2, line 202: Can you flip these conditions, we usually put the `!= 1` last.

dj2: Actually, can we move the comparison into the function and have it return bool? The 1/2 from these is ... confusing

paulthomson:
I think we have used `std::stringstream` in the image_diff tool:

https://github.com/google/amber/blob/main/samples/image_diff.cc#L81

Unsure of the exact trade-offs, but there is at least no need to handle `sscanf` vs `sscanf_s`.

dneto0:
> Unsure of the exact trade-offs, but there is at least no need to handle sscanf vs sscanf_s.

I used std::stringstream in the assembler.  (see https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/util/parse_number.h#L163 which is migrated from its original location from source).  But I found lots of incompatibilities with the c++11 spec.   I was trying to go old-school with the simple C API, but then the Microsoft CRT warnings got in the way.  I'll do one more round with sscanf, which I find eminently more readable for this use case.

dneto0:
Still testing...

dneto0:
Rebased. Ready for review again

samples/amber.cc
```diff
@@ -102,6 +105,29 @@ const char kUsage[] = R"(Usage: amber [options] SCRIPT [SCRIPTS...]
...
+#else
+  return std::sscanf(str, "%d%c", retval, &trailing) == 1;
+#endif
+}
+
+// Parses a decimal integer, then a period (.), then a decimal integer  from the
+// given string, and writes it to |retval|.  Returns true if parsing succeeded
+// and consumed the whole string.
+static int ParseIntDotInt(const char* str, int* retval0, int* retval1) {
+  char trailing = 0;
```
dj2, line 123: What is trailing for? Do we not care if it's parsed or what it parses as? If so, why parse it?

dneto0: We attempt to parse it, and hope for failure.  It will soak up *any* extra text at the end, if it exists.

The return value on sscanf (and sscanf_s) is the number of successfully converted / parsed conversion-specs. So with %d.%d%c we hope for the %d and . and %d to be fulfilled and the %c to not be fulfilled.

Sorry, this is an idiom I'm used to.

PR #909, opened 2020-09-03
[AmberScript] Add support for runtime arrays

asuonpaa:
This CL adds the ability to specify the data_type for an AmberScript buffer to be an array.


paulthomson (approved):
Looks good to me! Could you add to the PR description something about why you changed type parser? Was it incorrectly using assertions instead of failing, and you ended up hitting these assertions when you add some new strings to one of the tests?

asuonpaa:
> Looks good to me! Could you add to the PR description something about why you changed type parser? Was it incorrectly using assertions instead of failing, and you ended up hitting these assertions when you add some new strings to one of the tests?

I just copied everything from Dan's commit. I'll let him explain the changes so I don't make up a false explanation.

Looks like the old asserts now propagate down the stack as return values but I'm not sure where they are handled. I could check that in debugger if needed.

PR #910, opened 2020-09-07
Fix dynamic offset order when calling vkCmdBindDescriptorSets

asuonpaa:
vkCmdBindDescriptorSets expects the dynamic offset values to be sorted by descriptor binding order which might be different from what was provided in AmberScript pipeline block.

### DIFF
--- a/src/vulkan/pipeline.cc
+++ b/src/vulkan/pipeline.cc
@@ -438,20 +438,26 @@ void Pipeline::BindVkDescriptorSets(const VkPipelineLayout& pipeline_layout) {
     if (descriptor_set_info_[i].empty)
       continue;
 
-    std::vector<uint32_t> dynamic_offsets;
-
     // Sort descriptors by binding number to get correct order of dynamic
     // offsets.
-    std::sort(std::begin(descriptor_set_info_[i].descriptors),
-              std::end(descriptor_set_info_[i].descriptors),
-              [](const std::unique_ptr<Descriptor>& a,
-                 const std::unique_ptr<Descriptor>& b) {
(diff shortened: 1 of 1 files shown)


paulthomson (commented):
Nice catch! Could you change your PR description to say `vkCmdBindDescriptorSets` instead of "this function"? It never hurts to be explicit. 

It feels slightly strange to _mutate_ `descriptor_set_info_` when calling `BindVkDescriptorSets`. Maybe we should create a shallow, sorted copy? Or sort the dynamic offsets as early as possible elsewhere?

Is `descriptor_set_info_` guaranteed to be in order?

asuonpaa:
> Nice catch! Could you change your PR description to say `vkCmdBindDescriptorSets` instead of "this function"? It never hurts to be explicit.
> 
> It feels slightly strange to _mutate_ `descriptor_set_info_` when calling `BindVkDescriptorSets`. Maybe we should create a shallow, sorted copy? Or sort the dynamic offsets as early as possible elsewhere?
> 
> Is `descriptor_set_info_` guaranteed to be in order?

We don't want to sort earlier because one of my CTS tests requires the Vulkan engine to match the binding order of the Amber Script in layout creation. 

The descriptor vector being sorted contains unique pointers so we cannot just make a copy. And I didn't feel it was worth changing the type to be shared pointer just to be able to do that. Instead of sorting the descriptors maybe I should just extract the binding number and dynamic offset pairs and sort those. Let me try that next.

paulthomson (approved):
LGTM! I was going to suggest storing descriptors as an ordered set, but actually it might be important to keep the unorderedness of them for reproducing driver bugs.

PR #912, opened 2020-09-10
Crash fix + debugger improvement

ben-clayton:
This PR contains two changes:

### dxc: Use a CComPtr for the include handler to prevent use-after-release crash.

The include handler was constructed and passed down to `IDxcCompiler::Compile()`, but was not referenced by the caller.
Within `IDxcCompiler::Compile()`, the include handler may be referenced, released, and an attempt to be referenced again (using deleted memory).

Use `CComPtr` to keep the COM object alive for the full duration of the compile.

### Pass a unique shader filename down to DXC 

In order for the debugger to be able to fetch the source of a shader that has its source embedded within an amber script file, we need to have a known and unique file name.

The file name was previously generated just before calling into DXC to compile the file, and could easily have multiple shaders with the same name.

Instead we now add a optional file path property to the `amber::Shader`, which is populated by the parser with either the `VIRTUAL_FILE` path, or a generated path from the shader name for embedded shaders.
The parser also adds the shader's source to the `VirtualFileStore` so it can be found by the debugger engine.

If no file path is specified by the parser, dxc_helper defaults to its previous naming scheme.

### DIFF
--- a/src/amberscript/parser.cc
+++ b/src/amberscript/parser.cc
@@ -515,34 +515,40 @@ Result Parser::ParseShaderBlock() {
     if (!token->IsIdentifier() && !token->IsString())
       return Result("expected virtual file path after VIRTUAL_FILE");
 
-    r = ValidateEndOfStatement("SHADER command");
-    if (!r.IsSuccess())
-      return r;
-
     auto path = token->AsString();
 
     std::string data;
     r = script_->GetVirtualFile(path, &data);
-    if (!r.IsSuccess()) {
-      return r;
-    }
-
-    shader->SetData(data);
-  } else {
-    r = ValidateEndOfStatement("SHADER command");
     if (!r.IsSuccess())
       return r;
 
-    std::string data = tokenizer_->ExtractToNext("END");
-    if (data.empty())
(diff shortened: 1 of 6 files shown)


jaebaek (approved):

src/amberscript/parser.cc
```diff
@@ -515,34 +515,40 @@ Result Parser::ParseShaderBlock() {
...
     r = script_->GetVirtualFile(path, &data);
-    if (!r.IsSuccess()) {
+    if (!r.IsSuccess())
       return r;
-    }
 
     shader->SetData(data);
-  } else {
-    r = ValidateEndOfStatement("SHADER command");
+    shader->SetFilePath(path);
```
jaebaek, line 526: nit: could you please add unit tests to check setting the file path?

jaebaek (commented):
nit: could you please add some unit tests to check the parser update?

src/amberscript/parser.cc
ben-clayton: Done

dj2 (approved):

PR #915, opened 2020-10-08
Debugger tests

ben-clayton:


tests/run_tests.py
```diff
@@ -315,6 +317,9 @@ def Run(self):
     parser.add_option('--use-swiftshader',
                       action="store_true", default=False,
                       help='Tells test runner swiftshader is the device')
+    parser.add_option('--test-debugger',
```
dj2, line 320: Does this need to turn on use-swiftshader? What happens if I use debugger but don't use swiftshader?

ben-clayton: If the driver doesn't support debugging, then it'll fail to connect and sensibly error.
Yes, SwiftShader is currently the only driver that implements debugging via DAP, but it's our hope that this will not always be the case.

dj2 (approved):

PR #917, opened 2020-10-26
Fix build on Clang 11

mvainola:
Explicitly specify some destructor overrides to suppress `-Wsuggest-destructor-override` warnings.

This affects building the CTS with Clang 11.

[amber_build_failure_clang_11.txt](https://github.com/google/amber/files/5440683/amber_build_failure_clang_11.txt)


dj2 (approved):
Awesome. Thank you.

PR #918, opened 2020-10-27
Add case for edge case behaviour of packHalf2x16

dneto0:


dj2 (approved):
LGTM, but I'm not sure if this will end up running as part of the CTS if that's your intention.

dneto0:
I'm not intending this to be included in a CTS, not in this form anyway.  I mainly wanted a place to stash this interesting test case.

asuonpaa:
Do we still want to keep this open?

dneto0:
> Do we still want to keep this open?

I converted it to draft.  I'm referencing it from WebGPU CTS as an interesting case.

dj2:
Can we close the PR? I believe as long as you don't delete the branch it should still be accessible from the URL you used in the link, but it will remove it from the amber PR list which makes it a lot clearer it won't be landing.

PR #919, opened 2020-10-28
roll-all: Fix for GNU find

dneto0:
Use -mindepth 1 instead of -depth 1
Also, put global options before non-global options


dneto0:
I needed this to make it work on Linux

dj2 (approved):

PR #921, opened 2020-10-28
Skip data readback for read-only descriptors

asuonpaa:
This works as an optimization but also fixes an issue where using multiple combined image samplers with different samplers
were failing because Amber was trying to copy the same resource multiple times.


paulthomson:
Is it possible to add a test that previously failed?

src/vulkan/buffer_descriptor.cc
```diff
@@ -123,5 +127,21 @@ std::vector<Resource*> BufferDescriptor::GetResources() {
...
+      return true;
+    case DescriptorType::kStorageBuffer:
+    case DescriptorType::kStorageBufferDynamic:
+    case DescriptorType::kStorageTexelBuffer:
+      return false;
+    default:
+      assert(false && "Unexpected descriptor type");
+      return false;
+  }
+}
```
paulthomson, line 144: Perhaps this should be implemented in `Descriptor` for all types? We could add assertions elsewhere (constructors?) so that the `type_` must make sense for the subclass.

asuonpaa:
> Is it possible to add a test that previously failed?

It is. I'll add a test that would've asserted with a previous version.

src/vulkan/buffer_descriptor.cc
asuonpaa: I would move this to BufferBackedDescriptor instead as it doesn't make any sense for samplers, and it's only used for subclasses of BufferBackedDescriptor.

src/vulkan/buffer_descriptor.cc
```diff
@@ -74,7 +74,11 @@ Result BufferDescriptor::CreateResourceIfNeeded() {
 }
 
 Result BufferDescriptor::MoveResourceToBufferOutput() {
-  Result r = BufferBackedDescriptor::MoveResourceToBufferOutput();
+  Result r;
+
+  if (!IsReadOnly())
+    r = BufferBackedDescriptor::MoveResourceToBufferOutput();
+
```
paulthomson, line 81: I think it would make more sense to do this in `BufferBackedDescriptor::MoveResourceToBufferOutput()`, as we check IsReadOnly() in the other methods of BufferBackedDescriptor.

src/vulkan/image_descriptor.cc
```diff
@@ -125,18 +125,24 @@ Result ImageDescriptor::CreateResourceIfNeeded() {
...
   return {};
 }
 
 Result ImageDescriptor::MoveResourceToBufferOutput() {
-  Result r = BufferBackedDescriptor::MoveResourceToBufferOutput();
+  Result r;
+
+  if (!IsReadOnly())
+    r = BufferBackedDescriptor::MoveResourceToBufferOutput();
+
```
paulthomson, line 145: Same as above; I think we can do this in BufferBackedDescriptor::MoveResourceToBufferOutput().

src/vulkan/buffer_backed_descriptor.cc
```diff
@@ -45,7 +45,10 @@ Result BufferBackedDescriptor::RecordCopyDataToResourceIfNeeded(
   for (size_t i = 0; i < resources.size(); i++) {
     if (!buffers[i]->ValuePtr()->empty()) {
       resources[i]->UpdateMemoryWithRawData(*buffers[i]->ValuePtr());
-      buffers[i]->ValuePtr()->clear();
+      // Keep the data if the resource is read only. Otherwise the data will
+      // be regenerated after a draw.
```
paulthomson, line 49: The "regenerated" word confuses me. Maybe just say:

If the resource is read-only, keep the buffer data; Amber won't copy read-only resources back into the host buffers, so it makes sense to leave the buffer intact.

paulthomson (commented):
Looks good. One thought: it looks like we are adding the first ever Amber test that binds the same buffer multiple times within a pipeline. Due to the way Amber copies data back into host-side buffers after running a pipeline, this seems like something that should not be supported, and should fail with an error. However, it does make sense that binding a read-only buffer multiple times in a pipeline should be fine, and Amber should not try to copy this data back to the host. So your change makes sense, but we should make sure that we want to support this. @dj2 What do you think?

asuonpaa:
> Looks good. One thought: it looks like we are adding the first ever Amber test that binds the same buffer multiple times within a pipeline. Due to the way Amber copies data back into host-side buffers after running a pipeline, this seems like something that should not be supported, and should fail with an error. However, it does make sense that binding a buffer multiple times to _read-only resources_ in a pipeline should be fine, and Amber should not try to copy this data back to the host. So your change makes sense, but we should make sure that we want to support this. @dj2 What do you think?

For what it's worth, the original need for this kind of support originated from GraphicsFuzz.

paulthomson:
Yeah I think it makes complete sense because it allows you to have multiple samplers that sample from the same buffer/texture. I just wanted to give @dj2 a heads up!

dj2:
I think the concept is fine, makes sense why you'd want to do this. Can we restrict it to just read-only bufffers? Or is that not really possible? Maybe add a warning if we copy the same buffer back multiple times?

paulthomson:
Yeah I think it is restricted to read-only Vulkan objects, which is decided in the "BIND" command. I think we will indeed get an error if Amber does try to copy back multiple times to the same buffer.

paulthomson (approved):
LGTM!

PR #926, opened 2020-12-04
Fix Vulkan synchronization validation errors

jeremyg-lunarg:
In Vulkan CTS TransferImage::ImageBarrier() is often used to
synchronize images which are then used as attachments in
a RenderPass. Depending on the load operation used, the RenderPass
may need either read or write access. (see Section 8.1 Render Pass
Creation in the Vulkan 1.2.x specification).

To allow any load operation to work correctly, set both read and write
access in the dstAccessMask when transitioning to an attachment layout.


jeremyg-lunarg:
@googlebot I signed it!

dj2 (approved):

dj2:
I'm not sure what's happened to the bots, re-running them.

PR #929, opened 2020-12-11
VERTEX_DATA: add format, offset, stride parameters

asuonpaa:


src/amberscript/parser.cc
```diff
@@ -1301,20 +1301,55 @@ Result Parser::ParsePipelineVertexData(Pipeline* pipeline) {
   const uint32_t location = token->AsUint32();
 
   InputRate rate = InputRate::kVertex;
+  uint32_t offset = 0;
+  Format* format = buffer->GetFormat();
+  uint32_t stride = format->SizeInBytes();
```
paulthomson, line 1306: This is kind of subtle but I think if somebody overrides the format, but not the stride, then the stride should be derived from the overridden format, not the format of the original buffer. I guess you can check if the stride was explicitly set after the loop (I.e. It is non-zero? Although I guess we should not allow an explicit 0 stride?) and, if not, set it based on the format. And I guess add a test for this as well.

paulthomson (commented):
LGTM! I left a few comments. And lets get @dj2 to quickly review the feature addition by reviewing `amber_script.md`.

docs/amber_script.md
```diff
@@ -588,8 +588,12 @@ when using dynamic buffers with BUFFER_ARRAY.
...
 ```groovy
   # Set |buffer_name| as the vertex data at location |val|. `RATE` defines the
-  # input rate for vertex attribute reading.
-  VERTEX_DATA {buffer_name} LOCATION _val_ [ RATE { vertex | instance } (default vertex) ]
+  # input rate for vertex attribute reading. `FORMAT` can be used to override
+  # the vertex buffer format specified by the |buffer_name|. `OFFSET` defines a
+  # byte offset used for the vertex data binding, and `STRIDE` overrides the
+  # default stride specified by the |buffer_name|.
+  VERTEX_DATA {buffer_name} LOCATION _val_ [ RATE { vertex | instance } (default vertex) ] \
+        [ FORMAT {format} ] [ OFFSET {offset} ] [ STRIDE {stride} ]
```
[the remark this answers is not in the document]
asuonpaa: done

src/amberscript/parser.cc
asuonpaa: done

dj2 (approved):

paulthomson (approved):
LGTM!

PR #936, opened 2021-01-13
Support binding the same vertex buffer to multiple locations

asuonpaa:
There's no need to restrict binding a vertex buffer to multiple locations. Now that we have OFFSET supported this has a real use case where the same buffer is used with different offsets.

### DIFF
--- a/src/vulkan/vertex_buffer.cc
+++ b/src/vulkan/vertex_buffer.cc
@@ -36,18 +36,22 @@ VertexBuffer::VertexBuffer(Device* device) : device_(device) {}
 
 VertexBuffer::~VertexBuffer() = default;
 
-void VertexBuffer::SetData(uint8_t location, Buffer* buffer, InputRate rate) {
-  auto format = buffer->GetFormat();
+void VertexBuffer::SetData(uint8_t location,
+                           Buffer* buffer,
+                           InputRate rate,
+                           Format* format,
+                           uint32_t offset,
+                           uint32_t stride) {
   const uint32_t binding = static_cast<uint32_t>(vertex_attr_desc_.size());
   vertex_attr_desc_.emplace_back();
   vertex_attr_desc_.back().binding = binding;
   vertex_attr_desc_.back().location = location;
-  vertex_attr_desc_.back().offset = 0u;
+  vertex_attr_desc_.back().offset = offset;
   vertex_attr_desc_.back().format = device_->GetVkFormat(*format);
 
   vertex_binding_desc_.emplace_back();
   vertex_binding_desc_.back().binding = binding;
-  vertex_binding_desc_.back().stride = format->SizeInBytes();
+  vertex_binding_desc_.back().stride = stride;
   vertex_binding_desc_.back().inputRate = GetVkInputRate(rate);
 
   data_.push_back(buffer);
--- a/src/vulkan/vertex_buffer.cc
+++ b/src/vulkan/vertex_buffer.cc
@@ -75,11 +75,23 @@ Result VertexBuffer::SendVertexData(CommandBuffer* command) {
   if (!is_vertex_data_pending_)
     return Result("Vulkan::Vertices data was already sent");
 
-  for (const auto& buf : data_) {
-    uint32_t bytes = buf->GetSizeInBytes();
+  for (size_t i = 0; i < data_.size(); i++) {
+    const auto& buf = data_[i];
 
+    // Search the previous bindings to see if the same Amber buffer is
+    // already bound.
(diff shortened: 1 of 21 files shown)


src/vulkan/vertex_buffer.cc
```diff
@@ -75,11 +75,23 @@ Result VertexBuffer::SendVertexData(CommandBuffer* command) {
...
-  for (const auto& buf : data_) {
+  for (size_t i = 0; i < data_.size(); i++) {
+    const auto& buf = data_[i];
+
+    // Search the previous bindings to see if the same Amber buffer is
+    // already bound.
+    for (size_t j = 0; j < i; j++) {
+      if (buf == data_[j]) {
+        // Found one. Reuse existing transfer buffer.
+        transfer_buffers_.emplace_back(transfer_buffers_[j]);
```
dj2, line 86: What's the benefit of reusing vs just creating a new one? If we're reusing, does the initial transfer not copy the offset'd data? Why do we need to add it multiple times?

tests/cases/vertex_data_two_locations.amber
```diff
@@ -0,0 +1,84 @@
...
+
+BUFFER vert_color DATA_TYPE R8G8B8A8_UNORM DATA
+255   0   0 255
+255   0   0 255
+255   0   0 255
+255   0   0 255
+255   0   0 255
+255   0   0 255
+
+  0 255   0 255
```
dj2, line 60: nit: this is strangely indented.

src/vulkan/vertex_buffer.cc
asuonpaa: The whole is copied regardless of the offset which is only used in descriptor write. Even if we would use the offset during the memcopy Amber wouldn't know how much to copy since it only knows the size of the whole buffer. We need to add the transfer buffer reference multiple times because the vector elements are later mapped directly to binding slots. In the end the same VkBuffer handle is being used for the descriptor write.

To be clear all this is for the sake of avoiding unnecessary memory copies. The implementation did work if I just disabled that one check for binding the buffer multiple times.

tests/cases/vertex_data_two_locations.amber
asuonpaa: So is the original test file I used as a base :). I'll fix this one.

src/vulkan/vertex_buffer.cc
paulthomson: If we just disabled that one check for binding the buffer multiple times, then wouldn't we actually end up creating and binding multiple VkBuffers, instead of binding one VkBuffer multiple times?

It is definitely a bit of pain for what seems like a simple feature but: I think we should avoid using shared_ptr. Is it possible to change this a bit, so that transfer_buffers_ is NOT used in `BindToCommandBuffer` to figure out the binding slots. `transfer_buffers_` could be a map from Buffer to TransferBuffer? Then instead loop over `data_` in `BindToCommandBuffer`?

asuonpaa: That's right. Using multiple VkBuffers would work but it's non-optimal.

I'll try using that map approach.

src/vulkan/vertex_buffer.cc
```diff
@@ -61,46 +61,39 @@ void VertexBuffer::BindToCommandBuffer(CommandBuffer* command) {
...
-      if (buf == data_[j]) {
-        // Found one. Reuse existing transfer buffer.
-        transfer_buffers_.emplace_back(transfer_buffers_[j]);
-        return {};
-      }
-    }
+  buffer_map_.clear();
+
+  for (const auto& buf : data_) {
+    if (buffer_map_.count(buf) == 0) {
```
dj2, line 80: nit:
```
if (buffer_map.count(buf) != 0) {
  continue;
}
```

src/vulkan/vertex_buffer.h
```diff
@@ -63,11 +64,12 @@ class VertexBuffer {
 
   bool is_vertex_data_pending_ = true;
 
-  std::vector<std::shared_ptr<TransferBuffer>> transfer_buffers_;
+  std::vector<std::unique_ptr<TransferBuffer>> transfer_buffers_;
 
   std::vector<Buffer*> data_;
   std::vector<VkVertexInputBindingDescription> vertex_binding_desc_;
   std::vector<VkVertexInputAttributeDescription> vertex_attr_desc_;
+  std::map<Buffer*, VkBuffer> buffer_map_;
```
dj2, line 72: Let's call this `buffer_to_vk_buffer_`

paulthomson (approved):

PR #937, opened 2021-01-14
Allow hex for values in expect commands

alan-baker:
* Handled the same as parsing buffer data

### DIFF
--- a/src/amberscript/parser.cc
+++ b/src/amberscript/parser.cc
@@ -2912,8 +2912,6 @@ Result Parser::ParseValues(const std::string& name,
                       " command: " + token->ToOriginalString());
       }
 
-      double val = token->IsHex() ? static_cast<double>(token->AsHex())
-                                  : token->AsDouble();
       Result r = token->ConvertToDouble();
       if (!r.IsSuccess())
         return r;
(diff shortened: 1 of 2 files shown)


src/amberscript/parser.cc
```diff
@@ -2907,23 +2907,26 @@ Result Parser::ParseValues(const std::string& name,
...
     if (type::Type::IsFloat(segs[seg_idx].GetFormatMode())) {
-      if (!token->IsInteger() && !token->IsDouble()) {
+      if (!token->IsInteger() && !token->IsDouble() && !token->IsHex()) {
         return Result(std::string("Invalid value provided to ") + name +
                       " command: " + token->ToOriginalString());
       }
 
+      double val = token->IsHex() ? static_cast<double>(token->AsHex())
+                                  : token->AsDouble();
       Result r = token->ConvertToDouble();
```
dj2, line 2917: Does this need to happen before the ->AsDouble in the line above?

src/amberscript/parser.cc
```diff
@@ -2907,23 +2907,26 @@ Result Parser::ParseValues(const std::string& name,
...
       if (!r.IsSuccess())
         return r;
 
       v.SetDoubleValue(token->AsDouble());
```
dj2, line 2921: Should `token->AsDouble` be `val` ?

src/amberscript/parser.cc
alan-baker: Actually this line is clearly useless. ConvertToDouble already handles hex.

dj2 (approved):

PR #940, opened 2021-01-20
Fix buffer size for D24_UNORM_S8_INT format

asuonpaa:
Vulkan expects depth component of D24_UNORM format to be padded to 32 bits. This change adds an 8-bit padding after the depth channel.

### DIFF
--- a/src/format.h
+++ b/src/format.h
@@ -67,8 +67,8 @@ class Format {
    private:
     bool is_padding_ = false;
     bool is_packable_ = false;
-    FormatComponentType name_ = FormatComponentType::kR;
-    FormatMode mode_ = FormatMode::kSInt;
+    FormatComponentType name_ = FormatComponentType::kX;
+    FormatMode mode_ = FormatMode::kUNorm;
     uint32_t num_bits_ = 0;
   };
 
--- a/src/type_parser_test.cc
+++ b/src/type_parser_test.cc
@@ -422,9 +422,10 @@ TEST_F(TypeParserTest, Formats) {
       {"D24_UNORM_S8_UINT",
        FormatType::kD24_UNORM_S8_UINT,
        0U,
-       2U,
+       3U,
        {
            {FormatComponentType::kD, FormatMode::kUNorm, 24},
+           {FormatComponentType::kX, FormatMode::kUNorm, 8},
            {FormatComponentType::kS, FormatMode::kUInt, 8},
        }},
       {"D32_SFLOAT",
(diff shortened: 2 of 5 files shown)


asuonpaa:
Looks like the newly added tests use depth format that is not supported by the bot environment which causes the Ubuntu builds to fail. What should we do with those?

src/format.h
```diff
@@ -67,8 +67,8 @@ class Format {
    private:
     bool is_padding_ = false;
     bool is_packable_ = false;
-    FormatComponentType name_ = FormatComponentType::kR;
```
dj2, line 70: Why did the default values change?

src/type_parser_test.cc
```diff
@@ -422,9 +422,10 @@ TEST_F(TypeParserTest, Formats) {
       {"D24_UNORM_S8_UINT",
        FormatType::kD24_UNORM_S8_UINT,
        0U,
-       2U,
+       3U,
        {
            {FormatComponentType::kD, FormatMode::kUNorm, 24},
+           {FormatComponentType::kX, FormatMode::kUNorm, 8},
```
dj2, line 428: What does this represent? This is no longer 32 bits, it's 40 bits isn't it?

dj2 (commented):
Isn't the D24UNORM_S8UINT format already 32 bits as there is 24bits of depth and 8 bits of stencil? Adding another 8 bits makes it 40 bits doesn't it?

paulthomson:
Yeah, I am not convinced this is correct, but I don't yet understand why (without your PR) the `tests/cases/draw_rectangles_depth_test_d24s8.amber` test fails and the `tests/cases/draw_rectangles_depth_test_x8d24.amber` test passes. I think we must be missing something.

paulthomson:
I think the issue is actually elsewhere in Amber. I have been trying to figure this out for a few hours and I could still be wrong. See here:

https://github.com/google/amber/blob/cc1dec6104f5e0e656d8439296265a19393628df/src/vulkan/transfer_image.cc#L221

I think Amber is copying the depth/stencil buffer to the host and back, and messing things up a bit when it does so. This is why the framebuffer ends up with the wrong image (the depth buffer was changed by Amber between each draw). It just so happens that making the format bigger means that we have enough room for things to work, but I don't think that is the real fix. I _think_ the real fix is to adjust the size of the host buffer when we need space for stencil data. Probably here:

https://github.com/google/amber/blob/cc1dec6104f5e0e656d8439296265a19393628df/src/vulkan/transfer_image.cc#L152

PR #941, opened 2021-01-21
Make room for stencil data in host buffer

asuonpaa:
When copying a D24_UNORM_S8_UINT depth/stencil buffer to or from the host, the depth and stencil data must be copied separately. This change adds extra space to the end of the Amber host buffer for the stencil data.

### DIFF
--- a/src/vulkan/transfer_image.cc
+++ b/src/vulkan/transfer_image.cc
@@ -81,6 +81,8 @@ TransferImage::TransferImage(Device* device,
           device,
           x * y * z *
               (format.SizeInBytes() +
+               // D24_UNORM_S8_UINT requires 32bit component for depth when
+               // performing buffer copies. Reserve extra room to handle that.
                (format.GetFormatType() == FormatType::kD24_UNORM_S8_UINT ? 1
                                                                          : 0))),
       image_info_(kDefaultImageInfo),
--- a/src/vulkan/transfer_image.cc
+++ b/src/vulkan/transfer_image.cc
@@ -77,7 +77,12 @@ TransferImage::TransferImage(Device* device,
                              uint32_t base_mip_level,
                              uint32_t used_mip_levels,
                              uint32_t samples)
-    : Resource(device, x * y * z * format.SizeInBytes()),
+    : Resource(
+          device,
+          x * y * z *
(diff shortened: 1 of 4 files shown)


src/vulkan/transfer_image.cc
```diff
@@ -77,7 +77,12 @@ TransferImage::TransferImage(Device* device,
                              uint32_t base_mip_level,
                              uint32_t used_mip_levels,
                              uint32_t samples)
-    : Resource(device, x * y * z * format.SizeInBytes()),
+    : Resource(
+          device,
+          x * y * z *
+              (format.SizeInBytes() +
+               (format.GetFormatType() == FormatType::kD24_UNORM_S8_UINT ? 1
+                                                                         : 0))),
```
paulthomson, line 85: We should add a comment to explain why.

paulthomson (commented):
Looks good! However, take a look at:

* `GraphicsPipeline::Clear`
* `GraphicsPipeline::Draw`

These both call `CopyBuffersToImages` and `TransferImagesToDevice`. Perhaps we could remove the code in these functions (and the functions that they call) that handles the depth stencil buffers, and instead handle them at a higher level (e.g. in `GraphicsPipeline::Draw`). Perhaps this would make it easier to create temporary buffers etc. and interleave the stencil data as needed.

src/vulkan/transfer_image.cc
asuonpaa: done

asuonpaa:
> Looks good! However, take a look at:
> 
> * `GraphicsPipeline::Clear`
> * `GraphicsPipeline::Draw`
> 
> These both call `CopyBuffersToImages` and `TransferImagesToDevice`. Perhaps we could remove the code in these functions (and the functions that they call) that handles the depth stencil buffers, and instead handle them at a higher level (e.g. in `GraphicsPipeline::Draw`). Perhaps this would make it easier to create temporary buffers etc. and interleave the stencil data as needed.

I don't see why we'd want to add code that prepares for interleaving unless we add the functionality with the same commit. Also wouldn't it just add redundancy if we moved the buffer handling higher where it would be required in multiple places?

paulthomson (approved):

PR #943, opened 2021-02-11
Add device API version checking for function pointer loading

asuonpaa:
Instance functions promoted to core in later Vulkan versions should only be loaded if the underlying physical device supports those versions.


paulthomson (approved):
Looks good! I also tested it after deleting `vk.xml`; seems to work fine in this case as well.

PR #944, opened 2021-03-10
Add VIEWPORT to AmberScript

gnl21:
This allows setting the viewport for graphics pipelines.

It has only been lightly tested and may not be compatible with all parts of Amber. I wanted to use this for some tests that I was writing and would like to upstream it. If the amberscript syntax and the basic idea of the change are OK then I'll fix any problems with the details of the change.


paulthomson (commented):
Thanks for the contribution! I think the basic syntax looks good and the basic changes look good.

Could you start by adding an example that uses the new command in `tests/cases/`? That way, we can get an idea of what the command looks like (although I think I get it, in this case, and I think we approve).

I would also suggest tweaking the syntax to:

`VIEWPORT {x} {y} SIZE {width} {height} [MIN_DEPTH {mind}] [MAX_DEPTH {maxd}]`

to more closely follow existing commands. I guess min depth can default to 0 and max depth to 1.

And then also document the above syntax in: `docs/amber_script.md`.

asuonpaa:
The changes look good. In addition to what Paul said I think we should also have some parser unit tests.

I'm wondering if we should also support multiple viewports to better match what Vulkan has. Basically allow multiple VIEWPORT commands and do a vector of viewports in PipelineData. HasViewportData could then be removed as an empty vector would serve the same purpose.

gnl21:
Thanks for the comments. I've updated the syntax and provided some quick examples.

I like the idea of using a vector of viewports to support the multiple viewports feature, but I don't have a device that supports the feature that I could test on, so I haven't done that. The approaches seem compatible, so adding that should be a simple extension for another day.

I forgot about parser unit tests but I'm happy to add those.

I haven't got any way of testing the dawn backend so I haven't made any changes there. I assume this means the token will be quietly ignored, which doesn't seem like a good thing.

src/amberscript/parser_viewport_test.cc
```diff
@@ -0,0 +1,340 @@
+// Copyright 2020 The Amber Authors.
```
asuonpaa, line 1: 2021

src/amberscript/parser_viewport_test.cc
```diff
@@ -0,0 +1,340 @@
...
+  BIND BUFFER my_fb AS color LOCATION 0
+  BIND BUFFER my_ds AS depth_stencil
+
+  VIEWPORT 0 0 SIZE 15 20
+END)";
+
+  Parser parser;
+  Result r = parser.Parse(in);
+  ASSERT_FALSE(r.IsSuccess()) << r.Error();
+  EXPECT_EQ("15: invalid offset for VIEWPORT command", r.Error());
```
asuonpaa, line 216: Could we also allow integers and just convert them to floats? If not, at least the error message should be more specific.

tests/cases/viewport.amber
```diff
@@ -0,0 +1,68 @@
...
+
+  FRAMEBUFFER_SIZE 256 256
+  VIEWPORT 10.0 10.0 SIZE 100.0 100.0
+
+  BIND BUFFER framebuffer AS color LOCATION 0
+END
+
+CLEAR_COLOR pipeline1 255 255 255 255
+CLEAR pipeline1
+RUN pipeline1 DRAW_ARRAY AS TRIANGLE_FAN START_IDX 0 COUNT 4
```
asuonpaa, line 60: You could probably use DRAW_RECT instead and drop the position buffer.

tests/cases/viewport_depth.amber
```diff
@@ -0,0 +1,70 @@
...
+  BIND BUFFER framebuffer AS color LOCATION 0
+  BIND BUFFER depth AS depth_stencil
+END
+
+CLEAR_DEPTH pipeline1 1.0
+CLEAR_COLOR pipeline1 255 255 255 255
+CLEAR pipeline1
+RUN pipeline1 DRAW_ARRAY AS TRIANGLE_FAN START_IDX 0 COUNT 4
+
+EXPECT depth IDX 0 TOLERANCE 1.0e-6 EQ 0.6
```
asuonpaa, line 70: I know it's not possible to probe different locations of a depth buffer, but you could have a separate pipeline that converts the depth data to a grayscale format that you then could use. There are some amber files in the same directory that does that, so you can use one as a reference. Then you could have the viewport size something else than the whole frame buffer.

asuonpaa (changes requested):

paulthomson (commented):
Thanks a lot for taking care of the unit tests! I think we can skip the dawn backend, for now.

tests/cases/viewport_depth.amber
paulthomson: I would be happy for this to come as a followup PR (perhaps even you, Ari, can take care of this?), as this sounds fairly unintuitive.

gnl21: I looked at this a little bit because the tests that I was originally trying to create are for depth-related things. Since rendering to floating-point images is relatively common (and is particularly useful for testing), could we instead create a syntax for verifying floating-point images as images? I haven't really looked into this, so I don't know how complex it would be, but having to add extra pipelines to reprocess any image with a bit-depth over 8 seems a bit clunky to me.

As for making this test use a viewport other than the whole framebuffer, this was actually my original usecase, just setting a depth range for the viewport, so that's where this test came from. Certainly it could be made to use a different viewport setting once the verification is changed.

src/amberscript/parser_viewport_test.cc
gnl21: Oops, yes, thanks

src/amberscript/parser_viewport_test.cc
gnl21: I'm fine with that, although I don't think it wasn't done like that in whatever command I copied this from. They have to be floats in Vulkan. Is there a general policy on how these commands work in amber?

src/amberscript/parser.cc
```diff
@@ -949,6 +951,64 @@ Result Parser::ParsePipelineFramebufferSize(Pipeline* pipeline) {
   return ValidateEndOfStatement("FRAMEBUFFER_SIZE command");
 }
 
+Result Parser::ParsePipelineViewport(Pipeline* pipeline) {
+  float vp[6] = {0, 0, 0, 0, 0, 1};
```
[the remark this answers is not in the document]
gnl21: I agree that a struct would be nicer, but I didn't do it that way because everything else in PipelineData was individual data members. I didn't want to put in individual setters because that creates a whole lot of invalid states where one viewport component has been set but the others have not, which I'd rather avoid.

tests/cases/viewport.amber
gnl21: Done.

src/amberscript/parser_viewport_test.cc
gnl21: Actually, I found the places in the existing code that were doing this and have updated this change to allow integers as well.

paulthomson (approved):
Looks good to me! Please rebase on main (or I can merge main) and then I will merge.

gnl21:
I thought it would be easiest to just hit the "Merge in main" button, but it looks like I was wrong because now my email addresses is messed up in the merge commit. I'll rebase the branch onto main and re-push it. Sorry about that.

paulthomson:
No worries, and thanks!

PR #947, opened 2021-04-09
Add support to define descriptor buffer offset and range

### LINKED ISSUE #945: Add descriptor buffer offset and range
We can already set the buffer offset for vertex buffer bindings using `OFFSET` argument in `VERTEX_DATA` command.

Same should be possible for uniform buffers. This could be implemented by adding `DESCRIPTOR_OFFSET` and `DESCRIPTOR_RANGE` argument to the `BIND {BUFFER | BUFFER_ARRAY}` command. Note that the `OFFSET` keyword is already used to set dynamic offset(s) for dynamic buffers, so we need to use `DESCRIPTOR_` prefix.

These new arguments are used to set the members `offset` and `range` of Vulkan struct `VkDescriptorBufferInfo`.

The current implementation doesn't allow binding the same buffer into different array elements within one `BIND BUFFER_ARRAY` command. This should be allowed too.

ilkkasaa:
Descriptor buffers can now be bound with byte offset and range via DESCRIPTOR_OFFSET and DESCRIPTOR_RANGE parameters.

Descriptor array elements can now point to same buffer.

Fixes problem with dynamic uniform/storage buffers: when a dynamic offset is > 0, VK_WHOLE_SIZE must not be used as buffer range.

Fixes #945

### DIFF
--- a/src/amberscript/parser.cc
+++ b/src/amberscript/parser.cc
@@ -22,6 +22,7 @@
 #include <utility>
 #include <vector>
 
+#include "amber/vulkan_header.h"
 #include "src/image.h"
 #include "src/make_unique.h"
 #include "src/sampler.h"
@@ -1243,10 +1244,61 @@ Result Parser::ParsePipelineBind(Pipeline* pipeline) {
           }
         }
 
+        // Set default descriptor buffer offsets to 0 and descriptor buffer
+        // ranges to VK_WHOLE_SIZE.
+        std::vector<VkDeviceSize> descriptor_offsets(buffers.size(), 0);
+        std::vector<VkDeviceSize> descriptor_ranges(buffers.size(),
+                                                    VK_WHOLE_SIZE);
+        if (buffer_type == BufferType::kUniformDynamic ||
+            buffer_type == BufferType::kStorageDynamic ||
+            buffer_type == BufferType::kStorage ||
+            buffer_type == BufferType::kUniform) {
+          token = tokenizer_->PeekNextToken();
+          if (token->IsIdentifier() &&
+              token->AsString() == "DESCRIPTOR_OFFSET") {
+            token = tokenizer_->NextToken();
+            for (size_t i = 0; i < buffers.size(); i++) {
+              token = tokenizer_->NextToken();
+              if (!token->IsInteger()) {
+                if (i > 0) {
+                  return Result(
+                      "expecting an DESCRIPTOR_OFFSET value for each buffer in "
+                      "the array");
+                } else {
+                  return Result(
+                      "expecting an integer value for DESCRIPTOR_OFFSET");
+                }
+              }
+              descriptor_offsets[i] = token->AsUint64();
+            }
+          }
+
+          token = tokenizer_->PeekNextToken();
+          if (token->IsIdentifier() &&
+              token->AsString() == "DESCRIPTOR_RANGE") {
+            token = tokenizer_->NextToken();
+            for (size_t i = 0; i < buffers.size(); i++) {
+              token = tokenizer_->NextToken();
+              if (!token->IsInteger()) {
+                if (i > 0) {
+                  return Result(
+                      "expecting an DESCRIPTOR_RANGE value for each buffer in "
+                      "the array");
+                } else {
+                  return Result(
+                      "expecting an integer value for DESCRIPTOR_RANGE");
+                }
+              }
+              descriptor_ranges[i] = token->AsUint64();
+            }
+          }
+        }
+
         pipeline->ClearBuffers(descriptor_set, binding);
         for (size_t i = 0; i < buffers.size(); i++) {
           pipeline->AddBuffer(buffers[i], buffer_type, descriptor_set, binding,
-                              base_mip_level, dynamic_offsets[i]);
+                              base_mip_level, dynamic_offsets[i],
+                              descriptor_offsets[i], descriptor_ranges[i]);
         }
       } else if (token->IsIdentifier() && token->AsString() == "KERNEL") {
         token = tokenizer_->NextToken();
--- a/src/amberscript/parser.cc
+++ b/src/amberscript/parser.cc
@@ -1260,7 +1260,7 @@ Result Parser::ParsePipelineBind(Pipeline* pipeline) {
               if (!token->IsInteger()) {
                 if (i > 0) {
                   return Result(
-                      "expecting an DESCRIPTOR_OFFSET value for each buffer in "
+                      "expecting a DESCRIPTOR_OFFSET value for each buffer in "
                       "the array");
                 } else {
                   return Result(
@@ -1280,7 +1280,7 @@ Result Parser::ParsePipelineBind(Pipeline* pipeline) {
               if (!token->IsInteger()) {
                 if (i > 0) {
                   return Result(
-                      "expecting an DESCRIPTOR_RANGE value for each buffer in "
+                      "expecting a DESCRIPTOR_RANGE value for each buffer in "
                       "the array");
                 } else {
                   return Result(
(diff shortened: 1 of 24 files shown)


src/amberscript/parser.cc
```diff
@@ -22,6 +22,7 @@
 #include <utility>
 #include <vector>
 
+#include "amber/vulkan_header.h"
```
paulthomson, line 25: I don't think you are supposed to include the Vulkan header in the parser. We cannot use Vulkan types here.

ilkkasaa: Oh, I see. Makes sense.

paulthomson (commented):
This looks really good, especially all the tests!

However, you have changed a few vectors into unordered_maps, and we iterate over these. We almost never want to iterate over maps because the order is nondeterministic. This can cause errors to become (even more) nondeterministic, when they don't need to be. Thus, we ideally always want to have a vector (containing the keys) available for the cases where we iterate, and possibly a map alongside if needed (or another vector, but maps/sets are good for preventing duplicates). An ordered map is not ideal either; the order can still be nondeterministic (e.g. if storing pointers), and it is more intuitive to iterate in the order that items were added, not some other ordering. Please can you find a way to remove the iteration over maps? Thanks.

ilkkasaa:
Yeah, that's a good point! So, now it doesn't loop over maps anymore. The transfer images/buffers are still stored in maps for easier inserts and `amber_buffer`->`transfer_buffer` lookups, but when looping, the `amber_buffers` vector is used to get the keys or an ordered vector of `amber_buffer`->`transfer_buffer` pairs.

ilkkasaa:
While doing this PR, I was thinking, that now we can share a buffer between descriptors within a descriptor array, but the same doesn't apply to separate bindings or descriptor sets. e.g. following Amber code is valid, but it will actually create two buffers with same contents:
```
BIND BUFFER buf0 AS uniform DESCRIPTOR_SET 0 BINDING 0 
BIND BUFFER buf0 AS uniform DESCRIPTOR_SET 0 BINDING 1
```
The real problem is with buffers that the shaders can update, like SSBOs, e.g. this doesn't even work:
```
BIND BUFFER buf0 AS storage DESCRIPTOR_SET 0 BINDING 0
BIND BUFFER buf0 AS storage DESCRIPTOR_SET 0 BINDING 1
```
I think that Amber needs some completely different way of handling descriptor buffers to resolve this problem, but that's probably out of this PR's scope.

paulthomson:
```
[arm64-v8a] Compile++      : amber <= buffer_descriptor.cc
../src/vulkan/buffer_descriptor.cc:158:19: error: 'auto' not allowed in lambda parameter
        [&](const auto& buffer) { return buffer.first == amber_buffer; });
                  ^~~~
In file included from ../src/vulkan/buffer_descriptor.cc:15:
In file included from ../src/vulkan/buffer_descriptor.h:19:
In file included from /tmpfs/src/android-ndk-r21/sources/cxx-stl/llvm-libc++/include/unordered_map:409:
In file included from /tmpfs/src/android-ndk-r21/sources/cxx-stl/llvm-libc++/include/__hash_table:17:
/tmpfs/src/android-ndk-r21/sources/cxx-stl/llvm-libc++/include/algorithm:933:13: error: no matching function for call to object of type '(lambda at ../src/vulkan/buffer_descriptor.cc:158:9)'
        if (__pred(*__first))
            ^~~~~~
../src/vulkan/buffer_descriptor.cc:156:30: note: in instantiation of function template specialization 'std::__ndk1::find_if<std::__ndk1::__wrap_iter<std::__ndk1::pair<amber::Buffer *, amber::vulkan::Resource *> *>, (lambda at ../src/vulkan/buffer_descriptor.cc:158:9)>' requested here
    const auto& image = std::find_if(
                             ^
../src/vulkan/buffer_descriptor.cc:158:9: note: candidate function not viable: no known conversion from 'std::__ndk1::pair<amber::Buffer *, amber::vulkan::Resource *>' to 'int &' for 1st argument
        [&](const auto& buffer) { return buffer.first == amber_buffer; });
        ^
2 errors generated.
make: *** [/tmpfs/src/android-ndk-r21/build/core/build-binary.mk:478: /tmpfs/src/github/amber/build/app/local/arm64-v8a/objs/amber/src/vulkan/buffer_descriptor.o] Error 1
```

ilkkasaa:
Thanks, I'll fix that.

src/vulkan/buffer_descriptor.cc
```diff
@@ -39,17 +42,24 @@ Result BufferDescriptor::CreateResourceIfNeeded() {
...
+      continue;
+
     uint32_t size_in_bytes =
         amber_buffer ? static_cast<uint32_t>(amber_buffer->ValuePtr()->size())
                      : 0;
-    transfer_buffers_.emplace_back(MakeUnique<TransferBuffer>(
-        device_, size_in_bytes, amber_buffer->GetFormat()));
+
+    auto transfer_buffer = MakeUnique<TransferBuffer>(
+        device_, size_in_bytes, amber_buffer->GetFormat());
```
paulthomson, line 61: The ternary expression above suggests that `amber_buffer` can be nullptr. But then we use it here anyway.  You haven't actually changed this. But it seems odd. Perhaps we should just remove the ternary expression?

paulthomson (commented):
Nice work! Left a few minor comments.

paulthomson:
> While doing this PR, I was thinking, that now we can share a buffer between descriptors within a descriptor array, but the same doesn't apply to separate bindings or descriptor sets. e.g. following Amber code is valid, but it will actually create two buffers with same contents:
> 
> ```
> BIND BUFFER buf0 AS uniform DESCRIPTOR_SET 0 BINDING 0 
> BIND BUFFER buf0 AS uniform DESCRIPTOR_SET 0 BINDING 1
> ```
> 
> The real problem is with buffers that the shaders can update, like SSBOs, e.g. this doesn't even work:
> 
> ```
> BIND BUFFER buf0 AS storage DESCRIPTOR_SET 0 BINDING 0
> BIND BUFFER buf0 AS storage DESCRIPTOR_SET 0 BINDING 1
> ```
> 
> I think that Amber needs some completely different way of handling descriptor buffers to resolve this problem, but that's probably out of this PR's scope.

Oh really!? That is a shame. I think we should fix this, if you can think of a way.

src/vulkan/buffer_descriptor.cc
ilkkasaa: Yeah, that seems odd. `amber_buffer` is actually used also before the ternary expression:
```
if (amber_buffer->ValuePtr()->empty())
      continue;
```
I removed the ternary expression.

ilkkasaa:
> > While doing this PR, I was thinking, that now we can share a buffer between descriptors within a descriptor array, but the same doesn't apply to separate bindings or descriptor sets. e.g. following Amber code is valid, but it will actually create two buffers with same contents:
> > ```
> > BIND BUFFER buf0 AS uniform DESCRIPTOR_SET 0 BINDING 0 
> > BIND BUFFER buf0 AS uniform DESCRIPTOR_SET 0 BINDING 1
> > ```
> > 
> > 
> > The real problem is with buffers that the shaders can update, like SSBOs, e.g. this doesn't even work:
> > ```
> > BIND BUFFER buf0 AS storage DESCRIPTOR_SET 0 BINDING 0
> > BIND BUFFER buf0 AS storage DESCRIPTOR_SET 0 BINDING 1
> > ```
> > 
> > 
> > I think that Amber needs some completely different way of handling descriptor buffers to resolve this problem, but that's probably out of this PR's scope.
> 
> Oh really!? That is a shame. I think we should fix this, if you can think of a way. But, as you say, not necessarily in this PR.

Yeah, I have some ideas. I can look into that after this PR.

paulthomson (approved):

PR #951, opened 2021-05-19
Use shared transfer buffers

### LINKED ISSUE #950: Add ability to use same buffer in multiple bindings
Buffer descriptor offsets and ranges are now supported, so it makes sense to bind one buffer to multiple bindings.
For example the following is not allowed currently:

```
BIND BUFFER buf0 AS storage DESCRIPTOR_SET 0 BINDING 0
BIND BUFFER buf0 AS storage DESCRIPTOR_SET 0 BINDING 1 DESCRIPTOR_OFFSET 256
```

ilkkasaa:
Descriptors were using separate transfer buffers when they were
using same buffer. This prevented e.g. using the same buffer in
multiple storage buffer bindings.

This commit moves the transfer buffer/image ownership from descriptors to
pipeline to allow descriptors to share the transfer buffers. Now the
descriptors can have the same buffer binding even if the descriptors
are of different types e.g. storage/uniform/storage_texel_buffer etc.

Fixes #950

### DIFF
--- a/src/vulkan/buffer_descriptor.cc
+++ b/src/vulkan/buffer_descriptor.cc
@@ -67,10 +67,9 @@ Result BufferDescriptor::CreateResourceIfNeeded() {
       transfer_buffer->SetReadOnly(IsReadOnly());
       transfer_resources[amber_buffer] = std::move(transfer_buffer);
     } else {
-      auto& transfer_buffer = transfer_resources[amber_buffer];
       // Unset transfer buffer's read only property if needed.
-      if (!IsReadOnly() && transfer_resources[amber_buffer]->IsReadOnly()) {
-        transfer_buffer->SetReadOnly(false);
+      if (!IsReadOnly()) {
+        transfer_resources[amber_buffer]->SetReadOnly(false);
       }
     }
 
--- a/src/vulkan/buffer_descriptor.cc
+++ b/src/vulkan/buffer_descriptor.cc
@@ -29,65 +29,60 @@ BufferDescriptor::BufferDescriptor(Buffer* buffer,
                                    DescriptorType type,
                                    Device* device,
                                    uint32_t desc_set,
-                                   uint32_t binding)
-    : BufferBackedDescriptor(buffer, type, device, desc_set, binding) {}
+                                   uint32_t binding,
+                                   Pipeline* pipeline)
+    : BufferBackedDescriptor(buffer,
+                             type,
+                             device,
+                             desc_set,
+                             binding,
+                             pipeline) {}
 
 BufferDescriptor::~BufferDescriptor() = default;
 
 Result BufferDescriptor::CreateResourceIfNeeded() {
-  if (!transfer_buffers_.empty()) {
-    return Result(
-        "Vulkan: BufferDescriptor::CreateResourceIfNeeded() must be called "
-        "only when |transfer_buffers| is empty");
+  auto& transfer_resources = pipeline_->GetDescriptorTransferResources();
+
+  VkBufferUsageFlags flags =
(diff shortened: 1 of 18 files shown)


src/vulkan/buffer_descriptor.cc
```diff
@@ -29,65 +29,60 @@ BufferDescriptor::BufferDescriptor(Buffer* buffer,
...
+          static_cast<uint32_t>(amber_buffer->ValuePtr()->size());
+      auto transfer_buffer = MakeUnique<TransferBuffer>(
+          device_, size_in_bytes, amber_buffer->GetFormat());
+      transfer_buffer->SetReadOnly(IsReadOnly());
+      transfer_resources[amber_buffer] = std::move(transfer_buffer);
     } else {
-      return Result("Unexpected buffer type when deciding usage flags");
+      auto& transfer_buffer = transfer_resources[amber_buffer];
+      // Unset transfer buffer's read only property if needed.
+      if (!IsReadOnly() && transfer_resources[amber_buffer]->IsReadOnly()) {
```
asuonpaa, line 72: Does the SetReadOnly do something else than just set a member variable to true or false? If not, could simply do SetReadOnly(IsReadOnly()) outside these blocks.

src/vulkan/pipeline.cc
```diff
@@ -409,6 +423,20 @@ Result Pipeline::SendDescriptorDataToDeviceIfNeeded() {
       }
     }
 
+    // Initialize transfer buffers. Transfer images are already initialized in
```
asuonpaa, line 426: Why do we need to initialize them in different places?

tests/cases/compute_one_buffer_in_multiple_bindings.amber
```diff
@@ -0,0 +1,94 @@
+#!amber
+# Copyright 2019 The Amber Authors.
```
asuonpaa, line 2: 2021

asuonpaa (changes requested):

src/vulkan/pipeline.cc
ilkkasaa: Transfer buffers needed to be initialized in later phase, because the buffer usage flags could change while looping through the descriptors in the loop above. Both transfer buffers and images are now initialized here to make the code more readable.

src/vulkan/buffer_descriptor.cc
ilkkasaa: I simplified this a little bit. The transfer buffer's readOnly property should never be changed from false to true, because some descriptors may need only read access, but some other may need write access too.

asuonpaa (approved):
Looks good to me!

src/vulkan/transfer_buffer.h
```diff
@@ -34,7 +34,9 @@ class TransferBuffer : public Resource {
   TransferBuffer(Device* device, uint32_t size_in_bytes, Format* format);
   ~TransferBuffer() override;
 
-  Result Initialize(const VkBufferUsageFlags usage);
+  TransferBuffer* AsTransferBuffer() override { return this; }
+  void AddUsageFlags(VkBufferUsageFlags flags) { usage_flags_ |= flags; }
```
paulthomson, line 38: Should we assert that `buffer_` is `VK_NULL_HANDLE`? I.e. you cannot add usage flags after Initialize?

paulthomson (commented):
Nice! One comment. And please rebase and/or merge from main.

src/vulkan/transfer_buffer.h
ilkkasaa: Good point. I'll add that

paulthomson (approved):

PR #958, opened 2021-08-17
Added feature check support for subgroup properties

asuonpaa:
This is needed for unifying with the Vulkan CTS requirements check.


paulthomson (approved):
Nice! LGTM!

PR #963, opened 2021-09-21
Allow AmberScript to define the number of control points used by tessellation

asuonpaa:
Previously only VkScript had the access to set the number of patch control points. Now this can also be set in AmberScript pipeline settings.

### DIFF
--- a/docs/amber_script.md
+++ b/docs/amber_script.md
@@ -441,6 +441,11 @@ The following commands are all specified within the `PIPELINE` command.
   POLYGON_MODE {mode}
 ```
 
+```groovy
+  # Set the number of patch control points used by tessellation.
+  PATCH_CONTROL_POINTS {control_points}
+```
+
 #### Compare operations
  * `never`
  * `less`
--- a/docs/amber_script.md
+++ b/docs/amber_script.md
@@ -442,7 +442,7 @@ The following commands are all specified within the `PIPELINE` command.
 ```
 
 ```groovy
-  # Set the number of patch control points used by tessellation.
+  # Set the number of patch control points used by tessellation. The default value is 3.
   PATCH_CONTROL_POINTS {control_points}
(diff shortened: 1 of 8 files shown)


docs/amber_script.md
```diff
@@ -441,6 +441,11 @@ The following commands are all specified within the `PIPELINE` command.
   POLYGON_MODE {mode}
 ```
 
+```groovy
+  # Set the number of patch control points used by tessellation.
+  PATCH_CONTROL_POINTS {control_points}
```
paulthomson, line 446: Can you document that the default is 3?

And is it a good idea to make the default 3, or would it be better to change the default to 0 and then fail if we end up trying to use the invalid value 0? I suppose this might break existing vkscript code.

paulthomson (commented):
One suggestion

docs/amber_script.md
asuonpaa: Yeah, I'll add that comment.

I'm quite sure some older tests, maybe even in Vulkan CTS, rely on that hard coded value. I'd just keep the existing value there.

asuonpaa:
Not sure what's going on with Windows CI builds. The last commit only touches the documentation and the previous commit passed all checks.

ben-clayton:
The error is:

```
-- Installing SPIRV-Header
CMake Error at third_party/vulkan-headers/CMakeLists.txt:21 (cmake_minimum_required):
  CMake 3.10.2 or higher is required.  You are running version 3.10.1
```

It looks like this project uses a VM from a 'dynamic' pool. Maybe the VM image changed.
My group maintain a VM image that we control (and likely has more up-to-date toolchains). I'll take a look at migrating this project over to using that.

ben-clayton:
All fixed.

paulthomson (approved):

PR #965, opened 2021-09-26
Update and improve android_gradle

### LINKED ISSUE #964: android_gradle no longer builds
The Android app under `android_gradle/` no longer builds because the jcenter and bintray repositories were shut down.

I have a fix. We can essentially take the latest sample Android app project and update our build files to match. While we are here, there are a few other things we can improve. E.g. we can have gradle automatically build the native code via CMake, rather than having to manually build via ndk-build and copy in. We may even be able to eventually remove the Android.mk files, as I don't think they are used.

paulthomson:
* Update CMakeLists.txt files so we can build libamber_ndk.so for
Android via CMake.

* `android_gradle/`:

  * Update Gradle version, settings, and build files to match the latest
  Android sample. Importantly, this removes the use of jcenter and
  bintray repositories, which were shut down.

  * Specify NDK version so that Gradle can install required native build
  tools automatically.

  * Define externalNativeBuild so Gradle can perform the Android CMake
  build automatically.

* Replace `main` with `android_main` when building the Amber tool as a
library for Android to avoid compiler warnings, and rename other
Android-related symbols to avoid confusion.

Fixes #964


asuonpaa (approved):
Looks good to me.

ben-clayton (approved):

PR #968, opened 2021-11-04
Add support for the new location of CTS inl files

ilkkasaa:
The generated inl files in Vulkan CTS have been moved to another directory and the files
are being generated at build time. To support this, Amber needs a cmake dependency
to the target that generates the files and the directory needs to be included.

This commit adds those changes.


paulthomson (approved):
Thanks!

PR #970, opened 2021-11-10
Add support for multisample resolve

asuonpaa:
Multisample images were already supported, but graphics pipelines were not able to resolve these into single sample images.

### DIFF
--- a/docs/amber_script.md
+++ b/docs/amber_script.md
@@ -555,6 +555,11 @@ contain image attachment content, depth/stencil content, uniform buffers, etc.
   # pipelines.
   BIND BUFFER {buffer_name} AS depth_stencil
 
+  # Attach |buffer_name| as a multisample resolve target. Each color attachment
+  # with more than one sample must have a corresponding single sample resolve
+  # target image.
+  BIND BUFFER {buffer_name} AS resolve
+
   # Attach |buffer_name| as the push_constant buffer. There can be only one
   # push constant buffer attached to a pipeline.
   BIND BUFFER {buffer_name} AS push_constant
--- a/docs/amber_script.md
+++ b/docs/amber_script.md
@@ -555,9 +555,9 @@ contain image attachment content, depth/stencil content, uniform buffers, etc.
   # pipelines.
   BIND BUFFER {buffer_name} AS depth_stencil
 
-  # Attach |buffer_name| as a multisample resolve target. Each color attachment
-  # with more than one sample must have a corresponding single sample resolve
-  # target image.
+  # Attach |buffer_name| as a multisample resolve target. The order of resolve
+  # target images match with the order of color attachments that have more than
+  # one sample.
   BIND BUFFER {buffer_name} AS resolve
 
   # Attach |buffer_name| as the push_constant buffer. There can be only one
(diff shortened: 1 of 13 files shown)


paulthomson:
```
[1472/1501] Building CXX object src/vulkan/CMakeFiles/libamberenginevulkan.dir/frame_buffer.cc.o
FAILED: src/vulkan/CMakeFiles/libamberenginevulkan.dir/frame_buffer.cc.o
/bin/clang-10.0.0/bin/clang++  -DAMBER_CTS_VULKAN_HEADER=0 -DAMBER_ENABLE_CLSPV=0 -DAMBER_ENABLE_DXC=0 -DAMBER_ENABLE_LODEPNG=1 -DAMBER_ENABLE_RTTI=0 -DAMBER_ENABLE_SHADERC=1 -DAMBER_ENABLE_SPIRV_TOOLS=1 -DAMBER_ENABLE_VK_DEBUGGING=1 -DAMBER_ENGINE_DAWN=0 -DAMBER_ENGINE_VULKAN=1 -I../include -I../ -I../third_party/spirv-tools/include -I. -I../third_party/vulkan-headers/include -I../third_party/cppdap/include -g -fPIC   -fno-exceptions -fvisibility=hidden -Wall -Werror -Wextra -Wno-padded -Wno-switch-enum -Wno-unknown-pragmas -pedantic-errors -fno-rtti -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-format-pedantic -Wno-unknown-warning-option -Weverything -Wno-zero-as-null-pointer-constant -std=gnu++14 -MD -MT src/vulkan/CMakeFiles/libamberenginevulkan.dir/frame_buffer.cc.o -MF src/vulkan/CMakeFiles/libamberenginevulkan.dir/frame_buffer.cc.o.d -o src/vulkan/CMakeFiles/libamberenginevulkan.dir/frame_buffer.cc.o -c ../src/vulkan/frame_buffer.cc
../src/vulkan/frame_buffer.cc:38:7: error: field 'depth_stencil_attachment_' will be initialized after field 'resolve_targets_' [-Werror,-Wreorder-ctor]
      depth_stencil_attachment_(depth_stencil_attachment),
      ^
1 error generated.
```

docs/amber_script.md
```diff
@@ -555,6 +555,11 @@ contain image attachment content, depth/stencil content, uniform buffers, etc.
   # pipelines.
   BIND BUFFER {buffer_name} AS depth_stencil
 
+  # Attach |buffer_name| as a multisample resolve target. Each color attachment
+  # with more than one sample must have a corresponding single sample resolve
```
paulthomson, line 559: When you say "must have", does this mean your change might break existing tests, including those in Vulkan CTS?

paulthomson (commented):
Looks good to me, but I left one question.

docs/amber_script.md
asuonpaa: I should probably reword this. I wasn't seeing a validation error without these so it should be still valid if I don't add the resolve images. Not that there's any point using multisample images without resolve as the results cannot be copied out of GPU memory with a transfer queue. Previously the multisample Amber usage in CTS has been compute shader tests where the results are checked inside the same shader where the samples get written. So nothings is getting broken.

I will change the text to something like "The order of resolve target images match with the order of color buffers that have more than one sample."

paulthomson (approved):

PR #976, opened 2022-01-27
Roll dependencies

asuonpaa:


paulthomson (approved):
Thanks for all your work on this, finding bugs in the Vulkan loader, etc.!

PR #977, opened 2022-01-28
Remove HLSL debugging tests from suppression list.

asuonpaa:
After some modifications these tests were passing locally so these should not be on a suppression list.


asuonpaa:
Release build bots seem to fail on a single debugging test regardless of which I disable. Maybe one of the tests break the following one as they all communicate through a debugger port.

PR #980, opened 2022-03-11
Check error return when pushing descriptors

gnl21:
Pipeline::SendDescriptorDataToDeviceIfNeeded was not checking the error
return when initializing transfer buffers, which can lead to later code
attempting to use invalid buffers and crashing.


gnl21:
@paulthomson is there a process for getting this PR reviewed? It's causing problems for us in the Vulkan CTS. Thanks.

paulthomson (approved):

gnl21:
Great. No problem. Thanks.

PR #987, opened 2022-04-19
Add depth compare support for samplers

asuonpaa:


ilkkasaa:
LGTM

docs/amber_script.md
```diff
@@ -363,6 +363,8 @@ SAMPLER {name} \
     [ MIN_LOD _val_ (default 0.0) ] \
     [ MAX_LOD _val_ (default 1.0) ] \
     [ NORMALIZED_COORDS | UNNORMALIZED_COORDS (default NORMALIZED_COORDS) ]
+    [ COMPARE _val_ (default off) ] \
+    [ COMPARE_OP _val_ (default never) ] \
```
paulthomson, line 367: Technically, I think you should add a `\` at the end of the existing line, and you should remove the final `\` that you added.

Also, could you please list the compare values (on and off?) above, and compare ops -> looks like these are listed below under "Pipeline Content", but I suggest that you just repeat them above.

asuonpaa: Done

paulthomson (approved):

PR #993, opened 2022-07-27
kokoro: remove pre-existing core file if it exists

dneto0:


dneto0:
trying to see if I can avoid a failure as in CI-ubuntu-clang-release in #992

dneto0:
https://github.com/google/clspv/pull/878 is a pre-existing solution.

PR #997, opened 2022-09-01
Require C++17 for clspv and swiftshader

alan-baker:


dj2:
Does the Vulkan CTS support c++17?

alan-baker:
I'm not sure, but newer LLVMs require it. So we're going to get dragged along at some point.

dj2 (approved):

Issue #999, opened 2022-12-07
SwiftShader part of the build fails


dneto0:
Building from scratch with the DEPS as specified, I get this compile error on Ubuntu with Clang-14 as the compiler:

[12/125] Building CXX object third_party/swiftshader/src/Vulkan/CMakeFiles/vk_swiftshader.dir/Debug/EventListener.cpp.o
FAILED: third_party/swiftshader/src/Vulkan/CMakeFiles/vk_swiftshader.dir/Debug/EventListener.cpp.o 
[...]

PR #1001, opened 2022-12-08
Remove shader debugging from Amber

ben-clayton:
SwiftShader was the only implementation that supported this, but has disabled this code due to lack of use. If there's no driver that supports shader debugging, then it doesn't need to exist and be maintained in Amber.

If one day, we want this again, then this CL can be reverted.

Fixes: #999


kokoro/android/build.sh
```diff
@@ -26,6 +26,10 @@ ANDROID_ABI="armeabi-v7a with NEON"
 
 TOOLCHAIN_PATH="$ANDROID_NDK/build/cmake/android.toolchain.cmake"
 
+# Disable git's "detected dubious ownership" error - kokoro checks out the repo with a different
+# user, and we don't care about this warning.
```
dneto0, line 30: Heh, fascinating.

dneto0 (approved):
With regret, this is the right choice.
Thank you, this was/is amazing work.

PR #1004, opened 2022-12-08
git-sync-deps: show the tail, not the head of what's being checked out

dneto0:
This is more informative when in a deep path.

### DIFF
--- a/tools/git-sync-deps
+++ b/tools/git-sync-deps
@@ -116,11 +116,11 @@ def is_git_toplevel(git, directory):
 
 
 def status(directory, checkoutable):
-  def truncate(s, length):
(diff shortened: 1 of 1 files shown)


tools/git-sync-deps
```diff
@@ -116,11 +116,11 @@ def is_git_toplevel(git, directory):
 
 
 def status(directory, checkoutable):
-  def truncate(s, length):
-    return s if len(s) <= length else s[:(length - 3)] + '...'
+  def tail_truncate(s, length):
+    return s if len(s) <= length else s[-(length - 3):] + '...'
```
ben-clayton, line 120: Maybe put the `...` on the LHS of the trimmed string?

PR #1005, opened 2022-12-08
Roll many deps

dneto0:
Many projects are updated to top of tree.
Except vulkan-validationlayers is only rolled to 19314aa5901ce13273adac4fa630239e37544dfa from October 10 or so.


dneto0:
Trying to get ahead of any bit-rotting.

dneto0:
Updated to avoid rolling vulkan-headers, vulkan-loader, and vulkan-validationlayers.

Vulkan-loader and vulkan-headers have been modified to not be able to build with a single "cmake" followed by a single build statement.

PR #1006, opened 2022-12-09
Android cmake build: use cmake 3.14 since Glslang needs it

dneto0:
Change-Id: Ie83f7ed2b1f27f0f92ea09077cb669015c90498f


dneto0:
The CI-check-format is failing due to a python3 issue.
https://github.com/google/amber/issues/1007

dneto0:
CI-check-format is failing due to #1007: Python 3 vs. cpplint.

PR #1011, opened 2023-01-06
Remove -Weverything compiler option

jbauman42:
This flag includes warnings that are not intended for end-users to use - for example it includes the experimental -Wunsafe-buffer-usage flag which only makes sense when attempting to use a safe subset of C++.

See https://quuxplusone.github.io/blog/2018/12/06/dont-use-weverything/ for a more in-depth explanation of the problems with this flag.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -225,7 +225,6 @@ function(amber_default_compile_options TARGET)
         -Wno-c++98-compat-pedantic
         -Wno-format-pedantic
         -Wno-unknown-warning-option
-        -Weverything
       )
     endif()
   endif()


dj2:
We've had `-Weverything` enabled since the project started. It triggers new flags now and then, but the preference is to disable the warning if it doesn't make sense for Amber as opposed to removing `-Weverthing`. As is, this will turn off a whole pile of warnings which are currently running against the code base.

Adding a cmake option to remove `-Weverything` (and probably `-Werror`) would also be fine, as long as they're enabled by default.

dneto0:
I agree with @dj2 here.
We explicitly opt into the pain of having to explicitly turn off new problematic flags.

I believe the -Wunsafe-buffer-usage flag was proven to have a buggy implementation.
We've explicitly turned it off in other projects maintained in our group.  See https://github.com/google/clspv/pull/996/files#diff-1ad599c51cce14fef2ea1a929fa6d8098e79b702cefaaa9dd173f92f62dfe152R316

dneto0:
> See https://quuxplusone.github.io/blog/2018/12/06/dont-use-weverything/ for a more in-depth explanation of the problems with this flag.

I should be clear.  I did read that blog post and I *still* think this project should turn it on.
Basically I'm appealing to the robustness principle.

I think the example given in the blog post is not persuasive.  I would want to add the new line at the end of file anyway.

dneto0 (changes requested):
I don't want this change.  See @dj2's advice if you do want to push on this further.

jbauman42:
The problem is when amber is used in other projects like VK-GL-CTS; if we ever want to use a new compiler with the vulkan conformance tests, for correctness every version of amber that's used in a released version of VK-GL-CTS needs to have the flag added, and rolled into that version of VK-GL-CTS.

dj2:
Right, that's why I suggest a cmake flag to turn off `-Weverything` and `-Werror` then the VK-GL-CTS can just flip the flag off if they desire.

PR #1013, opened 2023-01-06
Add AMBER_DISABLE_WEVERYTHING and AMBER_DISABLE_WERROR options

jbauman42:
New compiler versions may add additional warnings; in particular, -Weverything includes warnings that are unstable and may not be generally-applicable.

To assist projects like VK-GL-CTS that may depend on old versions of amber for a long period of time, add AMBER_DISABLE_WERROR and AMBER_DISABLE_WEVERYTHING flags to allow disabling those warnings downstream. That allows those projects to be built with newer versions of clang without needing to fork amber and disable unwanted warnings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,8 @@ option(AMBER_ENABLE_SWIFTSHADER
   "Build using SwiftShader" ${AMBER_ENABLE_SWIFTSHADER})
 option(AMBER_ENABLE_RTTI
   "Build with runtime type information" OFF)
+option(AMBER_DISABLE_WERROR "Build without the -Werror flag" ${AMBER_DISABLE_WERROR})
+option(AMBER_DISABLE_WEVERYTHING "Build without the -Weverything flag" ${AMBER_DISABLE_WEVERYTHING})
 
 if (${AMBER_ENABLE_VK_DEBUGGING})
   message(FATAL_ERROR "Amber no longer supports Vulkan debugging")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -49,6 +49,8 @@ option(AMBER_ENABLE_SWIFTSHADER
   "Build using SwiftShader" ${AMBER_ENABLE_SWIFTSHADER})
 option(AMBER_ENABLE_RTTI
   "Build with runtime type information" OFF)
+option(AMBER_DISABLE_WERROR "Build without the -Werror flag" ${AMBER_DISABLE_WERROR})
```
dj2, line 52: I think this will make both of these default to off: https://cmake.org/cmake/help/latest/command/option.html

Probably want `ON` instead of `${AMBER_DISABLE_WERROR}`.

jbauman42: I was thinking we'd leave the same default behavior as before (with Werror enabled) so AMBER_DISABLE_WERROR should be off by default.

dj2: Ah, good point. Double negatives are confusing. So, `OFF` by default means enable WError and WEverything

dj2 (approved):

dj2:
@dneto0 are you happy with this? It looks good to me, but wanted to make sure you were good.

dneto0 (approved):

dj2:
Apologies, I just realized this didn't get merged. I'm going to run the bots and get it merged for you. Sorry for the delay.

PR #1025, opened 2023-06-08
Added support for the Vulkan performance pipeline runtime layer

mdagois:
Added support for the Vulkan performance pipeline runtime layer

Contrary to the validation layer, this is an opt-in layer.
Opt-in is achieved by passing --enable-runtime-layer to the CLI.


s-perron (approved):

s-perron:
@mdagois Can you run it through clang format when you get a chance?

mdagois:
> @mdagois Can you run it through clang format when you get a chance?

I ran it and committed. I used the .clang-format in the repository, but I am not sure I have the same version of clang-format as used in CI.

PR #1029, opened 2023-06-22
Vkqueuewaitidle correct author

phoenixxxx:
google/vulkan-performance-layers requires a call to vkDeviceWaitIdle or vkQueueWaitIdle. Since we want to be able to use that layer in conjunction with Amber we need to somehow communicate that the Amber script has completed.


s-perron:
I restarted to two failing tests. They might be flaky. Looking at the history of the test, it seems to fail every once in a while.

src/vulkan/command_buffer.cc
```diff
@@ -117,6 +118,15 @@ Result CommandBuffer::SubmitAndReset(uint32_t timeout_ms) {
...
   }
 
+    /*
+  google/vulkan-performance-layers requires a call to vkDeviceWaitIdle or
+  vkQueueWaitIdle in order to report the information. Since we want to be
+  able to use that layer in conjunction with Amber we need to somehow
+  communicate that the Amber script has completed.
+  */
+  if (pipeline_runtime_layer_enabled)
+    device_->GetPtrs()->vkQueueWaitIdle(device_->GetVkQueue());
```
s-perron, line 128: Why don't we run this all of the time? Does it hurt?  It seems like this adds a lot of code just to pass a flag.

dj2 (approved):

PR #1041, opened 2024-06-06
CMake: Use find_package(Python3) to find Python

dneto0:
Fixes CMake deprecation warning CMP0148


dneto0:
https://cmake.org/cmake/help/latest/policy/CMP0148.html

dj2 (approved):

PR #1044, opened 2024-06-19
Add geometry flags support

archimedus:
Extend acceleration structure geometries (both AABBs and traingles) with support of flags.


archimedus:
And that's minor functionality extension for acceleration structures, that I separated out from library PR.
Please give it a try.

dj2 (approved):

PR #1045, opened 2024-07-24
Remove unneeded memory requirement

archimedus:
Remove VK_MEMORY_PROPERTY_HOST_CACHED_BIT bit, that actually it is optional and may lead to situation memory will not be allocated.


archimedus:
@dj2 

Can you please run kokoro tests?

Thank you.

dj2 (approved):

archimedus:
I think it is ready for review and merge.

PR #1048, opened 2024-08-22
Amber script documentation for new TIMED_EXECUTION  flag for RUN

petermcneeleychromium:
Amber script documentation for new TIMED_EXECUTION flag for RUN

This is for benchmarking effort:
crbug.com/361315331


dj2 (approved):

petermcneeleychromium:
I have added the implementation in Amber which also includes the vk query backend implementation

crbug.com/361315331

src/command.h
```diff
@@ -148,10 +148,23 @@ class PipelineCommand : public Command {
   Pipeline* pipeline_ = nullptr;
 };
 
+/// Base class for commands which contain a pipeline.
+class TimedPipelineCommand : public PipelineCommand {
```
dj2, line 152: Are there any pipelines which aren't timed? If not, why the split?

src/command.h
```diff
@@ -148,10 +148,23 @@ class PipelineCommand : public Command {
...
+
+  bool timed_execution_ = false;
+};
+
 /// Command to draw a rectangle on screen.
-class DrawRectCommand : public PipelineCommand {
+class DrawRectCommand : public TimedPipelineCommand {
  public:
-  DrawRectCommand(Pipeline* pipeline, PipelineData data);
+  DrawRectCommand(Pipeline* pipeline, PipelineData data, bool timed_execution);
```
dj2, line 167: I think I'd prefer a `SetTimedExecution` similar to the setters for all the other params. Keep sit a bit more consistent with how the rest of the pipeline works. Can default to `false`, so we really only need to set it to enable it.

dj2: Could also add the setter to the base class, which minimizes the number of changes in here.

src/vkscript/command_parser.cc
```diff
@@ -46,6 +46,9 @@ ShaderType ShaderNameToType(const std::string& name) {
   return kShaderTypeVertex;
 }
 
+// Timed execution is not supported in VkScript.
+constexpr bool kTimedExecutionSupported = false;
```
dj2, line 50: If we default to false and move to a setter,none of these changes should be necessary

src/amberscript/parser_run_timed_execution_test.cc
```diff
@@ -0,0 +1,248 @@
+// Copyright 2019 The Amber Authors.
```
dj2, line 1:
```suggestion
// Copyright 2024 The Amber Authors.
```

src/amberscript/parser_run_timed_execution_test.cc
```diff
@@ -0,0 +1,248 @@
...
+
+#include "gtest/gtest.h"
+#include "src/amberscript/parser.h"
+
+namespace amber {
+namespace amberscript {
+
+using AmberScriptParserTest = testing::Test;
+
+TEST_F(AmberScriptParserTest, RunComputeTimedExecution) {
```
dj2, line 23: Maybe one more test, similar to this but drop the `TIMED_EXECUTION` and validate `EXPECT_FALSE(cmd->AsCompute()->IsTimedExecution())`

src/vulkan/device.cc
```diff
@@ -460,7 +466,8 @@ Result Device::Initialize(
     const VkPhysicalDeviceFeatures2KHR& available_features2,
     const VkPhysicalDeviceProperties2KHR& available_properties2,
     const std::vector<std::string>& available_extensions) {
-  Result r = LoadVulkanPointers(getInstanceProcAddr, delegate);
+  delegate_ = delegate;
```
dj2, line 469: If we need to store this, can we move it to the constructor?

src/vulkan/engine_vulkan.cc
```diff
@@ -760,13 +760,15 @@ Result EngineVulkan::DoDrawGrid(const DrawGridCommand* command) {
   vertex_buffer->SetData(0, buf.get(), InputRate::kVertex, buf->GetFormat(), 0,
                          buf->GetFormat()->SizeInBytes());
 
-  DrawArraysCommand draw(command->GetPipeline(), *command->GetPipelineData());
+  DrawArraysCommand draw(command->GetPipeline(), *command->GetPipelineData(),
+                         command->IsTimedExecution());
```
dj2, line 764: Similar to above, this seems like it would be better as a `SetTimedExecution` call.

src/vulkan/graphics_pipeline.cc
```diff
@@ -916,6 +917,10 @@ Result GraphicsPipeline::Draw(const DrawArraysCommand* command,
     frame_->CopyBuffersToImages();
     frame_->TransferImagesToDevice(GetCommandBuffer());
 
+    // Timing must be place outside the render pass scope. The full pipeline
+    // barrier used by our specific implementation cannot be within a
+    // renderpass.
+    BeginTimerQuery(is_timed_execution);
```
dj2, line 923: There are early returns below, do we need to end and destroy the timer queries?

src/vulkan/pipeline.cc
```diff
@@ -253,6 +260,82 @@ void Pipeline::UpdateDescriptorSetsIfNeeded() {
   }
 }
 
+void Pipeline::CreateTimingQueryObjectIfNeeded(bool is_timed_execution) {
+  if (!is_timed_execution ||
+      !device_->IsTimestampComputeAndGraphicsSupported()) {
+    return;
+  }
+  VkQueryPoolCreateInfo pool_create_info{
```
dj2, line 268: nit: blank line above

src/vulkan/pipeline.cc
```diff
@@ -253,6 +260,82 @@ void Pipeline::UpdateDescriptorSetsIfNeeded() {
...
+  device_->GetPtrs()->vkCreateQueryPool(
+      device_->GetVkDevice(), &pool_create_info, nullptr, &query_pool_);
+}
+
+void Pipeline::DestroyTimingQueryObjectIfNeeded(bool is_timed_execution) {
+  if (!is_timed_execution ||
+      !device_->IsTimestampComputeAndGraphicsSupported()) {
+    return;
+  }
+  // Flags set so we may/will wait on the CPU for the availiblity of our
```
dj2, line 284: nit blank line above

src/vulkan/pipeline.cc
```diff
@@ -253,6 +260,82 @@ void Pipeline::UpdateDescriptorSetsIfNeeded() {
...
+  // queries.
+  const VkQueryResultFlags flags =
+      VK_QUERY_RESULT_WAIT_BIT | VK_QUERY_RESULT_64_BIT;
+  uint64_t time_stamps[kNumQueryObjects];
```
dj2, line 288: nit: std::array<uint64_t, kNumQueryObjects> time_stamps = {};

src/vulkan/pipeline.cc
```diff
@@ -253,6 +260,82 @@ void Pipeline::UpdateDescriptorSetsIfNeeded() {
...
+                                          VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT,
+                                          query_pool_, kBeginQueryIndexOffset);
+}
+
+void Pipeline::EndTimerQuery(bool is_timed_execution) {
+  if (!is_timed_execution ||
+      !device_->IsTimestampComputeAndGraphicsSupported()) {
+    return;
+  }
+  // Full barrier ensures that work after our timing is executed before the
```
dj2, line 327:
```suggestion
  // Full barrier ensures that work included in our timing is executed before the
```

samples/amber.cc
```diff
@@ -627,6 +643,29 @@ int main(int argc, const char** argv) {
...
+      for (auto& timing : execution_timing) {
+        if (!is_first_iter) {
+          std::cout << ", ";
+        }
+        is_first_iter = false;
+        std::cout << timing;
+      }
+      std::cout << std::endl;
+      std::sort(execution_timing.begin(), execution_timing.end());
+      auto report_median =
```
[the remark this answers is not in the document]
petermcneeleychromium: Median is the mean of the 2 middle values if the list is even.

dj2 (approved):

dj2:
Awesome, thank you.

PR #1052, opened 2024-09-25
Make extention functions to be optional

archimedus:


archimedus:
This is not on top of main due to VK-GL-CTS does not compile for me with main.

dj2 (approved):

PR #1054, opened 2024-09-25
Minor fix to remove vulkan_core header

petermcneeleychromium:
and made the delegate function have an default empty implementation (instead of pure virtual)


lordalcol:
I would really like this to be merged, as we depend on the fix for CTS.

archimedus:
With this fix I can compile CTS main branch locally.

dj2 (approved):

PR #1055, opened 2024-10-03
Reproduction for webgpu cts atomic operation failure on mobile gpus (mali, adreno)

petermcneeleychromium:
This very simple amber file demonstrates a repo of the bug in webgpu CTS atomicAdd

This test was manually run on a mali (pixel 6) and it outputs the value 8 rather than 10.
Incorrect behavior appears to be that the shader is only executed on one thread (instead of 3).

crbug.com/42241359

### DIFF
--- a/tests/cases/compute_workgroup_atomic.amber
+++ b/tests/cases/compute_workgroup_atomic.amber
@@ -46,7 +46,7 @@ void main() {
 }
 END
 
-BUFFER buf_uint DATA_TYPE uint32 SIZE 1024 FILL 123
+BUFFER buf_uint DATA_TYPE uint32 SIZE 16 FILL 123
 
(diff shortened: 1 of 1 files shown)


petermcneeleychromium:
Are you concerned about a pollution of Amber test cases with bug reproductions?

dj2:
No, just more that this seems like the kind of thing that isn't betting tested by the Vulkan CTS, but probably should be. So it would be nice to have this in the CTS so it doesn't slip into other drivers/releases.

petermcneeleychromium:
That makes sense. I was really just doing this PR as a way of saving my work and making it publicly available.

petermcneeleychromium:
This test is in the process of being added to vk cts.

PR #1056, opened 2024-10-10
Support loading shaders from a FILE and as SPIRV-BIN

nhaehnle:
This is convenient in a number of cases. For example, one can more conveniently refer to a standalone shader source file. And one can pass-through a SPIR-V file that may be using a shader-only extension that Amber is unaware of.


src/amberscript/parser_buffer_test.cc
```diff
@@ -48,6 +48,10 @@ class DummyDelegate : public amber::Delegate {
 
     return {};
   }
+
+  amber::Result LoadFile(const std::string, std::vector<char>&) const override {
+    return Result("DummyDelegate::LoadFile not implemented");
```
dj2, line 53:
```suggestion
    return Result("PlaceholderDelegate::LoadFile not implemented");
```

dj2 (approved):

src/amberscript/parser_buffer_test.cc
nhaehnle: Are you sure? The class name is DummyDelegate.

nhaehnle:
Thank you for taking a look!

The license check complains that shader_spirv_bin.spv doesn't have a license. I suppose that's true, but given it's a binary file that may be difficult to satisfy. And while checking in binary files is obviously usually a bit questionable, it's in the nature of the test. Do you have any suggestions?

src/amberscript/parser_buffer_test.cc
dj2: Oh, you're right. I'd thought I'd renamed that already. Leave this and I'll rename the delegate in the future.

dj2:
Hm, I've been looking but can't find the documentation for license-check. I also don't have admin, and am unable to ignore the check. @dneto0 do you have the ability to ignore the failing required check and merge this regardless? (or do you know the magic to make license-checker skip the binary file?)

dneto0:
> Hm, I've been looking but can't find the documentation for license-check. I also don't have admin, and am unable to ignore the check. @dneto0 do you have the ability to ignore the failing required check and merge this regardless? (or do you know the magic to make license-checker skip the binary file?)

I can't bypass the license checker.
However, we should be able to exclude .spv files by modifying the `license-checker.cfg` in the project root dir.

I'll do it in a new PR.

dneto0:
Landed as #1059

nhaehnle:
Thanks!

PR #1059, opened 2024-11-20
Support loading shaders from a FILE and as SPIRV-BIN

dneto0:
This is convenient in a number of cases. For example, one can more
    conveniently refer to a standalone shader source file. And one can
    pass-through a SPIR-V file that may be using a shader-only extension
    that Amber is unaware of.

This is the work of @nhaehnle

This is a rebase of #1056 on top of main


dj2:
lgtm. The update button didn't work on the other PR?

dj2 (approved):

dneto0:
> lgtm. The update button didn't work on the other PR?

no.

PR #1060, opened 2024-12-03
Check Vulkan 1.4 features for indexTypeUint8

lordalcol:


src/vulkan/device.cc
```diff
@@ -767,6 +765,22 @@ Result Device::Initialize(
         return amber::Result("Missing compute full subgroups feature");
       }
     }
+
+    // If Vulkan 1.4 structure exists the features are set there.
```
dj2, line 818: What happens if vulkan 1.4 isn't provided, does this mean it can no longer find the feature it previously found? Does the old way not work anymore, like is `VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_INDEX_TYPE_UINT8_FEATURES_EXT` no longer returned above so the pointer isn't set?

src/vulkan/device.cc
```diff
@@ -610,6 +614,10 @@ Result Device::Initialize(
       return amber::Result(
           "Subgroup extended types requested but feature not returned");
     }
+    if (feature == kAccelerationStructure &&
```
dj2, line 617: This seems unrelated?

lordalcol: yes, I think it's some merge gone bad. I'll re-do it

lordalcol:
Abandoned in favour of https://github.com/google/amber/pull/1061

PR #1061, opened 2024-12-03
Add support for Vulkan 1.4

lordalcol:


dj2 (approved):

lordalcol:
@gfxstrand Since you were the original author of this, could you help replying and amending if needed?

dj2:
The change looks good to me, just need to see if the bots are happy with it.

lordalcol:
> The change looks good to me, just need to see if the bots are happy with it.

What about that other comment you left, that constrains uint8 to Vulkan 1.4?

dj2:
Either I mis-read the other PR, or this one is different because it does do an `if vk1.4 {} else { old thing }`

dj2:
(looking again, I mis-read the other PR, heh)

lordalcol:
Builds failed. I'll take a look at what is needed to fix.

lordalcol:
I have a feeling it may need a Vulkan headers update. Not sure how to do it here.

dj2:
Change the sha in the `vulkan_headers_revision` in the `DEPS` file.

dj2:
I'm guessing you're going to need to do all the vulkan libraries as they tend to be intertwined.

lordalcol:
> I'm guessing you're going to need to do all the vulkan libraries as they tend to be intertwined.

Yes, I'm doing them all now, checking locally too.

dj2:
Merged. Thanks a lot for the fixes.

Issue #1062, opened 2024-12-11
Vulkan validation layer error when running latest


Zhaeong:
With latest when running it locally I'm getting:

```
./amber
[ERROR] validation layer (Validation):
Validation Error: [ VUID-VkDeviceCreateInfo-pNext-pNext ] Object 0: handle = 0x5e4978bc0d30, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0x901f59ec | vkCreateDevice(): pCreateInfo->pNext<VkPhysicalDeviceShaderFloat16Int8Features> includes a pointer to a VkPhysicalDeviceShaderFloat16Int8Features, but when creating VkDevice, the parent extension (VK_KHR_shader_float16_int8) was not included in ppEnabledExtensionNames.
The Vulkan spec states: Each pNext member of any structure (including this one) in the pNext chain must be either NULL or a pointer to a valid struct for extending VkDeviceCreateInfo (https://vulkan.lunarg.com/doc/view/1.3.296.0/linux/1.3-extensions/vkspec.html#VUID-VkDeviceCreateInfo-pNext-pNext)
[ERROR] validation layer (Validation):
Validation Error: [ VUID-VkDeviceCreateInfo-pNext-pNext ] Object 0: handle = 0x5e4978bc0d30, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0x901f59ec | vkCreateDevice(): pCreateInfo->pNext<VkPhysicalDevice8BitStorageFeatures> includes a pointer to a VkPhysicalDevice8BitStorageFeatures, but when creating VkDevice, the parent extension (VK_KHR_8bit_storage) was not included in ppEnabledExtensionNames.
The Vulkan spec states: Each pNext member of any structure (including this one) in the pNext chain must be either NULL or a pointer to a valid struct for extending VkDeviceCreateInfo (https://vulkan.lunarg.com/doc/view/1.3.296.0/linux/1.3-extensions/vkspec.html#VUID-VkDeviceCreateInfo-pNext-pNext)
[ERROR] validation layer (Validation):
[...]

PR #1063, opened 2024-12-11
Fix missing used extensions

petermcneeleychromium:
This fix was motivated by a validation layer error.
https://github.com/google/amber/issues/1062

This code is in need of a some refactors but it is best to fix first and refactor later.


petermcneeleychromium:
This is only a failure for vulkan 1.1<=   which is why it was missed in local testing with -v 1.2

dj2 (approved):

PR #1069, opened 2024-12-23
set cmake_minimum_required to 3.22

dneto0:
That's the cmake verion included in Ubuntu 2022.04

Change-Id: I5d6da0d33c86eae5020bba1b9a8805fc8d7e7868

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cmake_minimum_required(VERSION 3.13)
+cmake_minimum_required(VERSION 3.22)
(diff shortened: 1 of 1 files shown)


dj2 (approved):

dneto0:
The MSVC 2017 configs are invalid now, and have been deleted on the Kokoro config side internal to Google

dneto0:
The MSVC failures seem to be related to:
1. various subprojects manually change `/MD` to `/MT` to force static linking for MSVC
2. Cmake changed how it sets those flags. See https://gitlab.kitware.com/cmake/cmake/-/commit/fb3370b6a1681190ffd8daf63975c44ce8fc1c49

PR #1072, opened 2025-01-07
Add VK instance and device creation logging.

dj2:
Add the result string into the error logs for device and instance creation in order to provide a bit more context as to what failed.

Change-Id: I6c0ce8ee3ed45a7c138f4f4f90be1e602f3a0423


dneto0 (approved):
Nice!

PR #1073, opened 2025-01-07
Fixup acceleration structure and raytracing flags.

dj2:
When determining if acceleration structures and ray tracing are supported it isn't enough to just check for the device extension. Both of these extensions maybe available, then the features contain a flag which determines if it is implemented. So, if we've got the features, we need to additionally check the flag and reset our internal tracking based on the implementation flag.

Change-Id: I09b04007a47e5b5bec34c7ed910949401ca18a3b

### DIFF
--- a/samples/config_helper_vulkan.cc
+++ b/samples/config_helper_vulkan.cc
@@ -920,7 +920,7 @@ amber::Result ConfigHelperVulkan::CheckVulkanPhysicalDeviceRequirements(
     vkGetPhysicalDeviceFeatures2KHR(physical_device, &features2);
     available_features_ = features2.features;
 
-    // Just having the feature does not necessarily mean that the feature is available. We have to
+    // Just having the extension does not necessarily mean that the feature is available. We have to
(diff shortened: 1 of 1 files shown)


samples/config_helper_vulkan.cc
```diff
@@ -920,6 +920,15 @@ amber::Result ConfigHelperVulkan::CheckVulkanPhysicalDeviceRequirements(
     vkGetPhysicalDeviceFeatures2KHR(physical_device, &features2);
     available_features_ = features2.features;
 
+    // Just having the feature does not necessarily mean that the feature is available. We have to
```
dneto0, line 923: Recommend changing "feature" to "extension" like this:

> Just having the extension does not....

dneto0 (changes requested):
please update the comment.

samples/config_helper_vulkan.cc
dj2: done.

dneto0 (approved):

PR #1074, opened 2025-01-08
Add DEVICE_FEATURE for depth clamp zero one

gnl21:


dj2 (approved):

gnl21:
I see that the NDK autobuild is failing to find any of the tokens that are used in this change, but the version of the vulkan headers in DEPS does include them. Do you know if there's any special that I should be aware of around that?

dneto0:
I think the in the Android.mk build the vulkan headers are taken from the NDK.  The bots use ndk-r25b which I think is  acouple of years old.

I'll update the bots.

dneto0:
The NDK r27c includes vulkan.h for 1.3.275, and the needed tokens are in there.

dneto0:
#1076 updates the scripts to use NDK r27c.
Please rebase on top of main

PR #1075, opened 2025-01-08
Fix Vulkan extension checking

gnl21:
The wrong copy of the descriptor indexing struct was being added to the extension chain meaning that everything else was discarded.

### DIFF
--- a/samples/config_helper_vulkan.cc
+++ b/samples/config_helper_vulkan.cc
@@ -987,7 +987,7 @@ amber::Result ConfigHelperVulkan::CheckVulkanPhysicalDeviceRequirements(
     if (supports_descriptor_indexing_) {
       descriptor_indexing_features.sType =
           VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DESCRIPTOR_INDEXING_FEATURES_EXT;
-      descriptor_indexing_feature_.pNext = next_ptr;
+      descriptor_indexing_features.pNext = next_ptr;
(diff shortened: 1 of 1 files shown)


gnl21:
@petermcneeleychromium it looks like this was incorrect in your recent change. Your refactor fixed a very similar issue with a different pNext chain, but unfortunately this still seems quite error prone.

dj2 (approved):

dj2:
I wonder if moving all of the `feature_` things in `ConfigHelperVulkan` into a separate structure to make the naming more obvious. Then you'd have `features_.descriptor_indexing` when accessing the stored one and `descriptor_indexing_feature` for use inside the method as we have now.

gnl21:
I think that something like that would be beneficial. There's a source of bugs here that I haven't tried to solve, just to correct the thing that's wrong at the moment.

A renaming like you suggest would make the differences clearer, which would be nice. A refactor like Peter's that does the adding to the chain in a standard way would also help because, since you only have to type the name of each struct once, if you type it wrong then your tests won't work and you'll notice straightaway. What happened here is the mix of struct names meant the tests for the code that was added worked fine but everything else was broken.

dj2:
I think you need to sign the CLA for me to land this. If you take a look at https://github.com/google/amber/blob/main/CONTRIBUTING.md it has directions.

dj2:
Oh, nm, cla bot says it's ok.

PR #1076, opened 2025-01-08
Use NDK r27c for Android builds

dneto0:
It has vulkan.h version 1.3.275

Change-Id: I833f9ed7fba8b406ccf31a7c32914eb166d77a8a


dneto0:
We should use the docker image for these, I think. That would save the download each time these are run(!)

PR #1078, opened 2025-01-09
Group members of ConfigHelperVulkan.

dj2:
Move the various members into separate structures to a) group them and b) make the names more distinct from other versions used in the helper.

This adds a `vk_` group for all the vulkan objects, a `supports_` struct for all the booleans and a `features_` struct for all the feature objects.

### DIFF
--- a/samples/config_helper_vulkan.h
+++ b/samples/config_helper_vulkan.h
@@ -103,45 +103,51 @@ class ConfigHelperVulkan : public ConfigHelperImpl {
   /// Writes information related to the vulkan instance to stdout.
   void DumpPhysicalDeviceInfo();
 
-  VkInstance vulkan_instance_ = VK_NULL_HANDLE;
-  VkDebugReportCallbackEXT vulkan_callback_ = VK_NULL_HANDLE;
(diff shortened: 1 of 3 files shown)


dj2:
@gnl21 fyi, this is the change I suggested as maybe making things clearer. Does this seem a bit better?

samples/config_helper_vulkan.h
```diff
@@ -103,45 +103,51 @@ class ConfigHelperVulkan : public ConfigHelperImpl {
...
+    bool buffer_device_address = false;
+    bool ray_tracing_pipeline = false;
+    bool descriptor_indexing = false;
+    bool deferred_host_operations = false;
+    bool spirv_1_4 = false;
+    bool shader_float_controls = false;
+  } supports_;
+
+  struct {
+    VkPhysicalDeviceFeatures device{};
```
petermcneeleychromium, line 136: I feel a bit weird about this name but I guess it makes sense since you will always be accessing this device member via the features_ object.

petermcneeleychromium (approved):

samples/config_helper_vulkan.h
dj2: Yea, that one isn't great, we can always rename it again in the future.

PR #1080, opened 2025-01-10
Remove custom MakeUnique.

dj2:
Since Amber was updated past c++14 we can now use `std::make_unique` instead of the local `amber::MakeUnique`. This Cl updates all uses and removes the custom version.


.gitignore
```diff
@@ -35,3 +35,4 @@ cmake-build-*/
 
 ### The 'compile_commands' file can be generated at root
 compile_commands.json
+.DS_Store
```
petermcneeleychromium, line 38: I assume this is from Mac. Maybe the finder creates this.

petermcneeleychromium (approved):
lgtm

PR #1081, opened 2025-01-10
Change ShaderFormat and ShaderType into enum classes.

dj2:
This CL changes the `enum` for both `ShaderFormat` and `ShaderType` into an `enum class` with a specific size of `uint8_t`. This makes the code a bit nicer to read, and allows checks for switch coverage.


dj2:
Actually, probably can't do this without breaking downstream as the shader info is public.

PR #1084, opened 2025-01-20
Check depthClampZeroOne feature bit.

dj2:
The presence of the DepthClampZeroOne extension isn't enough to verify it's usable, the `depthClampZeroOne` feature flag also needs to be checked in the features structure.

Change-Id: Ic75a2375604d91251f19a10c539e3a5cb180f850


dnovillo (approved):
Thanks!

dneto0 (approved):

PR #1085, opened 2025-01-23
kokoro: use gcc 13

dneto0:


dneto0:
Will try again after updating Clspv and LLVM: https://github.com/google/amber/pull/1086

dj2 (approved):

PR #1087, opened 2025-01-31
Add feature structures only for requires features

s-perron:
This not done in the most efficient way. However, we have a device and
driver that seems to have an issue with an extension it claims to
support. This works around that issue by not having using the physical
device features for features that are not required.

### DIFF
--- a/samples/config_helper_vulkan.cc
+++ b/samples/config_helper_vulkan.cc
@@ -739,6 +739,12 @@ std::string stageFlagBitsToNames(const VkShaderStageFlags bits) {
   return result.str();
 }
 
+bool StartsWith(const std::string& str, const std::string& prefix) {
+  if (prefix.size() > str.size())
+    return false;
+  return std::equal(prefix.begin(), prefix.end(), str.begin());
+}
+
 }  // namespace
 
 ConfigHelperVulkan::ConfigHelperVulkan() = default;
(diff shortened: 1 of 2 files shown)


samples/config_helper_vulkan.cc
```diff
@@ -1242,52 +1249,102 @@ amber::Result ConfigHelperVulkan::CreateDeviceWithFeatures2(
...
+      if (feature == kSubgroupSizeControl) {
+        features_.subgroup_size_control.subgroupSizeControl = VK_TRUE;
+      } else if (feature == kComputeFullSubgroups) {
+        features_.subgroup_size_control.computeFullSubgroups = VK_TRUE;
+      }
+    } else if (feature == kShaderSubgroupExtendedTypes) {
+      init_feature(
+          supports_.shader_subgroup_extended_types,
+          features_.shader_subgroup_extended_types,
+          VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_SUBGROUP_EXTENDED_TYPES_FEATURES,
```
s-perron, line 1318: Clang-format is fine with this, but the linter complains that it is more than 80 characters wide. It is hard to keep it to 80 when the macro is 73 characters. There are something we can do. Let me now if you have strong thoughts on how to handle it.

dneto0: I would recommend turning the linter off for these lines.

s-perron: Good idea. Done.

dj2 (approved):

PR #1095, opened 2025-08-02
Fix naive memory barrier pipeline stages

ziga-lunarg:
Host access masks are used, but VK_PIPELINE_STAGE_HOST_BIT is not provided. 

VK_PIPELINE_STAGE_ALL_COMMANDS_BIT  does not include VK_PIPELINE_STAGE_HOST_BIT


ziga-lunarg:
This actually looks like a bug in VVL, closing

dj2 (approved):

PR #1100, opened 2026-01-19
Adding shader long vector feature

pingpongcat:
Adds support for requesting the shader long vector to be used
DEPS update to alight witch v1.4.337


dj2:
This looks good to me, need to see what the bots say. Is it possible to add an amber example script into the repo to show the functionality?

dj2:
Thanks a lot for the CL. I'm ignoring the NDK build, I think it needs some work to get running so merged as is.

